PR #12, opened 2015-07-30
Docker quickstart: clarify supportability of Docker images

flavio:
Make it clear we are going to support only the Docker images built via the Docker build system.

Ping @eotchi and @taroth21

### DIFF
--- a/xml/vt_docker_quick.xml
+++ b/xml/vt_docker_quick.xml
@@ -459,10 +459,9 @@
 
   <para>
    Some users might be confused about which build system to use for what
-   task. The recommended approach is to build the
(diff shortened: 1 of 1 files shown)


taroth21:
Thanks (and sorry for the delay in accepting)!

taroth21:
@flavio: I just noticed that the link http://docs.docker.com/terms/image/#base-image-def does not seem to work (gives me a 404). Has it changed in the meantime?

flavio:
@taroth21 you are right, can you make it point to: http://docs.docker.com/reference/glossary/#base-image

PR #27, opened 2015-10-22
vt-best-practices: Add a section describing Xen autoballooning

jfehlig:
Add a sub-section about autoballooning and setting Domain-0 memory
to the Memory and Pages section.


aginies:
I will merge and do some XML fixes.

PR #31, opened 2015-11-04
Added article about Amazon EC2 Container Service

schaefi:
This article describes how to use the ECS image provided by the
public cloud development team in Amazon's container service.
Thanks to toms for migrating my db4 sources to db5


schaefi:
for reference this article will referenced on https://aws.amazon.com/docker/partners/ once it is publicly available

PR #38, opened 2015-11-24
Docker quickstart: non interactive installation

flavio:
Explain how to add the Containers Module without using YaST.
This can be really useful for people scripting stuff.

Ping @eotchi and @taroth21

### DIFF
--- a/xml/vt_docker_quick.xml
+++ b/xml/vt_docker_quick.xml
@@ -328,7 +328,7 @@
    </para>
 <!--taroth 2015-07-02: http://doccomments.provo.novell.com/comments/28724 -->
    <procedure>
-    <title>Enabling the Container Module</title>
+    <title>Enabling the Container Module Using &yast;</title>
     <step>
      <para>
       Start &yast;, and select <menuchoice> <guimenu>Software</guimenu>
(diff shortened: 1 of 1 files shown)


xml/vt_docker_quick.xml
```diff
@@ -328,7 +328,7 @@
    </para>
 <!--taroth 2015-07-02: http://doccomments.provo.novell.com/comments/28724 -->
    <procedure>
-    <title>Enabling the Container Module</title>
+    <title>Enabling the Container Module - interactive way</title>
```
sknorr: Wouldn't the main difference be "from the GUI" versus "from the command line"? [Also, please use Title Case for headlines.]

flavio: It's more about having to perform manual operation (hence interactive) vs no human interaction (non interactive).

You can add the containers module also from the command line using yast2, but that would require human interactions.

I'll fix the title case. Do you have better ideas about the value to put in there?

sknorr: I could try to make the point that ncurses YaST is actually half-graphical, but I won't. It does run on the command line after all.

Maybe `Enabling the Controller Module Using &yast;` vs. `Enabling the Controller Module Directly with SUSEConnect`? That is quite long, though.

flavio: What about `Enabling the Controller Module Using &yast;` and `Enabling the Controller Module Using SUSEConnect` ?

sknorr: Oh, I thought SUSEConnect was always used in the background... (as you can tell, no, I am not a subject matter expert). But sure, it would work without "Directly".

flavio: I think that technically speaking it YaST should use the same set of libraries of SUSEConnect, it shouldn't be invoking SUSEConnect. I proposed these new titles because they seem more concise to me. But as you know it's always a matter of personal taste

flavio:
I updated the code according to what discussed with @sknorr

PR #40, opened 2015-12-04
dns-server, dhcp-server and firewall sections

ancorgs:
Firewall documentation is quite complete. dns-server and dhcp-server are in a minimum-effort basis (documented by examples) since they have way too many options.


schubi2:
Thanks for the docu. LGTM --> merged.

PR #41, opened 2016-01-28
Remove haspcmcia as it's not used by AutoYaST anymore

imobach:
`haspcmcia` is not used since SLE 12 (including 'SP0'). Should I create a PR also against `develop` branch?

Thanks!


sknorr:
I think we can cherry-pick it there. Thanks!

imobach:
Great. Thank you!

PR #42, opened 2016-02-17
Small grammar correction

markkp:
Changed reference from YaST->System to YaST-Software

I just tested the Online Migration module, and it was under Software, not System.


taroth21:
Thanks a lot, Mark!

PR #48, opened 2016-06-07
feature/fate317784: Implement a check and warning to make sure we don't start a Service Pack migration with insufficient disk space

Vogtinator:
This adds a note about the new check for free space in YaST during SP installation and also removes some whitespace.

The reorganization of the "General Preparations" section caused some id changes. Is that ok?

### DIFF
--- a/xml/sle_update_upgrading.xml
+++ b/xml/sle_update_upgrading.xml
@@ -27,7 +27,7 @@ Check with Simona, if this is supported or needed.
         12. No new installation is needed. Existing data, such as home
         and data directories and system configuration, is kept intact.
         You can update from a local CD or DVD drive or from a central
-        network installation source. </para>
+        network installation source.</para>
         <para>This chapter explains how to manually upgrade your &sle;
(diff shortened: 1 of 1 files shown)


xml/sle_update_upgrading.xml
```diff
@@ -369,7 +293,102 @@ tmpfs          506M     0  506M   0% /dev/shm
...
+      secure your data before increasing the available space
+      by resizing partitions, for example.
+
+      There is no general rule regarding how much space each partition
+      should have. Space requirements depend on your particular partitioning
+      profile and the software selected.
+    </para>
+
+    <note>
+      <title>Check for enough Space in &yast;</title>
```
sknorr: I know this one is not your fault... But you should use gerund for verbs in headlines [as long as it makes sense]: "Check*ing* for..."
There are more cases of this.

Vogtinator: I fixed two of them (the ```<sect2/>``` ones), this is the only title that I didn't fix as @fsundermeyer wanted to discuss this tomorrow.

PR #49, opened 2016-06-09
feature/fate319624: Import SSH host keys and configuration

Vogtinator:
Also ran daps-xmlformat and fix duplicated os="sled".


Vogtinator:
I forgot to commit the images...

PR #50, opened 2016-06-10
Clarify configuration location of apache2 instance

Vogtinator:
The first sentence could be understood as if the sysconfig file were the apache configuration before.


fsundermeyer:
Thanks!

PR #52, opened 2016-06-21
feature/fate320269: Apache2 chapter improvements

Vogtinator:


xml/apache2.xml
```diff
@@ -2161,8 +2161,8 @@ sudo cp new.cert.key /etc/apache2/ssl.key/server.key</screen>
      have an officially signed certificate.
     </para>
     <para>
-     The best-known commercial CAs are Comodo
-     (<link xlink:href="https://www.comodo.com"/>) and Symantec
+     The best-known commercial CAs are Thawte
```
sknorr: Thawte is now Symantec (even their logo is "Thawte from Symantec" now). Also, Comodo is now the largest CA, I think. How about adding GoDaddy or DigiCert too?

Vogtinator: > Also, Comodo is now the largest CA, I think.

https://letsencrypt.org//2016/06/23/defending-our-brand.html

> How about adding GoDaddy or DigiCert too?

I'd be fine with those.

Vogtinator: @keichwa agrees with me that that sentence should be removed completely

fsundermeyer:
Hanx!

PR #56, opened 2016-07-12
Fix typo and use other wording

jubalh:
It's CSV not CVS :-)
Also the 'that' sounds not right here, I am not sure whether 'which' is
the best solution though.


cwickert:
Thanks a lot, @jubalh!

PR #60, opened 2016-08-12
Adjusted `Language` option

kobliha:
## Changes
- fixed language -> languages (secondary languages)
- adjusted example for languages (it did not make sense to have en_GB for both pripary and secondary language)
- added note about fixing unknown language by replacing with the default one en_US (bsc#991001)

The profile still validates using `daps`.

### DIFF
--- a/xml/ay_bigfile.xml
+++ b/xml/ay_bigfile.xml
@@ -4139,7 +4139,7 @@ config:type="boolean"&gt;true&lt;/diag&gt;</screen>
      <screen>
        &lt;language&gt;
          &lt;language&gt;en_GB&lt;/language&gt;
-         &lt;languages&gt;de_DE,en_GB&lt;/languages&gt;
+         &lt;languages&gt;de_DE,en_US&lt;/languages&gt;
        &lt;/language&gt;
      </screen>
    </example>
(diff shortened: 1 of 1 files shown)


xml/ay_bigfile.xml
```diff
@@ -4156,6 +4156,11 @@ config:type="boolean"&gt;true&lt;/diag&gt;</screen>
       </tgroup>
    </informaltable>
 
+   <para>
+   If a primary language is unknown, it will be automatically changed into
```
kobliha: Maybe `unknown` -> `erroneous`, decide what you like more

xml/ay_bigfile.xml
```diff
@@ -4156,6 +4156,11 @@ config:type="boolean"&gt;true&lt;/diag&gt;</screen>
...
+   the default one, which is <literal>en_US</literal>.
```
sknorr: I think in the current configuration of the sentence, the "it" is a bit unclear. Would the following work? "If the configured value for the primary language is unknown, it will be reset to the default, <literal>en_US</literal>."

xml/ay_bigfile.xml
sknorr: unknown sounds better to me and also leaves all the possibilities open.

sknorr:
@kobliha Can you take another look here?

xml/ay_bigfile.xml
kobliha: Sure, that would work

kobliha:
@sknorr Text has been adapted

sknorr:
Thanks! Merged.

PR #63, opened 2016-08-30
Add signatures-handling self-update tip

imobachgs:
I've added a tip about signatures-handling during installer self-update.


sknorr:
Thanks. Seemed good to me, so => merged.

imobachgs:
Thanks for reviewing and merging it!

PR #65, opened 2016-09-07
Several updates regarding AutoYaST + network

mchf:


xml/ay_bigfile.xml
```diff
@@ -14429,6 +14420,22 @@ echo -n 100
        <row>
         <entry>
          <para>
+          <literal>autoyast2</literal>
+         </para>
+        </entry>
+        <entry>
+         <para>
+          Location of the control file for automatic installation,
+          see <xref linkend="Commandline.ay"/> for details. Similar to
```
mvidner: The reference is wrong. https://en.opensuse.org/SDB:Linuxrc#p_autoyast2 states that it is Linuxrc parsing the URL and it uses different schemes than (Auto)YaST.

mvidner:
@wfeldt can you please review the factual accuracy of `autoyast=` and `autoyast2=` ?

xml/ay_bigfile.xml
mvidner: See the [rendering of Commandline.ay](https://www.suse.com/documentation/sles-12/singlehtml/book_autoyast/book_autoyast.html#Commandline.ay).

wfeldt:
The differences are

1. autoyast2 uses linuxrc URL syntax; this way it can read and parse the autoyast file
2. because of this, with autoyast2 the network is setup automatically
3. **but:** autoyast2 does **not** support AutoYaST rules and classes

wfeldt:
I would really mention that rules and classes are not supported with autoyast2.

else lgtm

wfeldt:
thanks, lgtm

mvidner:
LGTM

mchf:
@sknorr Could you pls check it? Thanks

sknorr:
Thanks. I hope the functional things were already taken care of. :)

sknorr:
I added 0441abd675837fc3b1688e69be50d68420163676 afterwards to tweak formatting and language a little (the diff is pretty large but that's mostly because there is a small tweak on every line).

I made one content change: I made it more explicit that you only need one linuxrc-configured interface for DHCP auto-configuration to turn off, see [here](https://github.com/SUSE/doc-sle/commit/0441abd675837fc3b1688e69be50d68420163676#diff-4e30d208f3ad62803c41dfe8f9c7dd28R1462).

Please shout if things are wrong now.

PR #68, opened 2016-09-21
Rewrite AutoYaST users section

imobachgs:
This PR was started to document the new [fate#319471](https://fate.suse.com/319471) feature in AutoYaST. However, I've ended up rewriting the whole users section. Now `<groups>`, `<login_settings>` and `<user_defaults>` elements are documented. Moreover several missing options in the `<users>` group have been added.

You can check the result in [w3.suse.de](https://w3.suse.de/~igonzalezsosa/autoyast/handbook/configuration.html#Configuration.Security.users_and_groups).

But I need the review from some docu-team member, please.

### DIFF
--- a/xml/ay_bigfile.xml
+++ b/xml/ay_bigfile.xml
@@ -8984,64 +8984,818 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
   </sect1>
 
 
-  <sect1 xml:id="Configuration.Security.users">
-   <title>Users</title>
+  <sect1 xml:id="Configuration.Security.users_and_groups">
+   <title>Users and Groups</title>
 
    <para>
-    The &rootuser; and at least one normal user can be added during
-    install using data supplied in the control file. User data and passwords
-    (encrypted or in clear text) are part of the
-    <literal>configure</literal> resource in the control file.
+    AutoYaST supports defining local users, groups, special login settings and even
+    default options for new users. Those settings are defined in the following sections:
    </para>
 
-   <para>
-    At least the &rootuser; should be configured during auto-installation
-    so you can login after the installation is finished. It will also ensure
-    nobody else can login to the system (in case the password is not set).
-   </para>
+   <variablelist>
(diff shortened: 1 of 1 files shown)


xml/ay_bigfile.xml
```diff
@@ -8984,64 +8984,818 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
...
+   </variablelist>
 
-   <para>
-    The two users in the following example are added during system
-    configuration.
-   </para>
+   <note>
+    <title>Users and groups set up during the first stage</title>
+    <para>
+     Starting with &productname; 12 SP1, users and groups are set up during the
```
schubi2: I would mention the the openSUSE part too. As far I am remembering  there are special tags for SLES and openSUSE. So we could distinguish between the two cases and produces different descriptions.

xml/ay_bigfile.xml
```diff
@@ -8984,64 +8984,818 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
...
+        </entry>
+        <entry>
+         <para>
+          Number
+         </para>
+<screen>&lt;uid&gt;1001&lt;/uid&gt;</screen>
+        </entry>
+        <entry>
+         <para>
+          Optional. User ID. It must be a unique and non-negative number.
```
schubi2: Good point. We have had an ugly error due equal UIDs  which has costed days to find out .-)

schubi2:
Really nice. Feel free to adapt my suggestions or not :-) LGTM

imobachgs:
@schubi2 Thanks! Sure I'll include your suggestions (now it's even easier as openSUSE is SLE-based).

xml/ay_bigfile.xml
imobachgs: Good point. Even better, I've removed the reference to old behavior.

imobachgs:
@sknorr Please, when you have time, could you have a look? Thanks in advance!

PR #78, opened 2016-11-04
Autoyast: Added bridge configuration example.

teclator:


xml/ay_bigfile.xml
```diff
@@ -5241,6 +5241,37 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
 &lt;/networking&gt;</screen>
    </example>
 
+   <example>
+     <title>Bridge interface configuration</title>
```
sknorr: Please use title style: `Bridge Interface Configuration`

xml/ay_bigfile.xml
```diff
@@ -5241,6 +5241,37 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
 &lt;/networking&gt;</screen>
    </example>
 
+   <example>
```
sknorr: It is generally a good idea to add `xml:id`s to formal elements that have a title (procedure, example, figure, ...)

teclator:
@sknorr by now we will just add the example for autoyast, let see if deserve to add something more in the future, updated the PR according to your comments

sknorr:
Thanks!

PR #79, opened 2016-11-23
Life patching lifecycle, https://fate.suse.com/322049

sven15:
As this is my first sect1, feedback and comments are very welcome.

### DIFF
--- a/xml/kgraft.xml
+++ b/xml/kgraft.xml
@@ -258,16 +258,10 @@
  <sect1 xml:id="sec.kgraft.lifecycle">
   <title>Patch Lifecycle</title>
   <para>Expiration dates of live patches can be accessed with <command>zypper
-  lifecycle</command>.</para>
-  <procedure>
-   <step>
-    <para>Make sure that the package <package>lifecycle-data-sle-live-patching
-    </package> is installed.</para>
-   </step>
-   <step>
-    <para>Use <command>zypper lifecycle</command> to list the lifecycles
-    of installed products or packages, including live patches.</para>
-    <screen>&prompt.user;<command>zypper lifecycle</command>
+   lifecycle</command>. Make sure that the package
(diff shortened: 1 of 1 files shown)


sven15:
[Release lifecycle data for Live Patching](https://trello.com/c/amYXmmxA/392-release-lifecycle-data-for-live-patching)

xml/kgraft.xml
```diff
@@ -255,6 +255,40 @@
    </procedure>
   </sect2>
  </sect1>
+ <sect1 xml:id="sec.kgraft.lifecycle">
+  <title>Patch Lifecycle</title>
+  <para>Expiration dates of life patches can be accessed with <command>zypper
```
sknorr: life -> live

xml/kgraft.xml
```diff
@@ -255,6 +255,40 @@
...
+  lifecycle</command>.</para>
+  <procedure>
+   <step>
+    <para>Make sure that the package <systemitem class="resource">
```
sknorr: Use `<package>` element instead of `<systemitem>` (at least as long as you are using DocBook 5, in DocBook 4 we indeed needed to use this syntax).

sknorr:
In commit messages referencing FATEs, use the short syntax FATE#nnnn.

sven15:
yeah, probably the liFecycle messed me up. Thank you @sknorr

sknorr:
Almost wanted to merge this ... but then: does it really make sense to use a procedure here?

Fundamentally, only the second step is really a step in the narrower sense of the word.

The first one is a prerequisite which can go into a simple para. The third one is extra information that does not require a step either and definitely should go into a simple para. Which means the second step is now alone... -> para + screen.

sven15:
Yes you're right. I just followed the style of the preceding and following section, but not using a procedure looks better.

PR #86, opened 2016-12-01
stated that self_update is a boot parameter

sven15:
Made it clear that self_update for YaST is a boot parameter. The link for further details does not make that fact clear immediately.


sven15:
[Automatic Update of YaST at beginning of the installation](https://trello.com/c/XIGmIxwS/211-automatic-update-of-yast-at-beginning-of-the-installation)

imobachgs:
Thanks!

PR #87, opened 2016-12-06
zSystem requires Upgrade=1 parameter for upgrade

sven15:
All update paths for zSystems require a special kernel parameter. Added a note in the "Upgrading SUSE Linux Enterprise" chapter and added the parameter in the parmfile section.

https://bugzilla.suse.com/show_bug.cgi?id=1010771&GoAheadAndLogIn=1

### DIFF
--- a/xml/deployment_prep_zseries.xml
+++ b/xml/deployment_prep_zseries.xml
@@ -116,6 +116,16 @@
       </para>
      </listitem>
     </varlistentry>
+    <varlistentry>
+     <term><command>Upgrade=&lt;0|1&gt;</command></term>
+     <listitem>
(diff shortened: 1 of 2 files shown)


xml/deployment_prep_zseries.xml
```diff
@@ -116,6 +116,16 @@
       </para>
      </listitem>
     </varlistentry>
+    <varlistentry>
+     <term><command>Upgrade=&lt;0|1&gt;</command></term>
+     <listitem>
+      <para>If you want to upgrade your &sle;, specify
```
sknorr: Remove "if you want" -> unnecessarily wordy.

PR #92, opened 2017-01-02
Mount Points with FCoE or iSCSI

sven15:
/usr and /var are not supported mount points for iSCSI and FCoE

https://bugzilla.suse.com/show_bug.cgi?id=1010297
uses text from https://bugzilla.suse.com/show_bug.cgi?id=1012335#c28

### DIFF
--- a/xml/storage_fcoe.xml
+++ b/xml/storage_fcoe.xml
@@ -106,6 +106,17 @@
     </imageobject>
    </mediaobject>
   </informalfigure>
+  <note>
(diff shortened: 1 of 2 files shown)


sven15:
[add NOTE: that separate /var and /usr partitions not supported for FCoE network boot](https://trello.com/c/47Tv0uBD/389-add-note-that-separate-var-and-usr-partitions-not-supported-for-fcoe-network-boot)

sven15:
Same text as in Pull Reqest #93

PR #93, opened 2017-01-03
system mount points for FCoE and iSCSI in deployment guide

sven15:
bsc#1010297 and bsc#1012335#c28

### DIFF
--- a/xml/inst_yast2.xml
+++ b/xml/inst_yast2.xml
@@ -2392,7 +2392,17 @@ sle_update.xml#sec.update.terminology: "Modules" -->
     </listitem>
    </itemizedlist>
   </note>
-
+  <note>
+   <title>Mount Points for FCoE and iSCSI Devices</title>
+   <para>FCoE and iSCSI devices will appear asynchronously during the
+    boot process. While the initrd guarantees that those devices are
+    setup correctly for the root file system, there are no such
+    guarantees for any other file systems or mount points like 
+    <filename>/usr</filename>. Hence any system mount points like 
+    <filename>/usr</filename> or <filename>/var</filename> are not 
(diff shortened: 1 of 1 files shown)


sven15:
[add NOTE: that separate /var and /usr partitions not supported for FCoE network boot](https://trello.com/c/47Tv0uBD/389-add-note-that-separate-var-and-usr-partitions-not-supported-for-fcoe-network-boot)

xml/inst_yast2.xml
```diff
@@ -2392,7 +2392,17 @@ sle_update.xml#sec.update.terminology: "Modules" -->
     </listitem>
    </itemizedlist>
   </note>
-
+  <note>
+   <title>Mount Points for FCoE and iSCSI Devices</title>
+   <para>FCoE and iSCSI devices will appear asynchronously during the
+    boot process. While the initrd guarantees that those devices are
+    setup correctly for the root file system, there are no such
```
sknorr: style: setup -> set up (this is a verb)

xml/inst_yast2.xml
```diff
@@ -2392,7 +2392,17 @@ sle_update.xml#sec.update.terminology: "Modules" -->
...
+    guarantees for any other file systems or mount points like 
+    <filename>/usr</filename>. Hence any system mount points like 
+    <filename>/usr</filename> or <filename>/var</filename> are not 
+    supported. If you want to use those devices, ensure correct
+    synchronisation with the respective services and devices.</para>
```
sknorr: Sounds a bit circular: "to use devices ... ensure sync with devices." - not sure here.

sven15: Yes, an "of" would be better.

PR #94, opened 2017-01-03
Updates are now referencing SLE12 SP2 instead of SP1, bsc#1016464

sven15:
The docu was still "describing" the update to SP1 instead of SP2.
Fix for https://bugzilla.novell.com/show_bug.cgi?id=1016464


sven15:
[SLES12SP2 docu: Upgrading to SLES 12 SP2 shows SP1 only](https://trello.com/c/zkG7SPbF/396-sles12sp2-docu-upgrading-to-sles-12-sp2-shows-sp1-only)

PR #98, opened 2017-01-16
Added Warning about Regular Expressions in multipath.conf

sven15:
* replaced IDE with SATA
* added warning about regex in multipath.conf


sven15:
[multipath-tools parse error](https://trello.com/c/P5kynSsN/390-multipath-tools-parse-error)

PR #99, opened 2017-01-20
Generate POT files from XML files

guoyunhe:
UPDATE:

1. Use itstool by GNOME project.
2. Generate single pot file for all xml document files.

---

This effort try to convert DocBook XML to GNU Gettext POT and take full advantage of Weblate (http://l10n.opensuse.org/) for community localization teams.

poxml (kde-l10n-devel), developed by KDE project, was used for the task. Generated POT files are generally good for using.

The only mistake is that a docmanager tag will be imported.

```po
#. Tag: dm:docmanager
#: zypper.xml:14
#, no-c-format
msgid "<dm:bugtracker></dm:bugtracker> <dm:translation>yes</dm:translation>"
msgstr ""
```

I can translate po files in a local environment without any issues, using Lokalize software.

spectacle lh4246

I add translation to develop branch because it would support all future openSUSE releases.

Old translations in XML can be imported by `split2po`.

How do you think?


sknorr:
Can you also provide a makefile (or at least a textual description of what you did)? Thanks!

guoyunhe:
Haven't try to build XML back. But I will do and tell you result.

I can write makefile if everything works as expected.

sknorr:
(fyi, I have just opened https://github.com/openSUSE/release-notes-openSUSE/issues/54 and will investigate whether this makes sense for the Leap 42.3 release notes.)

guoyunhe:
I can get translated XML with following command:

```
po2xml xml/zypper.xml l10n/zh_CN/zypper.po | xmllint --format - > l10n/zh_CN/zypper.xml
```

Raw XML generated by po2xml is very ugly. But after xmllint formating, it looks much better. It does not have the same spaces and line breaks as original XML. But I think they would generate the same HTML, PDF or other readable document.

spectacle bz4246

guoyunhe:
Now here are two scripts to convert between po and xml.

xml2pot.sh

```bash
#!/bin/bash
#
# Convert DocBook XML files to GNU Gettext POT files.
#
# POT files will be imported to Weblate http://l10n.opensuse.org/
# Weblate will sync POT and PO files automatically.
#

for xml in xml/*.xml;
do
    tmp=${xml#xml/}
    name=${tmp%.xml}
    echo $name
    xml2pot $xml > l10n/templates/$name.pot
done

```

po2xml.sh

```bash
#!/bin/bash
#
# Convert GNU Gettext POT files to DocBook XML files.
#
# PO files will be downloaded from Weblate http://l10n.opensuse.org/
# Weblate will sync POT and PO files automatically.
#

for xml in xml/*.xml;
do
    tmp=${xml#xml/}
    name=${tmp%.xml}
    echo $name
    for po in $(find l10n -name "$name.po");
    do
        langdir=${po%$name.po}
        mkdir -p $langdir/xml
        po2xml $xml $po | xmllint --format - > $langdir/xml/$name.xml
    done
done
```

guoyunhe:
Makefile done.

I am not very familiar with GNU Make. However, the makefile works as expected. Language directories are matched manually, that is why here is a very very long list of po --> xml matches. Watched languages are the same with openSUSE release notes and Weblate.

`xml2pot` will throw a segmentation fault when processing `troubleshooting.xml`. So I filter it out in makefile. Maybe it is too long to parse or xml2pot has some bug. Anyway, other XML files can be converted nicely.

guoyunhe:
Thanks @GroverChouT for shorten the Makefile.

guoyunhe:
@sknorr Could you review the pull request and see if here is something to be done?

sknorr:
Very sorry. I have done some things, but did not get everything to work properly yet: https://github.com/SUSE/doc-sle/tree/feature/trans-mf

svenseeberg:
We stumbled upon issues with the xml2po and po2xml tools. They are not working as expected ans xml2po currently only works with Python2. However, Python2 will not be supported with the next SLE and openSUSE releases. The other issue is that xml2po drops namespaces from tags. We want to enable translations but we need to tackle the tooling issues first.

guoyunhe:
Thanks for the update!

I understand your concern. I could contact the author of this tool to fix the problem. Or we can fork it once I get time.

svenseeberg:
Yes, sure. You're welcome to do so. At some point we need a working conversion tool that works with upcoming (open)SUSE releases. We are also internally discussing the option of spending some resources for bringing the xml2po script to Python3. One of our colleagues already started with the conversion: https://github.com/tomschr/xml2po-ng

sknorr:
So, it seems like we might be able to use itstool (packaged under the same name) -- we have not yet checked it very thoroughly but it seems the most modern and promising of all the XML->PO tools. Which would then mean, we'd "only" need to sort out the issues in the Makefile.

GroverChouT:
Hi all. Our Chinese wiki and translation team found that Installation page on English wiki just simply refer to doc.o.o, which means we can't go further without this works as Guo did try to help a few months ago. Our community wondering how soon can we start to translate this documents. We believe it will help us to attract more new users to use openSUSE.

svenseeberg:
I cannot promise anything right now, but I'll try to package the itstool and include it in the openSUSE Documentation:Tools repository. Then we can take care of the makefile.

guoyunhe:
Updated Makefile with itstool.

Gianvacca:
Hi @guoyunhe ,
as anticipated via mail, I would rather prefer a multi-pot approach.
Otherwise, I have only praises for your work.
Alessio

svenseeberg:
Thank you all for contributing. We continue the work and discussion in #358

PR #110, opened 2017-03-23
Fix mis-identified parameter for qcow2 images used in

bfrogers:
Virtualization Best Practices doc. 12-cache-size (twelve)
should instead read l2-cache-size (ell two, as in level 2).

Signed-off-by: Bruce Rogers <brogers@suse.com>


bfrogers:
Let me re-do the description a bit.

PR #111, opened 2017-03-23
Fix mis-identified parameter for qcow2 images

bfrogers:
The Virtualization Best Practices doc has a mis-identified
qcow2 image parameter. The string 12-cache-size (twelve)
should instead read l2-cache-size (ell two, as in level 2).

Signed-off-by: Bruce Rogers <brogers@suse.com>


fsundermeyer:
Thank you very much!

PR #115, opened 2017-04-07
added salt module

schubi2:


imobachgs:
LGTM. Let's wait for the documentation team. Thanks!

sknorr:
Are you sure that abbreviating "Configuration Management System" as CMS is a good idea? Personally, I find it a bit dangerous to redefine an abbreviation that is very widely used for "Content Management System".

Also see https://en.wikipedia.org/wiki/CMS#Computing which provides multiple resolutions of the abbreviation but does not mention "Configuration Management System" (yet).

imobachgs:
Good point. Yes, we probably need to find a better acronym. In the section you mention, I can read:

*Configuration management system, now known as software configuration management*

So what about SCM? The problem is that it's also the acronym for *source control management*. With a three-letters acronym we have a high chance to collide :)

sknorr:
Can you avoid using an acronym here? I know that makes some sentences longer but still... maybe just "Configuration Management" would suffice?
Afaics, this is what our SUSE Manager docs are using ( https://github.com/SUSE/doc-susemanager ).

sknorr:
Thanks!

PR #117, opened 2017-04-10
Revamped Upgrade Chapters

sven15:
* sorted all methods into "Offline" and "Online" methods


xml/sle_update_upgrading.xml
```diff
@@ -621,583 +638,90 @@ tmpfs          506M     0  506M   0% /dev/shm
...
+     <para>
+      Boot DVD, flash disk, ISO image, &ay;, <quote>plain RPM</quote> and
+      third-party tools
+     </para>
+    </listitem>
+   </varlistentry>
+  </variablelist>
+  <para>
+   Upgrading from &sle; 11 SP4 to &sle; 12 SP3 is only supported via
+   an offline upgrade.
```
tomschr: I think there is a missing transition. Could you add an additional text to lead over the following itemizedlist? I think it would be enough to amend the text with: "... an offline upgrade using one of the following methods:"

PR #121, opened 2017-04-12
Speed up and clean up .travis.yml

tomschr:
* Download Zip from docbook.org
* Cache entities
* Use yakkety distribution

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: bash
 env:
       global:
       - LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
+      - CACHE=$HOME/.cache/entities
 
 dist: yakkety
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,17 +1,33 @@
...
 
 script:
-    - xmllint --noout --noent xml/*.xml
+    - test -e $HOME/.cache/entities || mkdir -p $HOME/.cache/entities
+    - wget -P $HOME/.cache/entities http://docbook.org/xml/4.5/docbook-xml-4.5.zip
+    - unzip -d $HOME/.cache/entities/ $HOME/.cache/entities/docbook-xml-4.5.zip
+    - cp -vi $HOME/.cache/entities/*.mod xml/
+    - cp -avi $HOME/.cache/entities/ent xml/
+    # - wget -P xml/ http://www.oasis-open.org/docbook/xml/4.5/dbcentx.mod
+    - sed -i 's#dbcent PUBLIC .*>#dbcent SYSTEM "dbcentx.mod">#' xml/entity-decl.ent
```
sknorr: You can make this a bit safer by using `[^>]*`:

```
sed -i 's#dbcent PUBLIC [^>]*>#dbcent SYSTEM "dbcentx.mod">#' xml/entity-decl.ent
```

Otherwise, if you have another tag on the same line as this one, the second tag might be replaced too.

.travis.yml
```diff
@@ -1,17 +1,31 @@
...
     - echo $TRAVIS_COMMIT
-    - ls xml/*.xml
     - xmllint --version
 
 script:
-    - xmllint --noout --noent xml/*.xml
+    - test -e $CACHE || ( mkdir -p $CACHE; wget -P $CACHE http://docbook.org/xml/4.5/docbook-xml-4.5.zip; unzip -d $CACHE $CACHE/docbook-xml-4.5.zip )
+    - cp -vi $CACHE/*.mod xml/
+    - cp -avi $CACHE/ent xml/
+    - sed -i 's#dbcent PUBLIC .*>#dbcent SYSTEM "dbcentx.mod">#' xml/entity-decl.ent
```
sknorr: You can make this a bit safer by using `[^>]*`:

```
sed -i 's#dbcent PUBLIC [^>]*>#dbcent SYSTEM "dbcentx.mod">#' xml/entity-decl.ent
```

Otherwise, if you have another tag on the same line as this one, the second tag might be replaced too.

tomschr: Thanks! Great idea Stefan! 👍 

After thinking a bit more, the whole sed thing is quite unsafe in general. That's the nature if you try to manipulate XML stuff. For example, if someone adds a line break between `PUBLIC` or between the public identifier and the URI, all hell break loose...

Wondering if I should switch back to catalogs...

PR #126, opened 2017-04-18
Z requires NPIV enabled for automatic LUN scanning, fate#320750

sven15:
Based on comment #16 of FATE#320750


sven15:
[Document YaST zfcp interface on z Systems requirements](https://trello.com/c/QiXD8toc/143-document-yast-zfcp-interface-on-z-systems-requirements)

sknorr:
Looks good but seems like it might need a doc update?

PR #134, opened 2017-05-08
MPIO: default path_checker is tur, fate#320804

svenseeberg:


sven15:
[multipath: make TUR the default checker](https://trello.com/c/VNmyWS3I/161-multipath-make-tur-the-default-checker)

PR #136, opened 2017-05-08
vt best practices: Clarify note about qemu binary

jfehlig:
Section 3.2 recommends using qemu-system-x86_64 over qemu-system-i386.
This is a fine recommendation for KVM, but the opposite is true for
Xen. Clarify the section by adding notes about the preferred qemu
binary for both KVM and Xen.

Note: This change is the result of user comment in bugzilla

https://bugzilla.suse.com/show_bug.cgi?id=1037985#c4


fsundermeyer:
Thanks a lot!

PR #150, opened 2017-07-12
Fix all the docker build instructions

flavio:
Do not use the zypper's `gpg-auto-import-keys` flag when installing software, this is unneeded and causes security issues.

Fixes bsc#1048279

Signed-off-by: Flavio Castelli <fcastelli@suse.com>


cyphar:
LGTM.

fsundermeyer:
Thanks!

PR #152, opened 2017-07-12
unique UID in user settings

schubi2:


xml/ay_bigfile.xml
```diff
@@ -8958,6 +8958,68 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
 &lt;/users&gt;</screen>
     </example>
 
+    <para>
+     The next example shows a more complete example. The data in
+     <filename>/etc/default/useradd</filename> is used to determine the home
+     directory of the user to be created plus other parameters.
```
fsundermeyer: The following example shows a more complex scenario. System wide default setting from <filename>/etc/default/useradd</filename>, such as the shell or the parent directory for the home directory, are applied.

xml/ay_bigfile.xml
```diff
@@ -8958,6 +8958,68 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
...
+    </example>
+
+    <note>
+     <title><literal>authorized_keys</literal> file will be overwritten</title>
+     <para>
+      If the profile defines a set of SSH authorized keys for a user, the
+      content of the <literal>$HOME/.ssh/authorized_keys</literal> file will be
+      overwritten when using AutoYaST to perform a system upgrade or
+      configuration.
+     </para>
```
fsundermeyer: &lt;para&gt;
 If the profile defines a set of SSH authorized keys for a user in the <tag>authorized_keys</tag section, 
 an existing <filename>$HOME/.ssh/authorized_keys</filename> file will be overwritten. If not existing, the
 file will be created with the content specified. Avoid overwriteng an existing <filename>authorized_keys</filename> 
 by not specifying thev respective section in the &ay; control file. 
&lt;/para&gt;

xml/ay_bigfile.xml
```diff
@@ -8958,6 +8958,68 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
...
+     <para>
+      To avoid such behavior, the <literal>&lt;authorized_keys&gt;</literal> section
+      should be omitted.
+     </para>
+    </note>
```
fsundermeyer: ~~&lt;para&gt;
    To avoid such behavior, the <literal>&lt;authorized_keys&gt;</literal> section
     should be omitted.
&lt;/para&gt~~

xml/ay_bigfile.xml
```diff
@@ -8958,6 +8958,68 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
...
+    <note>
+     <title><literal>uid</literal> should be be a unique</title>
+     <para>
+      Either each user <literal>uid</literal> must be defined and should be
+      unique, or none <literal>uid</literal> will be defined in the &ay;
+      control file. In that case &ay; evalutes the UIDs
+      of each user. Otherwise it could be that UIDs are not unique on the
+      system which is not an error in general but could cause some problems
+      in other areas.
+     </para>
```
fsundermeyer: This section is unclear. I assume that you wanted to express something like:

* IF user IDs are specified, they should be unique
* to let the system automatically allocate a user ID, do not specify a uid

If correct, I suggest the following text:

&lt;note&gt;
 &lt;title&gt;User IDs should be Unique&lt;/title&gt;
 &lt;para&gt;
   User IDs should be unique throughout the system&mdash;although not manadatory, using the
   same ID for different users may lead to unforseen problems. Therefore, when specifying a <tag>uid</tag>
   for a user in the control file, amke sure it is not already in use. Alternatively, do not specify a <tag>uid</tag>
  in the control file. In this case, the system will automatically allocate a unique user ID.
 &lt;/para&gt;
&lt;/note&gt;

xml/ay_bigfile.xml
```diff
@@ -8958,6 +8958,68 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
...
+    &lt;password_settings&gt;
+      &lt;max&gt;120&lt;/max&gt;
+      &lt;inact&gt;5&lt;/inact&gt;
+    &lt;/password_settings&gt;
+  &lt;/user&gt;
+&lt;/users&gt;</screen>
+    </example>
+
+    <note>
+     <title><literal>authorized_keys</literal> file will be overwritten</title>
```
fsundermeyer: &lt;title>&lt;literal>authorized_keys&lt;/literal> File will be Overwritten&lt;/title>

xml/ay_bigfile.xml
schubi2: hm, not exactly. The problem is if there are entries with IDs and there are entries without IDs too. So YaST evaluates IDs for the not defined entries automatically which can be the same as the IDs which have already been defined in other entries.
So mixing user entries which have and have not IDs is not an good idea :-)

schubi2: I have talked with Frank. I can merge this PR and he will make additional changes by his own.
Thanks Frank !

PR #155, opened 2017-08-11
vt-doc: Add section on using macvtap

jfehlig:
Add a section to the libvirt documentation on using macvtap
to share a host network interface.


jfehlig:
Hmm, I've seemed to have messed up my branch. I'll open another request after fixing it.

PR #158, opened 2017-08-24
Add link to online release notes

cjschroder:
https://bugzilla.suse.com/show_bug.cgi?id=1045370

### DIFF
--- a/xml/opensuse_update.xml
+++ b/xml/opensuse_update.xml
@@ -43,6 +43,11 @@
     &opensuse; Tumbleweed</link>.
   </para>
  </important>
+ <para>
+The release notes are bundled in the installer, and you may also read them online 
+at <link xlink:href="https://doc.opensuse.org/release-notes/x86_64/openSUSE/Leap/
+42.3/#upgrade">openSUSE Leap 42.3 Release Notes</link>.
+ </para>
   <para>
    Software tends to <quote>grow</quote> from version to version. Therefore,
    take a look at the available partition space with <command>df</command>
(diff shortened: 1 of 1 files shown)


xml/opensuse_update.xml
```diff
@@ -43,6 +43,11 @@
     &opensuse; Tumbleweed</link>.
   </para>
  </important>
+ <para>
+The release notes are bundled in the installer, and you may also read them online 
+at <link xlink:href="https://doc.opensuse.org/release-notes/x86_64/openSUSE/Leap/
+42.3/#upgrade">openSUSE Leap 42.3 Release Notes</link>.
```
okurz: I guess that should be `<productnumber>&productnumber;</productnumber>` instead of "openSUSE Leap 42.3" so that this needs to be updated every year, right? But I don't know how the URL could be parameterized then. Probably also the paragraph needs to be indented like the surrounding paragraphs?

sknorr: The URL can be parameterized the same way, with `&productnumber;`. Alternatively, you can also define something like `&rn-url-opensuse;`=`https://doc.opensuse.org/release-notes/x86_64/openSUSE/Leap/&productnumber;`.
However, using entities in URLs can be dangerous -- someone might go *cr*az**y** with the next version and use e.g. `15_0 `in the URL instead of `15.0`). The second version somewhat mitigates the issue, because then, you at least have a chance of seeing the evil ways in which your `&productname;` entity is being used even when editing the entities file only.

cjschroder:
Thanks! Indentation is corrected, and I changed the external link to the general release notes page to avoid versioning drama.

PR #164, opened 2017-10-03
vt-doc: replace obsolete data-plane with IOThreads

jfehlig:
The experimental x-data-plane option was removed in qemu 2.5
and replaced with the more generic IOThreads. The option no
longer works with SLES12 >= SP2 and should be removed from the
virtualization docs. Replace it with information on configuring
its successor, IOThreads.

### DIFF
--- a/xml/kvm_support_status.xml
+++ b/xml/kvm_support_status.xml
@@ -314,17 +314,6 @@
     </listitem>
    </varlistentry>
    <varlistentry condition="kvm4x86">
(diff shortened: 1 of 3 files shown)


svenseeberg:
@jfehlig we need to update the SLE12SP3 documentation as well?

jfehlig:
Yes, SLES12 SP2 and SP3 both need updated.

svenseeberg:
I pulled your commit into a new branch and fixed the indentation. See #191

PR #170, opened 2017-11-03
[PoC] Use Docker at Travis

lslezak:
# Proof of Concept

I have noticed  that for Travis builds you use a [travis.sh](https://raw.githubusercontent.com/openSUSE/doc-ci/master/travis/travis.sh) build script. But that

- Does not use `daps` as described in the README
- Uses `docbook-xml-4.5.zip` while the README mentions DocBook 5
- Uses `xmllint` for validation while `daps` uses `jing`
- Uses Ubuntu 14.04, does not use packages from `Documentation:Tools` as suggested in README

I do not know if using simple `xmllint` is just fine for you or if you would like to run full `daps` which to me seems the recommended way.

I have some experience with Travis in YaST as we need to run the YaST tests in openSUSE distribution (Ubuntu is too different for us) so I have prepared this PoC. We run the tests inside a Docker container which contains an openSUSE system.

## Result

- Proposed `Dockerfile` to build the Docker image with the needed tools
  - Is based on openSUSE Tumbleweed (if that's too on the edge for you them switching to e.g. 42.3 is trivial)
  - Uses packages from the Documentation:Tools repository
- Updated `.travis.yml` which runs `daps` in Docker
- See the [example Travis log](https://travis-ci.org/lslezak/doc-sle/builds/296328304#L2541) with used `daps`, you can see the `daps` output there

This is just a proof of concept to collect your opinions about this approach. There are some issues with this approach. The most important one is probably the fact that currently the build takes more than 11 minutes. I do not know how often you change the documentation but this seems quite a lot to me.

## Issues

The very long build time can be split to two parts:

1. Building the Docker image (including `daps`) (takes about 4.5 minutes)
2. Running the tests (takes about 6.5 minutes)

#### The Docker Image

The Docker image can be pre-built, in YaST we build the image in Docker Hub (see e.g. the [yastdevel/ruby](https://hub.docker.com/r/yastdevel/ruby/) image).  Downloading the YaST images takes about 25 seconds in Travis, I think we could get similar numbers for the documentation image as well.

#### Running Tests

The `daps` validation is quite slow and there are almost 40 books, validating all of them takes almost 7 minutes. But we could run the checks in parallel (in separate Travis jobs).  With 3 parallel jobs the test run time could be reduced to about 2 minutes.

Pseudocode:

```
groups = 3
lines = `find . -maxdepth 1 -name 'DC-*' | wc -l`
max_lines = lines / 3
find . -maxdepth 1 -name 'DC-*' | sort | split -l $max_lines -d - TRAVIS_GROUP_
```

This results in `TRAVIS_GROUP_00`, `TRAVIS_GROUP_01` and `TRAVIS_GROUP_02` files which lists the files to process in each Travis job.

See the [.travis.yml](https://github.com/openSUSE/snapper/blob/master/.travis.yml) file in snapper where we build the package for 5 different distributions in parallel.

# Conclusion

I'm interested in your feedback about this approach. If it looks interesting for you and you would like to change the Travis build just let me know, I can help you with that.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.git/
+build/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,12 @@
+sudo: required
 language: bash
-env:
-      global:
-      - LD_PRELOAD=/lib/x86_64-linux-gnu/libSegFault.so
-      - CACHE="$HOME/.cache/entities"
-      - XML_CATALOG_FILES="catalog.xml"
-      - GITHUB_URL="https://raw.githubusercontent.com/openSUSE/doc-ci/master/travis/travis.sh"
+services:
+  - docker
 
-dist: yakkety
-
-cache:
-  directories:
-    - $CACHE
-
-addons:
-  apt:
-    packages:
-    - libxml2-utils
-    - libxml2-dev
-    - libxml2
-
-before_script:
-    - cat /etc/os-release
-    - echo $TRAVIS_COMMIT
-    - sudo apt-get install -y libxml2-utils
-    - xmllint --version
-    - wget $GITHUB_URL && chmod +x travis.sh
+before_install:
+  - docker build -t sle-doc-image .
+  # list the installed packages (just for easier debugging)
+  - docker run --rm -it sle-doc-image rpm -qa | sort
 
 script:
-    - ./travis.sh xml/*.xml
-
-branches:
-    except:
-        - trans/*
-
+  - docker run --rm -it sle-doc-image find . -maxdepth 1 -name 'DC-*' -exec echo "Validating {}..." \; -exec daps -d \{\} validate --remarks \;
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,7 @@ services:
   - docker
 
 before_install:
-  - docker build -t sle-doc-image .
-  # list the installed packages (just for easier debugging)
-  - docker run --rm -it sle-doc-image rpm -qa | sort
+  - docker pull svenseeberg/doc_ci
 
 script:
-  - docker run --rm -it sle-doc-image find . -maxdepth 1 -name 'DC-*' -exec echo "Validating {}..." \; -exec daps -d \{\} validate --remarks \;
+- docker run --rm -it svenseeberg/doc_ci find . -maxdepth 1 -name 'DC-*-all' -exec echo "Validating {}..." \; -exec daps -d \{\} validate --remarks \;
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,24 @@
+FROM opensuse:tumbleweed
+
+# "zypper dup" synchronizes with the current Tumbleweed (even downgrades if needed),
+# we need to install curl for downloading/installing the GPG key
+# Note: we can keep the metadata cache as this image is only used once at Travis
+RUN zypper --non-interactive dup --no-recommends && \
+  zypper --non-interactive in --no-recommends curl
+
+# import the Documentation:Tools GPG key
+RUN rpm --import https://build.opensuse.org/projects/Documentation:Tools/public_key
+
+RUN zypper ar -f https://download.opensuse.org/repositories/Documentation:/Tools/openSUSE_Tumbleweed/ Documentation:Tools
+
+# Note: unfortunately this pulls in a *lot* of not needed packages like samba, texlive
+# which are not needed for validation, but when we lock them the install command
+# below fails in non-interactive mode :-(
+# RUN zypper addlock texlive-* samba-* ghostscript cups-libs transfig *-fonts mozilla-*
+# saxon6 is required by jing but it's not installed automatically via dependencies
+RUN zypper --non-interactive in --no-recommends daps saxon6
+
+# copy the sources
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+COPY . /usr/src/app


sknorr:
Hi Ladislav! Thanks for the effort -- to answer some of your questions:

1. Ubuntu -- I think that was mostly an issue of effort involved. Whenever we in this team have used Travis, it was always with the default setting of Ubuntu, and @tomschr probably either did not know openSUSE was a possibility or thought it was too much effort.

2. Usage of xmllint -- Currently, we don't validate the documents at all -- we just check for well-formedness. xmllint is competent enough to check for well-formedness and (most likely) faster at that than jing because it is written in C and not Java. Note that the individual XML files are not always valid -- our XML becomes valid after profiling (kicking out document parts that do not apply). Therefore, if you want to validate, you have to use use DAPS and the DC files instead of going file by file. Personally, I like the thought of always checking for validity, but I think not everyone on our team thinks that all content has to be kept valid at all times.

3. Not using DAPS -- I think that decision was taken because of the issues getting DAPS running on Ubuntu (which you solved).

svenseeberg:
Wow, this is great! Stefan is right with the reasons why did not use daps yet. Using docker is, of course, a nice solution. I'll look into improving the running time. But we could already conditionally use this for PRs.

lslezak:
> @tomschr probably either did not know openSUSE was a possibility or thought it was too much effort

Actually we started with Ubuntu in YaST as well, the Docker feature is relatively new in Travis.

> it should be enough to check just DC-SLES-all, DC-SLED-all and DC-opensuse-all to ensure validity of everything we have

It's up to you to define the requirements, I do not know which parts make sense for you so I started with checking everything. If only a subset is important then it's an easy way how to make it faster...

> The situation with transfig and its TeX dependencies is indeed sad and something we need to solve in general at some point.

That could be "fixed" by some hack when building the Docker image at Docker Hub (like running `rpm -e --force --nodeps <not_needed_packages>` at the end).

svenseeberg:
Well, we definitely want to use this ;-) How much effort is it to pre-build the docker container? And instead of DC-* just use a DC-*-all. Maybe the improvements are good enough to use it.

lslezak:
### Building a Docker container

1. Create an account at https://hub.docker.com/
1. Create an organization there, add some colleagues so if something goes wrong there are more people able to fix that
1. Create a GitHub repo hosting the image definition, see e.g. https://github.com/yast/docker-yast-ruby add the `Dockerfile` there
1. You can even enable Travis for the Docker Git repo, see https://github.com/yast/docker-yast-ruby/blob/master/.travis.yml, you test the changes there as well
1. Create image at https://hub.docker.com/ in your organization, link the GitHub repo
1. Update the `Dockerfile` here to use the new image

Create the accounts/images, I can help you later with tweaking the `Dockerfile`.

tomschr:
This is a very interesting idea! Thanks @lslezak for your time and effort! 👍

Stefan already described the important things so I won't repeat them. 😉 My only concern is the time from submit to result. I think 10min is a bit too much just for validation. However, as Stefan and you already pointed out, this could be reduced. Maybe even further with some other, clever modifications.

Apart from the idea itself, I would suggest to make all changes in the [openSUSE/doc-ci](https://github.com/openSUSE/doc-ci) repository, not in SUSE/doc-sle.

The doc-ci repo was created just to have a single, dedicated place to organize all CI related things for documentation. The idea was to have consistency so we don't have to repeat again. Every change will be propagated to all the (doc) repos that are "connected" (currently only doc-sle).
This makes it (hopefully) easier. Of course, if we make any errors in doc-ci, this will affect *all* doc repos. 😉 

At the moment, the well-formedness check is only active on the doc-sle repository. This was on purpose to gain experiences (good or bad).

lslezak:
> I think 10min is a bit too much just for validation.

It could be reduced quite a lot, `daps 'DC-*-all'` takes about 17 seconds on my machine (with hot cache, the first run took ~25 seconds). Running in parallel in background (via '&') it takes about 8 seconds (that's the time of the longest job with hot cache). Travis has less CPUs, according to [this](https://docs.travis-ci.com/user/reference/overview/) 2 CPUs are available, but that still should help a bit.

The Docker image download takes about 25 seconds for the YaST images ([here](https://travis-ci.org/yast/yast-yast2#L461) it was 23).

**So I think the whole build time could be realistically reduced to less than a minute.** And that's reasonably fast, the YaST builds usually take 1-4 minutes (depending on the code coverage of the module).

> This needs some more time and experiments (maybe Hackweek?).

JFYI: I'll be working on something else, but I can provide help and support for anybody interested in this.

> Apart from the idea itself, I would suggest to make all changes in the openSUSE/doc-ci repository, not in SUSE/doc-sle.

In YaST we also share the [generic script for running the tests](https://github.com/yast/docker-yast-ruby/blob/master/yast-travis-ruby) in the Docker. But each module has to be still slightly adapted (modifying `.travis.yml`, adding `Dockerfile`, maybe installing some extra packages...). So it depends how much the potential other project will be similar and how many parts can be shared. But yes, sharing the most parts as possible is obviously a good approach.

## Yet Another Improvement

And with Travis you could actually go one step further. You could even build and publish the rendered HTML documentation as a preview. For rendering the YaST home page (http://yast.opensuse.org/, which is an alias for http://yast.github.io GitHub pages) we use the free [surge.sh](http://surge.sh/) service.

We can see the real rendered pages **before** merging a pull request. That's quite important for CSS or JS changes which are hard to imagine just from the diff. E.g. https://github.com/yast/yast.github.io/pull/129 can be checked at https://yast-pull-129.surge.sh/. The URL is added to the commit status, click the green :heavy_check_mark: checkmark next to the commit number.


Of course this is something for later, I'm writing it here just to share yet another improvement we already use in YaST. I do not know how much it is important for you to see the rendered preview.

svenseeberg:
Ignore this last commit. I got that I need to change the Dockerfile ;-)

lslezak:
NP, it just showed that downloading the docker image takes 27 seconds. :wink:

svenseeberg:
Thanks @lslezak! We updated our .travis.yml. The runtime is now below 3 minutes, but that included validating the DC-*-all files twice (with and without --validate).

lslezak:
Great! :tada: Glad to help!

PR #174, opened 2017-11-06
hint not to use autoyast2 option together with classes/rules

schubi2:


sknorr:
@schubi2 Lgtm! Fwiw, I pushed a small extra commit that tries to give a very short intro to rules/classes. Does that make sense to you?

sknorr:
Thanks! merged now.

PR #176, opened 2017-11-06
Stop validating with --remarks

sknorr:
I think, after all, validating with remarks is useless:

* I created a remark in an illegal position and tried to validate with
  and without remarks -> validation failed both times.
* I created a remark in a position where the document would be invalid
  if not for the remark -> it validated both times.

Sorry for arguing for validation with and without --remarks. Doesn't
seem to make any difference. Not even sure what the parameter --remarks
is good for when validating, ultimately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,6 @@ before_install:
   - docker run --rm -it sle-doc-image rpm -qa | sort
 
 script:
-  # It is important that we validate with and without remarks. Content between remarks is profiled out and this may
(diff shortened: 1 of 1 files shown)


svenseeberg:
There are two scenarios where it is useful:
* A xref links to an ID within remark tags requires validating without remarks
* linking to invalid IDs from within the remark requires validating with remarks

sknorr:
The thing is, the validation is _exactly_ the same. Nothing will break or not break because of --remarks, because `<remark>`s are handled purely by the DocBook stylesheets (which we only use when we need output) and don't seem to be handled by anything in the validation process chain.

`daps validate --remarks` does not seem to have much of a point right now. (However, maybe that is something that broke with the switch to Db5..?)

svenseeberg:
That seems to be correct. I just tried as well. Theory does not hold up. So yes, we can remove the validation with remarks to save us the minute.

PR #180, opened 2017-11-20
VT best practices: Add section describing aio-max-nr

jfehlig:
Add a section about tuning the fs.aio-max-nr setting when
running large numbers of virtual machines. For more details see

https://bugzilla.novell.com/show_bug.cgi?id=1067729


jfehlig:
It looks like the <screen> I added on line 915 needs to be moved out of the <para> block ending at line 916. I'll fix that and resubmit.

PR #186, opened 2017-11-27
Update the AutoYaST addons documentation (SLE15)

lslezak:
- The extension handling is a bit different in SLE 15:
  - The order is not important anymore, YaST automatically reorders the addons to meet the required registration order. (The SLE15 modular design with a lot of dependencies makes it difficult to write the correct order by hand.)
  - The dependent modules are automatically registered as well even though they are missing in the profile. (Again, to get all the dependencies right might be tricky, this makes the editing the profile by hand much easier. E.g. add just the "Server Applications Module" into the profile and do not care about that 3 dependent modules or much is that...)
  - See https://bugzilla.suse.com/show_bug.cgi?id=1065438 for the details
- Additionally removed/updated some SLE12 references

### DIFF
--- a/xml/ay_bigfile.xml
+++ b/xml/ay_bigfile.xml
@@ -1317,7 +1317,7 @@ exit 0
    <para>
     As an alternative to the fully automated registration, &ay; can also
     be configured to start the &yast; registration module during the
-    installation. this offers the possibility to enter the registration data
+    installation. This offers the possibility to enter the registration data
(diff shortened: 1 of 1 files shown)


svenseeberg:
I created a new branch that includes some changes to adhere to our style guide, see #187

lslezak:
OK, thanks! :+1:

svenseeberg:
Thanks for contributing :-)

PR #189, opened 2017-11-27
Fix Tip LVM Volume Groups

svenseeberg:


svenseeberg:
Already fixed with 2ba724563cd8fc068597ae3d8026257e8298bfb2

PR #194, opened 2017-11-28
Reworked and extended section "Securing the Network—Open Network Ports Detection"

jsegitz:


xml/hardening.xml
```diff
@@ -2521,6 +2546,13 @@ COMMAND  PID   USER   FD   TYPE    DEVICE SIZE/OFF NODE NAME
 sshd     2317  root   3u   IPv6    6579   0t0      TCP *:ssh (LISTEN)
 tftpd    2328  root   5u   IPv4    6698   0t0      TCP *:auth (LISTEN)
 sendmail 2360  root   3u   IPv4    6729   0t0      TCP 127.0.0.1:smtp (LISTEN)</screen>
+
+    <para>
+     Whatever method you decide to use: You should always check the results you get from
```
svenseeberg: I think this paragraph could be shortened somewhat. Example:

> To get a consistent view, run the scan from another machine and compare the results.

Alternatively split it into 2 shorter sentences, maybe without the colon.

xml/hardening.xml
```diff
@@ -2493,17 +2493,42 @@ Nmap run completed -- 1 IP address (1 host up) scanned in 221.669 seconds
...
-     you remove the UDP port scan (leave out the <option>-U</option> option),
-     then <command>nmap</command> will finish the port scan nearly immediately.
-     The results of <command>nmap</command> can vary widely and might not show
+     Note that running the <command>nmap</command> command will only scan 
+     certain well-known ports by default. If you want to scan all ports you need to 
+     specify <command>-p 1-65535</command>. Doing this will slow down the scan
+     considerably. Especially for UDP scans the use of an asynchronous scanner
+     such as masscan 
+     <link xlink:href="https://software.opensuse.org/package/masscan"/>
+     is recommended.
```
svenseeberg: I think a </para> is missing here. Then it probably validates.

xml/hardening.xml
jsegitz: Since we're talking about the lsof command right above I would prefer something like:
You should always check the results you get from a command run on a machine with an external scan.

xml/hardening.xml
jsegitz: the </para> is after the <screen> and for me it validates, strange

svenseeberg:
@jsegitz can you please have a look at #196 ?

PR #196, opened 2017-11-28
 Reworked and extended section "Securing the Network—Open Network Ports Detection"

svenseeberg:
Same as #194 with 2 minor fixes.


jsegitz:
LGTM, thanks.

PR #203, opened 2017-12-06
hardening guide: reworked BIOS section

mgerstner:
- an attempt to make some statements more precise
- grub does not really access the BIOS to determine how to boot the
  host, it's rather the other way around
- introduced a differentation between admin and boot passwords
- hint about boot device limitation, which is often the more interesting
  approach than a general boot password

### DIFF
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -1044,41 +1044,57 @@ FIXME:
      <title>Secure Boot</title>
      <para>
       This section describes only basic methods to secure the boot process. To
-      find out more about UEFI and the secure boot feature, see
-      <xref linkend="sec.uefi.secboot"/>.
+      find out about more advanced boot protection using UEFI and the
+      secure boot feature, see <xref linkend="sec.uefi.secboot"/>.
      </para>
     </tip>
 
     <para>
      The BIOS (Basic Input/Output System) or its successor UEFI (Unified
      Extensible Firmware Interface) is the lowest level of software/firmware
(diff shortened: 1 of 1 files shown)


xml/hardening.xml
```diff
@@ -1044,43 +1044,59 @@ FIXME:
...
-     not provide a high level of security (a BIOS can be reset, removed or
-     modified – assuming case access), it can be another deterrent.
+     Most BIOS implementations can be configured to prevent unauthorized users
+     from manipulating system and boot settings. This is typically done by
+     setting a BIOS admin or boot password. The admin password only needs to
+     be entered for changing the system configuration but the boot password
+     will be required during every normal boot. For most use cases it is
+     enough to set an admin password and restrict booting to the builtin
+     hard disk. This way an attacker will not be able to simply boot a Linux
+     live CD or USB thumb drive, for example. Although this does not provide a
```
svenseeberg: Remove "for example" (?) or move it to beginning of sentence.

xml/hardening.xml
```diff
@@ -1044,43 +1044,59 @@ FIXME:
...
+     high level of security (a BIOS can be reset, removed or modified –
+     assuming case access), it can be another deterrent.
     </para>
 
     <para>
-     Many BIOS capabilities have other various security settings – checking
-     with the system vendor, the system documentation or examine the BIOS
-     during a system boot.
+     Many BIOSs have various other security related settings – check with the
```
svenseeberg: @sknorr should that be BIOS' ?

PR #205, opened 2017-12-06
reworked section "Orphaned or Unowned Files"

jsegitz:


jsegitz:
the find command is dangerous, I'll fix it

PR #206, opened 2017-12-06
Reworked "Orphaned or Unowned Files"

jsegitz:


xml/hardening.xml
```diff
@@ -3075,7 +3075,35 @@ drwxrwxrwt 18 root root 16384 Dec 23 22:20 /tmp</screen>
...
+    You might need to extend the list of directories that are searched if you have a different filesystem structure.
+    </para>
+
+    <para>
+     A different problem are files that were not installed via the packaging
+     system and therefor don't receive updates. You can check for such files with
+     the following command:
+    </para>
+
+<screen># find /bin /lib /lib64 /usr -path /usr/local -prune -o -type f -a -exec /bin/sh -c "rpm -qf {} &amp;> /dev/null || echo {}" \;</screen>
```
svenseeberg: replace number sign (#) with \&root.prompt;

svenseeberg: Replace number sign (#) with \&prompt.root;

xml/hardening.xml
```diff
@@ -3075,7 +3075,35 @@ drwxrwxrwt 18 root root 16384 Dec 23 22:20 /tmp</screen>
...
+
+    <para>
+     You should run this command as an untrusted user (e.g. nobody) since crafted filenames might lead to command
```
svenseeberg: Remove "You should". "Run this command ..." is fine. "e.g." -> "for example"

xml/hardening.xml
jsegitz: This command must not be run as root

xml/hardening.xml
jsegitz: Thanks

xml/hardening.xml
svenseeberg: oh okay, then it should be > for which we have the \&prompt.user; entity. I'll fix it after the merge.

PR #216, opened 2017-12-12
Cleanup Deployment Guide

svenseeberg:
* Remove duplicate XML source by including xincludes
* Update / straighten chapter & section structure
* Straighten out XML IDs
* Clean out outdated stuff (heavy use of BIOS, etc)
* Reduce the amount of Warning / Info / Warning / Note boxes
* Clean up text when it is easy to do so


xml/deployment_bootoptions.xml
```diff
@@ -579,7 +549,11 @@ The normal installation mode.
...
        The default value is <literal>dhcp</literal> which configures the
-       network via DHCP. With the option
+       network via DHCP. Set <literal>netsetup=-dhcp</literal> when
+       configuring the network with with the boot parameters
+       <literal>hostip</literal>, <literal>gateway</literal> and
+       <literal>nameserver</literal>.
+       With the option
        <literal>netsetup=hostip,netmask,gateway,nameserver</literal> the
        installer asks for the network configuration during boot.
       </para>
```
fsundermeyer: This sounds weird and mixes things. 

* `netsetup=dhcp` -> will set up a network with DHCP even if linuxrc does not need it (e.g. when doing an installation from DVD
* `netsetup=hostip,netmask,gateway,nameserver` -> will prompt for the requested values and also set up a network in any case

See https://en.opensuse.org/SDB:Linuxrc#p_netsetup for more information

`hostname=` / `hostip=` / ... will only setup a network when _linuxrc_ needs it (e.g. when doing a remote installation). It will _not_ set up the network when doing an installation from a local image or DVD, for example. This behavior is _not_ something the customer expects! Also only sets up the network for the first interface that was found.

With `ifcfg=` you can configure networks on different interfaces. This is also the only way to configure VLANs.

See https://en.opensuse.org/SDB:Linuxrc#Network_Config for details on this topic.

xml/deployment_prep_aarch64_boot.xml
```diff
@@ -17,145 +17,7 @@
...
-  </para>
-  <para os="sled">
-   &ay; is extensively documented at
-   <link xlink:href="https://www.suse.com/documentation/sles-12/book_autoyast/data/book_autoyast.html"/>
-   .
-  </para>
- </sect2>
+ <xi:include xpointer="element(/1/3)" href="deployment_prep_x86_boot.xml"/>
+ <xi:include xpointer="element(/1/4)" href="deployment_prep_x86_boot.xml"/>
+ <xi:include xpointer="element(/1/5)" href="deployment_prep_x86_boot.xml"/>
```
fsundermeyer: Are you sure this works? What happens if someone changes deployment_prep_x86_boot.xml, e.g. by adding block elements?

svenseeberg: Yes this works, but yes, if someone adds a new para, then this may produce unwanted results. I'll check if we can get an xpointer that includes all para. But just in case that's why there is a XML comment with a warning at the beginning of this file.

PR #225, opened 2017-12-18
vt-doc: Improve text describing Xen dom0 memory management

jfehlig:
As noted by one of the SE's, the text regarding Xen dom0 memory
management in section 18.2 is a bit outdated. Update the examples
and improve the content.

### DIFF
--- a/xml/xen_virtualization_vhost.xml
+++ b/xml/xen_virtualization_vhost.xml
@@ -124,40 +124,43 @@
   <title>Managing &dom0; Memory</title>
 
   <para>
-   When the host is set up, a percentage of system memory is reserved for the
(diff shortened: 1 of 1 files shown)


svenseeberg:
I think we should somehow introduce the term "autoballooning". Something like "Autoballooning allows for over-committing the available memory to the hosted virtual machines. Autoballooning allocates memory to virtual machines when needed and takes the memory back that is not needed anymore."

jfehlig:
That is not an accurate description of autoballooning, but I see what you mean. I'll send another PR.

jfehlig:
Grrr, hit the wrong button, sorry. Intention was to update this PR, but now I actually have sent another - #226.

PR #226, opened 2017-12-18
vt-doc: Improve text describing Xen dom0 memory management

jfehlig:
As noted by one of the SE's, the text regarding Xen dom0 memory
management in section 18.2 is a bit outdated. Update the examples
and improve the content.


svenseeberg:
Thank you!

PR #231, opened 2017-12-22
Replace SuSEfirewall2 with firewalld

svenseeberg:
- SuSEfirewall2 will be replaced by firewalld in SLE-15
- SuSEfirewall2 will not even be in legacy any more
- It's unclear whether we'll get a replacement for the YaST firewall
module. Therefore I've focused on the command line configuration for the
moment.
- Adjusted references to SuSEfirewall sections in other documents


mgerstner:
Thanks for adjusting this!

I've noticed the failing Travis CI but somehow couldn't find out where the XML errors where coming from. The local checks of `daps -d <DOC> validate`  all went through. 

Can you give me a hint how to do better next time?

xml/security_firewall.xml
```diff
@@ -271,535 +271,400 @@
...
+
+    <screen>&prompt.root;firewall-cmd --zone=public --add-rich-rule='rule family="ipv4" \
+    source address="192.168.2.4" drop'</screen>
+   </sect3>
+
+   <sect3>
+    <title>Routing, Forwarding and Masquerading</title>
+
+    <para>
+     &firewalld; was originally designed to act as a desktop firewall and as
```
svenseeberg: @mgerstner do we need that firewalld is okayish for servers? This sounds like firewalld is only suitable for desktops.

svenseeberg:
About the validation: You need to install the geekodoc package. It will be used automatically by daps when installed. We're limiting the DocBook XML features for our SUSE documentation. In this case it screens are allowed in paragraphs in DocBook XML but not with geekodoc.

xml/security_firewall.xml
mgerstner: Yeah that's a difficult spot. What we want to communicate is that _firewalld_ is not designed to run as a fully fledged router. Maybe we should rephrase to something like "is not designed to implement a production router".

sknorr:
Sven wrote...
> It will be used automatically by daps when installed. We're limiting the DocBook XML features for our SUSE documentation.

Correction: It will not be used by DAPS automatically, unfortunately. So steps are:

1. install the `geekodoc` package
2. configure DAPS to use it:
    ```
    mkdir -p ~/.config/daps
    echo 'DOCBOOK5_RNG_URI=https://github.com/openSUSE/geekodoc/raw/master/geekodoc/rng/geekodoc5-flat.rng' > ~/.config/daps/dapsrc
    ```

mgerstner:
Thanks @svenseeberg, @sknorr for the info about geekodoc.

PR #232, opened 2017-12-22
Clarify how disklabel works

imobachgs:


svenseeberg:
Thank you for the contribution!

PR #240, opened 2018-01-23
SLE15 Offline Upgrade

lslezak:
This change is mainly related to [fate#323163](https://fate.suse.com/323163) and [fate#323395](https://fate.suse.com/323395).

## Summary

- The license agreement has been moved from the "welcome" screen later
(Background: the SLE15 medium contains several products (SLES, SLED, SLES4SAP), at that point we do not know which product will be upgraded yet. We do not know which license should be displayed. We can know that *after* selecting the partition to upgrade later in the workflow.)
- If the system is registered YaST uses SCC (or SMT) for providing the upgraded packages, the workflow is the same as in the online migration
- If the system is not registered user should use the "Packages" medium as suggested by YaST otherwise the system can be upgraded only to a minimal SLE15 system
- It is not possible to register the system during upgrade
  - If users want to use the packages from SCC they need to boot the original system and register it before running the upgrade
- The update options have been removed from the upgrade summary (see [bug#1071708](https://bugzilla.suse.com/show_bug.cgi?id=1071708) and related [bug#1075077](https://bugzilla.suse.com/show_bug.cgi?id=1075077))
- At the registration step the local registration servers are scanned before filling the registration code, you can select the server already that point

## Notes

- Feel free to reword or improve the text :wink: 
- The offline migration is still WIP, the code has not been submitted to SLE15 yet, thus even the latest build is not able to run the offline migration
- That means there still might be some changes, I'll let you know in that case
- The last item in the summary (scanning the local servers) is changed also in installation and in installed system, you may need to adapt also the other parts of the documentation...

### DIFF
--- a/xml/sle_update_offline.xml
+++ b/xml/sle_update_offline.xml
@@ -375,8 +375,8 @@ initrd (hd0,0)/boot/initrd.upgrade</screen>
    <step>
     <para>
      On the <guimenu>Welcome</guimenu> screen, choose
-     <guimenu>Language</guimenu> and <guimenu>Keyboard</guimenu> and accept the
-     license agreement. Proceed with <guimenu>Next</guimenu>.
+     <guimenu>Language</guimenu> and <guimenu>Keyboard</guimenu>.
+     Proceed with <guimenu>Next</guimenu>.
     </para>
     <para>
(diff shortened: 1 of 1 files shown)


svenseeberg:
Thank you, "fixed" indentation and one typo in #244

PR #242, opened 2018-01-24
seccheck: install john first otherwise no password auditing for you

vpereira:
A note about john and seccheck was added.


svenseeberg:
continued in #243

PR #243, opened 2018-01-24
Hardening Guide: package john needs to be installed

svenseeberg:


svenseeberg:
Thanks for the PR @vpereira !

PR #245, opened 2018-01-24
Document the fix to bsc#1075551

mvidner:
Try base product registration code first, https://github.com/yast/yast-registration/pull/352

I am not sure if it makes sense to document such a detail. Feel free to reject it, if it does not fit.


svenseeberg:
I think it is useful information.

PR #254, opened 2018-02-21
Add 'Migrate from openSUSE Leap to SLE'

svenseeberg:
* Add procedure for online migration with zypper
* FATE#323505


svenseeberg:
Waiting for Feedback, see FATE#323505

PR #258, opened 2018-03-05
Add a note about filesystem options

imobachgs:
Just to clarify that you should check documentation before specifying filesystem options in AutoYaST.


svenseeberg:
Thanks for the PR. Moved to #259 .

PR #263, opened 2018-03-06
corrected information about product to be selected

gilsonsouza:


lslezak:
The related AutoYaST fix is https://github.com/yast/yast-autoinstallation/pull/411

PR #266, opened 2018-03-09
Document CDATA; clarify config:type.

mvidner:
Prompted by https://bugzilla.suse.com/show_bug.cgi?id=1083622
YaST Scrum card: https://trello.com/c/esxuIcTv

I wanted to document the CDATA trick to distinguish between a missing element and an empty string.

Then I found out that the docs gets attributes quite wrong so I fixed that.

### DIFF
--- a/xml/ay_bigfile.xml
+++ b/xml/ay_bigfile.xml
@@ -317,6 +317,14 @@ exit 0
      restriction is specified in the schema files. A configuration component
      with more than one value must either be represented as an embedded list in a property value or as a nested resource.
     </para>
+    <para>
+      An empty element, such as <literal>&lt;foo&gt;&lt;/foo&gt;</literal> or
+      <literal>&lt;bar/&gt;</literal> will be NOT present in the parsed data
+      model. Usually this is interpreted as wanting a sensible default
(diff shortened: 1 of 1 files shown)


xml/ay_bigfile.xml
```diff
@@ -390,23 +381,22 @@ exit 0
      be treated as reserved words in the default namespace.
     </para>
     <para>
-     Global attributes are defined in the configuration namespace and must
```
mvidner: BTW I wondered why this is talking about "global" attributes, because I thought there is no such thing as local attributes.

Then I found that all the tables which document the individual **properties** of the various resources are using the heading **Attribute**. IMHO we should globally replace that with **Property** so that we don't create a confusion with the established XML term.

svenseeberg:
@mvidner thanks for the PR. Can I merge it?

mvidner:
@svenseeberg Well, yes, that's why I made the PR, right? If you mean whether it's finalized and ready, yes. (I haven't checked the rendering, but Emacs says the XML it's well-formed)

svenseeberg:
Just wanted to check if the [WIP] removal was final. And Travis says that everything is fine ;-)

PR #292, opened 2018-04-26
Replace SUSEFirewall2 with firewalld

svenseeberg:
* PCI-DSS (not verified)

### DIFF
--- a/xml/security_pcidss.xml
+++ b/xml/security_pcidss.xml
@@ -300,8 +300,8 @@
         interface <systemitem>eth0</systemitem> to server
         <systemitem class="ipaddress">10.0.0.1</systemitem>, use:
        </para>
-<screen>&prompt.root;<command>firewall-cmd</command> --permanent --direct --add-rule ipv4 filter OUTPUT 5 \
-    -d 10.0.0.1/32 -p udp -m udp --dport 53 -j ACCEPT
+<screen>&prompt.root;<command>firewall-cmd</command> --permanent --direct --add-rule ipv4 filter OUTPUT 23 \
+    -d 10.0.0.1/32 -o eth0 -p udp -m udp --dport 53 -j ACCEPT
 &prompt.root;<command>firewall-cmd</command> --reload</screen>
        <para>
         To block all other outbound traffic, see <xref linkend=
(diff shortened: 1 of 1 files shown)


svenseeberg:
https://forums.fedoraforum.org/showthread.php?301296-FIREWALLD-and-Transparent-Proxy-with-Squid

svenseeberg:
Creating a test setup for the basic router and transparent proxy configuration to verify the procedures.

xml/security_pcidss.xml
```diff
@@ -281,75 +281,71 @@
...
+        Deny all outbound and inbound traffic for which no exceptions
+        are defined as stated in the previous section. Forwarding is
+        usually completely disabled by a kernel parameter and should
+        not be enabled for endpoint servers.
+       </para>
+       <para>
+        To block all inbound traffic, set your Ethernet interface to
+        the zone <literal>block</literal>. Also allow already
+        established connections to persist:
+       </para>
```
abergmann: The firewalld default configuration already blocks all not specifically allowed incoming traffic. It has also the connection tracking rules. So there is no need to add them manually. Only the traffic for services that are explicitly enabled inside firewalld can come through.

By default all outgoing traffic is simply allowed. To change that we need to enable connection tracking on the OUTPUT chain and then drop all packet at the end.

The connection tracking rule need to be at the top of the chain. Therefore it get the priority of 0.
`firewall-cmd --permanent --direct --add-rule ipv4 filter OUTPUT 0 -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT`

The DROP rule for the OUTPUT chain should be at the end of the chain and therefore gets a high priority number.
`firewall-cmd --permanent --direct --add-rule ipv4 filter OUTPUT 99 -j DROP`

The `iptables` filter rules will look like this:
```
-A OUTPUT -j OUTPUT_direct
-A OUTPUT_direct -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
-A OUTPUT_direct -j DROP
```
And inside the firewalld XML configuration the rules look like this:
```
<?xml version="1.0" encoding="utf-8"?>
<direct>
  <rule ipv="ipv4" table="filter" chain="OUTPUT" priority="0">-m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT</rule>
  <rule ipv="ipv4" table="filter" chain="OUTPUT" priority="99">-j DROP</rule>
</direct>
```

xml/security_pcidss.xml
```diff
@@ -281,75 +281,71 @@
...
+        interface and not on the general network card if possible.
+        Define the source addresses that a service allows traffic from.
        </para>
        <para>
         For example, to allow only outbound DNS requests over the
         interface <systemitem>eth0</systemitem> to server
-        <systemitem class="ipaddress">10.0.0.4</systemitem>, use:
+        <systemitem class="ipaddress">10.0.0.1</systemitem>, use:
        </para>
-       <screen>&prompt.sudo;iptables -A OUTPUT -d 10.0.0.4/32 -o eth0 -p udp -m udp --dport 53 -j ACCEPT</screen>
```
abergmann: The new firewalld rule does not specify an interface.

`firewall-cmd --permanent --direct --add-rule ipv4 filter OUTPUT 23 -o eth0 -d 10.0.0.1/32 -p udp -m udp --dport 53 -j ACCEPT`

This rule has only an affect on the outgoing traffic if we set the general OUTPUT policy to DROP.  See 1.2.1.c.

abergmann: The priority (in my example 23) must be between the connection tracking priority and the DROP rule priority.

xml/security_pcidss.xml
abergmann: Maybe it is a good idea to show the firewalld configuration as well, so that the administrator can double check them after the command line was executed.

xml/security_pcidss.xml
svenseeberg: For the record: added an -o eth0 and updated the priority.

PR #303, opened 2018-05-17
add more info about loader type for autoyast

jreidinger:
Feel free to modify it as needed. It is mainly reaction to one bug report where it is not clear to user what is possible values.


svenseeberg:
Thanks for the pull request. I restructured the text into a bullet list. Can you please review #304 ?

PR #309, opened 2018-06-12
Add Raspberry Pi Quick Start

svenseeberg:
* Copied from SBP and adapted for SLE15


afaerber:
@svenseeberg Please revert that reformatting, it breaks my pending pull request. You can still reformat after merging it.

PR #311, opened 2018-07-05
Remove `/connect/` from SMT URL

lagartoflojo:
To configure the installer to use a registration proxy, the FQDN of SMT/RMT is enough. There's no need to append `/connect/`, since the SUSEConnect library takes care of that on its own. In fact, SUSEConnect does not handle paths, and even if `fqdn/connect` is given as the regurl, it will only use `fqdn`, and ignore the path. Giving a path can cause a bug when de-registering a system (bsc#954451).

### DIFF
--- a/xml/rmt_client.xml
+++ b/xml/rmt_client.xml
@@ -65,12 +65,12 @@
    The parameter needs to be entered as
    <literal>regurl=<replaceable>SMT_SERVER_URL</replaceable></literal>. The URL
    needs to be in the following format:
-   <literal>https://<replaceable>FQDN</replaceable>/connect/</literal> with
+   <literal>https://<replaceable>FQDN</replaceable></literal> with
    <replaceable>FQDN</replaceable> being the fully qualified host name of the
    &rmt; server. It must be identical to the FQDN of the server certificate
    used on the &rmt; server. Example:
   </para>
-  <screen>regurl=https://smt.&exampledomain;/connect/</screen>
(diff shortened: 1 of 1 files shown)


svenseeberg:
Thanks

lagartoflojo:
Thanks for merging @svenseeberg.
Any chance this change is reflected on SLE12 documentation?
For example, here: https://www.suse.com/documentation/sles-12/book_smt/data/smt_client_parameters.html
Maybe I can make the same PR against a different branch?

svenseeberg:
Is this valid for all SLE 12 Service Packs? I can do a cherry-pick into all maintenance branches.

svenseeberg:
@lagartoflojo: the /connect/ path also appears here: https://susedoc.github.io/doc-sle/SLE12SP3/SLES-smt/single-html/#smt.client.rhel.rhel52
Can I drop this as well?

lagartoflojo:
Actually, I think from that page, all these lines should be removed: https://github.com/SUSE/doc-sle/blob/maintenance/SLE12SP3/xml/smt_client.xml#L891-L894

It says:

>  or (for SUSE Customer Center) 

but clearly, it's not talking to SCC, but rather to SMT.
Additionally, `suse_register` does not know how to talk to SCC.

svenseeberg:
I dropped all sections about RHEL. So the last part is not relevant anymore. And I removed /connect/ and SCC URL from the SLE12 SP3 docs.

lagartoflojo:
@svenseeberg great, thank you!

PR #313, opened 2018-07-05
Update Quickstart guide for SLE15

afaerber:
Reflect SLE15 support of new Model B+ as well as SLE15 having gcc7 now.


svenseeberg:
Thanks!

PR #314, opened 2018-07-17
RMT handbook review

ikapelyukhin:
:heavy_check_mark: Ready to be picked up by @svenseeberg!

### DIFF
--- a/xml/rmt_overview.xml
+++ b/xml/rmt_overview.xml
@@ -71,7 +71,7 @@
   <para>
    &rmt; replaces SMT (Subscription Management Tool) which was used for &slea; 11
    and &slea; 12. For a feature comparison between &rmt; and SMT, see
-   <link xlink:href="https://github.com/SUSE/rmt#rmt-and-smt"/>.
+   <xref linkend="sec.rmt_migrate.notes.features" />.
   </para>
  </sect1>
  <sect1>
--- a/xml/rmt_client.xml
+++ b/xml/rmt_client.xml
@@ -33,7 +33,7 @@
   </listitem>
   <listitem>
    <para>
-    Use the <command>crmt-client-setup</command> command. See <xref
+    Use the <command>rmt-client-setup</command> command. See <xref
     linkend="sec.rmt_client.clientsetupscript"/>).
    </para>
   </listitem>
--- a/xml/rmt_config_files.xml
+++ b/xml/rmt_config_files.xml
@@ -44,19 +44,11 @@
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><literal>verify_rpm_checksums</literal></term>
-     <listitem>
-      <para>
-       Decides whether to mirror source RPM packages (architecture is
-       <literal>src</literal>).
-      </para>
-     </listitem>
-    </varlistentry>
-    <varlistentry>
      <term><literal>dedup_method</literal></term>
      <listitem>
       <para>
-       Enables deduplication when set to <literal>hardlink</literal>.
+       Creates hardlinks during mirroring when set to <literal>hardlink</literal>.
+       If the file system does not support hardlinks, can be set to <literal>copy</literal> instead.
        Possible values: <literal>hardlink</literal>,
        <literal>copy</literal>.
       </para>
--- a/xml/rmt_migrate_from_smt.xml
+++ b/xml/rmt_migrate_from_smt.xml
@@ -466,7 +466,7 @@
       </entry>
       <entry>
        <para>
-        no 
+        no
        </para>
       </entry>
      </row>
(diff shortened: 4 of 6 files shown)


xml/rmt_overview.xml
```diff
@@ -61,18 +56,6 @@
...
-   <listitem>
-    <para>
-     You are looking for a turnkey box staging solution for testing updates
-     before releasing them to the clients.
-    </para>
-   </listitem>
-   <listitem>
-    <para>
-     You want to have a quick overview of the patch status of your &sle; servers and
-     desktops.
```
ikapelyukhin: There's no patch status reporting in RMT.

xml/rmt_overview.xml
```diff
@@ -61,18 +56,6 @@
      or internal repositories into your update solution.
     </para>
    </listitem>
-   <listitem>
-    <para>
-     You are looking for a turnkey box staging solution for testing updates
-     before releasing them to the clients.
```
ikapelyukhin: There's no repository staging in RMT.

xml/rmt_overview.xml
```diff
@@ -38,11 +38,6 @@
      You want to update &sle; servers.
     </para>
    </listitem>
-   <listitem>
-    <para>
-     You want to get a detailed overview of your company's license compliance.
```
ikapelyukhin: RMT can't do that. (SMT tries to achieve that, but that doesn't work very well)

xml/rmt_overview.xml
```diff
@@ -61,18 +56,6 @@
      or internal repositories into your update solution.
```
ikapelyukhin: @svenseeberg That sentence seems weird: integrate additional software **OR** update external
     or internal repositories into your update solution?

ikapelyukhin:
@svenseeberg, here are some additional comments:

#### SCC UI has changed and mirroring credentials are on a separate tab now

https://github.com/SUSE/doc-sle/blob/119cc51a1084d374ce766423ee57a122c0561031/xml/rmt_mirroring.xml#L47-L56

This probably needs to be updated elsewhere in the docs (e.g., SUMA).
Also, this section should probably reference where to find the credentials as well:

https://github.com/SUSE/doc-sle/blob/119cc51a1084d374ce766423ee57a122c0561031/xml/rmt_config_files.xml#L123-L128

#### There's an alternative way to enable products to what is currently described

https://github.com/SUSE/doc-sle/blob/119cc51a1084d374ce766423ee57a122c0561031/xml/rmt_mirroring.xml#L149-L159 

The alternative way to do the same is `sudo rmt-cli product enable PackageHub/15/x86_64`. The product string is the same as in the output of `SUSEConnect --list-extensions`.

Maybe this could be documented as well, but I'm not sure.

#### Offline migrations

There's a section mentioning online migrations:

https://github.com/SUSE/doc-sle/blob/119cc51a1084d374ce766423ee57a122c0561031/xml/rmt_client.xml#L239-L253

Offline migrations are supported as well. Perhaps this also should be reflected in the documentation.

#### Clarification on the sync command

It might be a good idea to expand on the sync a bit:

https://github.com/SUSE/doc-sle/blob/119cc51a1084d374ce766423ee57a122c0561031/xml/rmt_scripts.xml#L96-L104

I think it's worth mentioning that during sync no data is being uploaded to SCC, only downloaded (product definitions, repository data, etc. that's currently available the SCC organization)

#### Move the paragraph?

This seems like a weird place for this paragraph to be:

https://github.com/SUSE/doc-sle/blob/119cc51a1084d374ce766423ee57a122c0561031/xml/rmt_scripts.xml#L259-L272

It would be better placed somewhere into the overview of the setup/process.

#### Clone of rmt-client-setup instructions

Currently there are two places where the same procedure is described:

https://github.com/SUSE/doc-sle/blob/119cc51a1084d374ce766423ee57a122c0561031/xml/rmt_config_files.xml#L195-L197

And:

https://github.com/SUSE/doc-sle/blob/119cc51a1084d374ce766423ee57a122c0561031/xml/rmt_client.xml#L180-L194

One uses `curl` and `sh`, the other uses `wget` and `chmod +x`. I would suggest unifying those.

ikapelyukhin:
@svenseeberg @tmuntaner

In the offline setup section enabling of the repos is the last item:

https://github.com/SUSE/doc-sle/blob/119cc51a1084d374ce766423ee57a122c0561031/xml/rmt_mirroring.xml#L354-L360

I think the original design of the procedure implied enabling the repos on `sirius`, bring the settings back to `sun`, mirroring the repos and bringing them back to `sun`?

xml/rmt_overview.xml
svenseeberg: Indeed. Let's change it.

svenseeberg:
@ikapelyukhin regarding the air-gapping procedure:

AFAIK one creates a normal RMT server (sun), connected to SCC, and downloads all data. After exporting all required data and importing it on the air-gapped server (sirius), the imported repos need to be enabled.

PR #320, opened 2018-08-03
add info from trello card

taroth21:
https://trello.com/c/E7nx1EfP

### DIFF
--- a/xml/docker_registry.xml
+++ b/xml/docker_registry.xml
@@ -180,7 +180,9 @@
   </para>
 
   <para>
-    Portus is accessible for &slea;&nbsp;12 customers as Docker image.
+    Portus is accessible for &slea;&nbsp;12 customers as Docker image. To
+    migrate from the RPM package to the Docker image, follow the instructions
+    at <link xlink:href="http://port.us.org/docs/migrate-from-rpm.html"></link>.
   </para>
 
   <para>


jordimassaguerpla:
The change looks good. However, portus is not yet available in the SUSE Registry.

taroth21:
@jordimassaguerpla : Thanks! Should we wait for the outcome of bsc#1080879c#9 before merging this?

jordimassaguerpla:
@taroth21 :  Whatever works best for you. We will need to update the Protus chapter once we know in which namespace is released inside the SUSE Registry, so we can tell the user to do "docker pull NAMESPACE/docker" but since it is not released, I don't know which is the namespace. Should I open a new bug for this or add a comment in bsc#1080879

taroth21:
@jordimassaguerpla : Ok, then I would say, let's merge this one as it is now. 

Please open a new bug report once you know the namespace, then we can update the docs accordingly. (As bsc#1080879 is closed already and originally just was about fixing broken links, I would prefer a new bug). 

After merging this PR, I will also set https://trello.com/c/BjaW9DsL on our SLE 15 doc board to 'DONE'. It covered two items:

- https://bugzilla.suse.com/show_bug.cgi?id=1098836 (already fixed)
- the reference to this card on your board: https://trello.com/c/E7nx1EfP/742-update-quick-start-guide-portus-chapter

As I do not have write access to your board, feel free to write down on the card what the current status is and what is still missing. Many thanks!

PR #321, opened 2018-08-03
remove content mentioned in bsc#1102817

taroth21:

PR #323, opened 2018-08-10
Docker Guide: remove (most) abbreviated productnames

taroth21:
- replace SLE/&slea;/&slsa; with &productname; entity 
  because the documentation should not contain abbreviated 
  productnames  
- only leave &slea; in very few places for readability reasons 
  (where full productname makes the sentence or title very hard
  to read)
- remove abbreviated productname from two section titles,
   added full productname to title of parent section instead


taroth21:
@sknorr : A second pair of eyes would be helpful - please review. :)
Contrary to your proposal in https://github.com/SUSE/doc-sle/pull/321#discussion_r208627388, I decided _not_ to use &productnumber;. For reasons unknown to me, some sections of this document deal with multiple product versions ATM and there, &productnumber; would need to be replaced with a hard coded version number for the next release - the risk is high to forget about this. for next revision

flavio:
All the SUSE container images are going to be published to the SUSE registry (registry.suse.com).

Right now (via RAW API request) I can see the following contents inside of the registry:

```
sles12                                sp3
sles12sp2                             
sles12sp3                             
sles12sp3-container                   
suse/sle15                            latest
suse/sles12sp3                        2.0.2
```

Needless to say, there's some confusion. I think we have to stay consistent with the naming. Right now maintenance QA is in charge of pushing the images over there.

Adding @celan from maintenance QA to shed some light on that.
Adding also @Vogtinator, maintainer of the SLE images too. Maybe he knows more about the naming/tagging scheme that is going to be used by our base images.

flavio:
**Note well:** the are no tags pushed for the `sles12sp2`, `sles12sp3` and `sles12sp3-container` images (not even `latest`!). Maybe that's due to the API I'm calling, however this looks really suspicious and I don't even know how this could have happened.

celan:
@flavio I reported this already several times in person. Now I opened a GitHub issue for the build service:
https://github.com/openSUSE/open-build-service/issues/5625

Vogtinator:
@flavio: The base container naming scheme was discussed in http://bugzilla.suse.com/show_bug.cgi?id=1083671

taroth21:
All: thanks for the information! 
@flavio, @Vogtinator: I agree about using consistent image names in the first place. :) As the older ones are called 'sles[productnumber]' and the newer one 'suse/sle[productnumber]', I'm struggling with how to refer to them generically in the text (when *not* citing the actual image? 'SUSE Linux Enterprise Server' images? or 'SUSE Linux Enterprise' images? (I would prefer a consistent term here). According to Stefan's last argument, it would probably make more sense to refer to them as  'SUSE Linux Enterprise Server' images. Would you be fine with that?

Also, the document mentions  'SUSE Linux Enterprise' as host of the Docker containers (which means the host could either be a SLES *or* SLED machine). Is this OK? Or should we change this to  'SUSE Linux Enterprise Server'?

Also, is it correct to say 'Adding SLE Extensions and Modules to Images' (which could also include adding the  'SUSE Linux Enterprise Workstation Extension'). Or should we say  'Adding SLES Extensions and Modules to Images'?

Vogtinator:
@taroth21 Yes. However, I'll restart the naming discussion yet again and ask whether sles or sle in the name is more correct.

flavio:
@taroth21 **note well** (@celan correct me if I'm wrong): SLE container images are going to be distributed only via the SUSE registry. That means the old images on the SLE12 containers module are not going to be updated. SLE12 customers will be able to pull the updated ones from the registry.

Basically `docker pull` is going to be the only way to get/update SLE images for all our customers, regardless of the SLE version installed on their docker host.

celan:
@flavio This is also an open topic, which PM has to decide. It seems customers still check for new RPMs in the Containers module.

@taroth21 The namespaces and names are not settled yet. Maybe a TID would make more sense here, because that could react faster to changes and can document temporary things.

taroth21:
@celan , @flavio: Any update on the latest questions (from Aug 13) yet?

taroth21:
@celan: Thanks! :)

However, looking through the comments from the beginning, it still seems the following issue is not solved yet: 

1. As the older Docker images are called 'sles[productnumber]' and the newer one 'suse/sle[productnumber]', I'm struggling with how to refer to them _generically_ in the text (when not citing the actual image? 

- 'SUSE Linux Enterprise Server' images? 
- 'SUSE Linux Enterprise' images?

    According to Stefan Knorr's argument , it would probably make more sense to refer to them as 'SUSE Linux Enterprise Server' images. Would you be fine with that?

2. Also, the document mentions 'SUSE Linux Enterprise' as host of the Docker containers (which means the host could either be a SLES or SLED machine). Is this OK? Or should we change this to 'SUSE Linux Enterprise Server'?

3. Also, is it correct to say 'Adding SLE Extensions and Modules to Images' (which could also include adding the 'SUSE Linux Enterprise Workstation Extension'). Or should we say 'Adding SLES Extensions and Modules to Images'?

Including @ZerwasMOD - maybe he has answers to those questions.

sknorr:
@taroth21:
> According to Stefan Knorr's argument , it would probably make more sense to refer to them as 'SUSE Linux Enterprise Server' images. Would you be fine with that?

You misread my comment, I was arguing for using "SLE" not "SLES".

taroth21:
@sknorr : Sorry, corrected this in my comment above.

celan:
The naming of the container images is a very good question. AFAIK they are supposed to be SLES base images, but somehow they may also serve as base containers for SLE in general. However, this is a question product management has to answer. Michal Svec is the assigned PM for containers.

m-rey:
implemented in #397 with ca8ee71cdfb142167385cf0eb75a760045b2c27b

PR #334, opened 2018-08-30
Fix typo in autoyast system registration section

karras:
This small PR adds some missing words in the AutoYaST Registration Server section.


svenseeberg:
Thanks for the PR.

PR #338, opened 2018-09-13
Fix changed paths in RMT Guide

svenseeberg:


svenseeberg:
Validated locally, Travis is too slow.

PR #340, opened 2018-09-15
Style edits

bbevan:
Hi,

I've made a few more style edits. Please let me know if anything should be changed. I will be glad to make those changes.

This time, there was no need to squash merge. I made one commit after I edited the documents for style.

Also, if any of these documents are a higher priority than others, please inform me. I will be glad to prioritize editing.

Thanks
-Brandon

### DIFF
--- a/xml/64bit_issues.xml
+++ b/xml/64bit_issues.xml
@@ -67,36 +67,35 @@
   </important>
 
   <para>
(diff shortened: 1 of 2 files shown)


svenseeberg:
A belated thank you for contributing!

PR #345, opened 2018-09-26
Document a Self-Update temporary repository

lslezak:
- Document a Self-Update temporary repository which might be used in some cases during installation.
- The reason is that in some cases there might be used extra `SelfUpdate0` repository during installation and this might be confusing for the users. Let's make it clear.
- This is related to bug https://bugzilla.suse.com/show_bug.cgi?id=1101016
- This change is valid only for SLE15-SP1/Leap 15.1

### DIFF
--- a/xml/deployment_yast_installer.xml
+++ b/xml/deployment_yast_installer.xml
@@ -356,6 +356,37 @@
      </para>
     </note>
    </sect3>
+
+   <sect3 xml:id="sec.yast_install.self_update.temporary_addon_repo">
+    <title>Temporary Self-Update Addon Repository</title>
+
+    <para>
+     Some packages distributed in the Self-Update repository should not be
+     applied as an update for the installer but should just override the
+     available packages from the installation medium.
+    </para>
+
+    <para>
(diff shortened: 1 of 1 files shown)


taroth21:
@lslezak: Thanks a lot for the PR! As this change is only valid for SLE15-SP1/Leap 15.1, we will merge the PR after we have created a maintenance branch for SLE 15 and will start to work on SLE 15 SP1 in 'develop'. 

Do you want to fix a few minor issues yourself (mostly make sure to comply with the style guide)? Then I would comment on them here. Or do you want us to apply the style changes ourselves when merging the PR? Both is fine with us.

lslezak:
Please fix the style issues, I'm not familiar with your style guide and it would take more time, you will be faster. Thank you!

taroth21:
@lslezak : No problem, will do so before merging. :)

taroth21:
- during installation, an extra ```SelfUpdate0``` repository might be used
- see also https://bugzilla.suse.com/show_bug.cgi?id=1101016

PR #355, opened 2018-10-19
Localization via itstool and .po files

Gianvacca:
Taking inspiration from the work done for the projects [](https://github.com/openSUSE/packages-i18n) and [](https://github.com/openSUSE/release-notes-openSUSE) I have created a Makefile to update .pot and .po files, to translate the .xml files for the book(s) of interest, and to generate a .pdf/.text/.html output.
The Makefile also requires a couple of bash script: po-selector (adapted from the homonymous script used in [release-notes](https://github.com/openSUSE/release-notes-openSUSE)) and xml-selector (totally mine, but inspired from the former).
I have also imported and adapted from [packages-i18n](https://github.com/openSUSE/packages-i18n) the script used to publish on Weblate, but I have no way to test it.

PR #358, opened 2018-10-24
Add translations with PO files

svenseeberg:
This PR is a continuation of #355 with the following changes to the ```develop``` branch:
* Add PO and POT files in the ```locale``` directory. The ```locale``` directory contains the https://github.com/openSUSE/doc-sle-translation repository as a git submodule.
* Add a Makefile for dumping translatable strings into POT files and PO files back into translated documentation.
* The .gitmodules file tells git to ignore the state of the locale submodule. It is of no interest to the parent repository, which branch is currently checked out in the submodule.


svenseeberg:
@Gianvacca can you please check if all PO files are correctly copied? This PR contains fewer files than your original. Thanks!

sknorr:
@svenseeberg Trying to review here is pretty awful because of the sheer size of the commit. Any chance you can split up commits between Makefile, infrastructure, POT files, and PO files (maybe also split PO file commits by language)?

Anyway, a few things I noticed:

* By moving stuff to the `locale` dir you have probably broken paths to everything that resides in the `50-tools` dir. I am also not quite sure whether the `50-tools` dir should be in the `doc-sle-translations` Git repo or in this repo.
  * PRO 50-tools in new repo: probably easier to access for translators who (frankly) tend to have more of a clue in terms of translations than us writers do anyway
  * PRO 50-tools in this repo: the translations repo stays cleaner and only contains actual translations; this repo does not just have a Makefile that would be broken without the extra tools

* `Makefile:157`: We don't profile for LIFECYCLE in our docs, so that currently does not make sense (doesn't hurt either, just does nothing).

* `Makefile:172`: This can go, we are not interested in generating docs to display in YaST's release notes screen.

* `Makefile:183`/`Makefile:221`: Yes, I think running fix-up.xsl probably still makes sense. This is something that translators occasionally break by translating too much and we don't have a viable way of excluding it from the translated content (our DocBook variant doses not include the ITS tag set)

svenseeberg:
I agree on moving the 50-tools folder directly into the doc-sle repo main directory.

svenseeberg:
Addition to the Makefile: It should automatically checkout the same branch name in the submodule as the doc-sle repository. This is useful to have the correct translations. Alternatively, it should at least print a warning if doc-sle and doc-sle-translations are on differently named branches.

svenseeberg:
Additional idea: Travis CI could automatically update the doc-sle-translations repository with a dump from the most current XML sources.

Gianvacca:
@sknorr Right now I'm working on the Makefile to account for the new directory tree. I will also delete those parts you have pointed out and credit Guo Yunhe. I hope I have time to getit ready for this evening, otherwise tomorrow.

tomschr:
> Additional idea: Travis CI could automatically update the doc-sle-translations repository with a dump ...

@svenseeberg 
If that can be automated, why not? I remember that the translations with PO files can get messy, especially with XML. I don't exactly recall what the problem was, but it the `gettext` tools had issues with XML. Hope, these issues are fixed with the `itstool` now. 

Maybe ask Karl for same historic context. :wink:

svenseeberg:
@guoyunhe ping, if you want to get involved.

Gianvacca:
@svenseeberg I have done the modifications to Makefile and 50-tools/po-selector in my repo. Should I make a pull request?

guoyunhe:
Haven't got time to check the code. But it is good to see some new progress!

svenseeberg:
@Gianvacca @sknorr @tomschr next round for feedback. AFAICT all your suggestions were implemented, excpet for splitting the commit of the PO files in the doc-sle-translations repo.

Gianvacca:
@svenseeberg actually there is still one @sknorr point missing: the one related to [fix-up.xsl](https://github.com/openSUSE/release-notes-openSUSE/blob/Leap_15.0/fix-up.xsl) used for translation in the [release-notes-openSUSE project](https://github.com/openSUSE/release-notes-openSUSE). I am not familiar with xsl, but if someone explains me what it does and how to use it, I can put it into code.

sknorr:
@Gianvacca About the XSLT thing:

So, the purpose of it: We have some metadata in our documents that we use to create Report Bug links. Unfortunately, we can't remove that from the PO file, hence translators sometimes translate the content, even though it is not meant to be. A German translator might translate the Bugzilla component "Documentation" to "Dokumentation" and that would then be invalid for Bugzilla. To prevent such mistakes from holding us up, I created the fix-up.xsl that resets these values to the correct ones.

* Uncomment Makefile:200-213
* Make sure that something like Makefile:162-168 is run for every file
* Not every file contains this kind of metadata, so make sure to check nothing fails if the metadat does not exist

tomschr:
@sknorr 
> To prevent such mistakes from holding us up, I created the fix-up.xsl that resets these values to the correct ones.

This is probably one way to solve it. Another way could be to add a custom ITS file and to mark any `dm:docmanager` tags as not being translated:

```xml
<its:rules
    xmlns:its="http://www.w3.org/2005/11/its"
    xmlns:itst="http://itstool.org/extensions/"
    xmlns:db="http://docbook.org/ns/docbook"
    xmlns:dm="urn:x-suse:ns:docmanager"
    version="2.0">

<!-- content from /usr/share/itstool/its/docbook5.its -->

<its:translateRule translate="no" selector="//dm:docmanager"/>
</its:rule>
```

The customized ITS file can be passed to the `itstool` command with the `-i`/`--its` option.

Not sure if this could be an option. If this is feasible, we can avoid any XSLT fixes. :wink:

Gianvacca:
Indeed I was looking into @tomschr approach that should be cleaner and more easily maintainable to me.

sknorr:
Okay--I hadn't really given that much thought (not quite sure why), but go for it -- though be sure to base it off itstool's existing DocBook ITS file.

Gianvacca:
The ITS rule posted by @tomschr works, I have tested it in my repo. I have created a `suse.its` file with it, without adding the content of /usr/share/itstool/its/docbook5.its, since I have discovered that `itstool` loads all rules from `/usr/share/itstool/its`, therefore it would be a duplicate. Where should this file be located and named?

tomschr:
@Gianvacca 
> The ITS rule posted by @tomschr works

Great! :+1: 

> I have discovered that itstool loads all rules from /usr/share/itstool/its, therefore it would be a duplicate.

Ahh, that's good! That makes life easier. :wink: 
Just for the records, if the default doesn't work as expected. The alternative would be to add a pointer to the original ITS file.

> Where should this file be located and named?

I'm not sure, I don't have any preferences yet.

One idea: Shouldn't that be included into the daps package? In that case we could add the `suse.its` file and install it where the other its files are stored. That would be `/usr/share/itstool/its`.

Not sure if daps should add another dependency or outsource it into a subpackage. I guess, this is only needed for translators and not the ordinary writer, right?

---

The ITS specification allows to refer to an external rule set which is included through XLink (see [5.4 Link to External Rules](https://www.w3.org/TR/its/#link-external-rules)).

If my understanding of the ITS specification is correct, that looks like this:

```xml
<its:rules version="2.0"
     xmlns:its="http://www.w3.org/2005/11/its"
     xmlns:xlink="http://www.w3.org/1999/xlink"
     xmlns:dm="urn:x-suse:ns:docmanager"
     xlink:href="file:///usr/share/itstool/its/docbook5.its">
   <its:translateRule translate="no" selector="//dm:docmanager"/>
</its:rules>
```

Just in case. :wink:

tomschr:
For the record and if someone is interested: I've built an RPM package for the ITS schema. The package is currently under review (https://build.opensuse.org/request/show/645905) and will end up in the [obs://Documentation/Tools](https://build.opensuse.org/project/show/Documentation:Tools) repo once it's accepted.

It's not really needed for this PR, but it doesn't hurt. :wink:

Gianvacca:
@tomschr Hi Thomas, I have noticed only now that, regardless of the new its rule, from a quite long list of list of xml files no translatable string are extracted, so the resulting pot is 0 B in size. This is the list of the first 10:

- 64bit_issues.pot
- aarch64_inst_problem.pot
- admin_docupdates.pot
- apparmor_intro.pot
- apparmor_managing.pot
- apparmor_pam.pot
- apparmor_repositories.pot
- art_kgraft.pot
- art_pcidss.pot
- audit_moreinfo.pot

I have inspected the sources, but  couldn't establish the culprit. Any idea?

tomschr:
Hi @Gianvacca

well, not sure how the usual steps are. I used the following steps which seems to work at least for `64bit_issues.xml` (I only tested this file):

1. Prepare the translation: create a `po` directory and a translation directory (here I use `de`).

1. Create the `suse.its` file with the following content:

       <its:rules version="2.0"
             xmlns:its="http://www.w3.org/2005/11/its"
             xmlns:xlink="http://www.w3.org/1999/xlink"
             xmlns:dm="urn:x-suse:ns:docmanager">
                <its:translateRule translate="no" selector="//dm:docmanager"/>
       </its:rules>

1. Run the `itstool`:

       $ itstool -l en --out po/64bit_issues.pot -i suse.its xml/64bit_issues.xml

   Repeat this step with every file you need.

1. Load the POT file in your favorite translation tool. I use KDE*s `lokalize`.

1. "Translate" *some* strings to see if everything goes right. You don't need to translate all. :wink: 

1. Convert the POT file into a MO file:

       $ msgfmt po/64bit_issues.pot -o po/64bit_issues.mo

1. Create the translated XML with:

       $ itstool -o de/64bit_issues.xml  -m po/64bit_issues.mo xml/64bit_issues.xml


Does that help?

Gianvacca:
@tomschr It's exactly what the Makefile does. So, if you don't get an empty pot, it could be a bug in the latest version of itstool I have built. I'll revert to the repo version and see what happens.

tomschr:
@Gianvacca 

>  It's exactly what the Makefile does

Ok, at least I haven't done anything wrong. :laughing:  I use the same version (2.0.2) on openSUSE Leap 15.0. Here is the summary of my versions:

```
$ rpm -q libxml2-2 libxml2-tools python2-libxml2-python python3-libxml2-python itstool
libxml2-2-2.9.7-lp150.2.3.1.x86_64
libxml2-tools-2.9.7-lp150.2.3.1.x86_64
python2-libxml2-python-2.9.7-lp150.2.3.1.x86_64
python3-libxml2-python-2.9.7-lp150.2.3.1.x86_64
itstool-2.0.2-lp150.1.7.noarch
```

Not sure if that helps.

Gianvacca:
Hi @svenseeberg, please notice that tabs have a precise meaning in Makefiles. They can't be substituted by white spaces. I used white spaces fore mere indentation (in functions or if blocks), but tabs are for recipes, [they have a functional role](https://www.gnu.org/software/make/manual/html_node/Recipe-Syntax.html). Without, the Makefile gets broken.

svenseeberg:
@Gianvacca oh, sorry. I'll fix that.

Gianvacca:
@svenseeberg Thanks, Sven, but unfortunately now there are tabs also where there shouldn't be. Maybe you can use the [Makefile in my repo](https://github.com/Gianvacca/doc-sle/blob/feature/translations2/Makefile) as a reference. If you find two spaces at the beginning of the line, those are for indentation within functions or if blocks (optional, they can be removed); if you find a tab, it's for recipe content (mandatory). By the way, in the future I'll stick to the indentation style you'll adopt.

svenseeberg:
@Gianvacca I hope that I fixed it now. It should only contain cosmetic fixes now.

svenseeberg:
@Gianvacca @tomschr @sknorr any reason to not merge for now? And who can take care of setting up Weblate?

sknorr:
@svenseeberg I could probably take care of Weblate.

50-tools/weblate-create-projects-oneshot.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+# This script generates initial projects import script. It was used only once
+# and it could be called never more. Once regular projects maintenance will be
+# implemented, it could be deleted. This is just a source of needed code.
+
+BRANCH_ON_GITHUB=develop
+BRANCH_ON_WEBLATE=$BRANCH_ON_GITHUB
+BRANCH_SLUT=develop
```
sknorr: This variable name seems unfortunate. And I don't know what it even means. Could you change it?
(I know this is probably not written by you.)

50-tools/weblate-create-projects-oneshot.sh
```diff
@@ -0,0 +1,66 @@
...
+	else
+		echo "weblate_create_component doc-sle \"$1 $BRANCH_ON_WEBLATE\" \"$1-$BRANCH_SLUT\" \"$(weblatelink_repo doc-sle MAIN.opensuse-$BRANCH_SLUT)\" \"\" \"\" $BRANCH_ON_GITHUB \"50-pot/$1.pot\" \"*/po/$1.*.po\"" >&3
+	fi
+}
+
+if test -f "weblate-create-projects-oneshot.sh" ; then
+	cd ..
+fi
+if ! test -f "50-tools/weblate-create-projects-oneshot.sh" ; then
+	echo "Please call this script from yast-translations top directory."
```
sknorr: Could you fix the error message here?

50-tools/weblate-create-projects-oneshot.sh
```diff
@@ -0,0 +1,66 @@
...
+	exit 1
+fi
+WORKDIR=$PWD
+
+. 50-tools/weblate-functions.inc
+
+exec 3>50-tools/weblate-create-projects-oneshot-run.sh
+cat >&3 <<EOF
+#!/bin/bash
+echo "This script was already executed. If cannot be started any more." >&2
```
sknorr: If -> It

50-tools/weblate-create-projects-oneshot.sh
```diff
@@ -0,0 +1,66 @@
...
+exit 1
+if test -f "weblate-create-projects-oneshot-run.sh" ; then
+	cd ..
+fi
+if ! test -f "50-tools/weblate-create-projects-oneshot-run.sh" ; then
+	echo "Please call this script from yast-translations top directory."
```
sknorr: Please fix the error message.

50-tools/weblate-functions.inc
```diff
@@ -0,0 +1,178 @@
+#!/bin/bash
+
+set -o errexit
+
+function weblate_django_shell() {
+	if test -z "$(which weblate-ssh)" ; then
+		echo "This command needs ssh access to l10n.opensuse.org server." >&2
+		echo "Please install/create weblate-ssh script wrapper." >&2
```
sknorr: Question: Do I need this tool? Where do I get it from?

svenseeberg:
@Gianvacca if you've got time for the branch checkout, yes please. I guess this is useful for preventing accidents.

And @sknorr found an interesting word/typo in https://github.com/SUSE/doc-sle/pull/358/files#diff-cd67d20e19734e2cf6344e5b8b9f8b13R9
The slut is probably supposed to be a slug?

50-tools/weblate-create-projects-oneshot.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
+
+# This script generates initial projects import script. It was used only once
+# and it could be called never more. Once regular projects maintenance will be
+# implemented, it could be deleted. This is just a source of needed code.
```
sknorr: What specifically do you need the script for? Why is it not enough to just create a new project on the Weblate Web interface?

Gianvacca: @sknorr If it's simpler to set things up in Weblate Web interface, I think we can trash it. I found it in my research for projects translated via Weblate and I imported it in my repo for future evaluation.

svenseeberg: @sknorr I fixed your comments above for now. However, I can drop it from the PR if you want.

50-tools/weblate-functions.inc
Gianvacca: @sknorr We could contact the author [Stanislav Brabec](https://github.com/stanislav-brabec) for explanations about it. What do you think?

sknorr: @stanislav-brabec What is the purpose of the whole weblate-create-projects script? Where would I get the weblate-ssh tool from?

50-tools/weblate-create-projects-oneshot.sh
stanislav-brabec: weblate-create-projects-oneshot.sh script creates initial setup for large Weblate projects with many projects and/or subprojects.

I submitted the weblate-create-projects-oneshot.sh script to several projects to document the process of the initial import. However it is no more useful for the project itself, the code can be recycled to create a new large project with many subprojects.

Looking at your project, it would be very useful to use such script, as it will save hours of clicking. Additionally, you would probably need a script to update projects. List of translation domains will probably change over the time, and it is always simpler (for me) to call a script instead of comparing lists manually.

You can add it to the pull request, and I will update it once the initial import will succeed, or you can mail it to me, and I will make a pull request with the script that worked.

However you can write and push such script, you cannot run it. To use it, you need openSUSE Heroes account, openSUSE VPN access, root access to the Weblate server and admin privileges in the Weblate. I can do run it for you.

Please decide about a structure of the Weblate project:

- You can have one project doc-sle with many components:
doc-sle/adm_support-develop
doc-sle/adm_support-sle-15-sp1
doc-sle/adm_wbem-develop
doc-sle/adm_wbem-sle-15-sp1

- Or you can have many projects, each with few components.
doc-sle-adm_support/develop
doc-sle-adm_support/sle-15-sp1
doc-sle-adm_wbem/develop
doc-sle-adm_wbem/sle-15-sp1

From translators' perspective, there is only a visual difference. But there is a big difference in behavior: components of one project share equal strings. If there will be many projects setup, only components inside one project share equal strings.

weblate-ssh is just a wrapper that calls correct ssh command to connect to the Weblate server. It was not published as it depends on openSUSE intranet changes. I will consult whether I can publish it, as it contains sensitive information. But anyway, you have not the correct SSH key to get in. Use of Django Shell works around limitations of Weblate API (in time of writing the script).

Gianvacca:
Hi guys! So, after Stanislav answer, has there been any progress? Any preference on the approach to follow?

svenseeberg:
AFAICT it does not hurt to merge for now. The script can be dropped later. It does not procude that many overhead-kilobytes.

stanislav-brabec:
@svenseeberg Surely, it does not block the merge.

But the Weblate component structure should be decided before performing the import to the Weblate (to make it available for translators).

I will pick the script and adjust as needed just before the import.

svenseeberg:
Okay, I'll merge now. @Gianvacca @sknorr @stanislav-brabec can you please connect and agree on the Weblate integration? I'm no expert here.

Gianvacca:
@stanislav-brabec @svenseeberg I think that maybe one project doc-sle with many components is the best solution. As a translator, @guoyunhe, what do you think?
@svenseeberg will you create a develop branch also in the subproject?

guoyunhe:
Dividing into components is good for translators to select which one they want to translate. Some articles are more important and can be translated first.

svenseeberg:
@Gianvacca `develop` branch is pushed.

guoyunhe:
When will it be available on Weblate?

Gianvacca:
@guoyunhe I'm sorry, I'm not sure I understood what is your preference: 1 or 2? Personally I am for number 1.
> 
> **1.** You can have one project doc-sle with many components:
>     doc-sle/adm_support-develop
>     doc-sle/adm_support-sle-15-sp1
>     doc-sle/adm_wbem-develop
>     doc-sle/adm_wbem-sle-15-sp1
> 
> **2.** Or you can have many projects, each with few components.
>     doc-sle-adm_support/develop
>     doc-sle-adm_support/sle-15-sp1
>     doc-sle-adm_wbem/develop
>     doc-sle-adm_wbem/sle-15-sp1

guoyunhe:
@Gianvacca I think solution 1 is better. Solution 2 will make our project list extremely long and make translators harder to find what they want to translate.

Gianvacca:
@guoyunhe Great!
@stanislav-brabec Hi Stanislav, so please review [script weblate-create-projects-oneshot.sh](https://github.com/SUSE/doc-sle/blob/develop/50-tools/weblate-create-projects-oneshot.sh) according to chosen structure and send a pull request. Many thanks.

stanislav-brabec:
@Gianvacca Should I run the script and create a translation project on [openSUSE Weblate](https://l10n.opensuse.org/)?

I will create one project with many components that share translations. (Examples of this model: [desktop-file-translations](https://l10n.opensuse.org/projects/desktop-file-translations/), [packages-i18n](https://l10n.opensuse.org/projects/packages-i18n/).

Do you want to apply some restrictions (e. g. only contracted translators can translate)?

Gianvacca:
@svenseeberg Should @stanislav-brabec proceed with the creation of the translation project (one project with many components that share translations, I confirm) for the **develop** branch?
I would do it without restrictions.
However I would first merge [pull request 1 of doc-sle-translations](https://github.com/openSUSE/doc-sle-translations/pull/1) subproject.

svenseeberg:
I would refer this to @sknorr . He's got more experience with Weblate than I do.

Gianvacca:
@sknorr Hi Stefan, do you think we are ready to create the Weblate project?

Gianvacca:
@sknorr Hi Stefan, any progress on this?

guoyunhe:
Hi all, is here any progress for this? Can we start the Weblate integration?

stanislav-brabec:
@guoyunhe Could somebody add me to a list of users who can see the project and opensuse-i18n to a list of users who can commit?

PR #361, opened 2018-10-30
Add Pull Request template

svenseeberg:
This adds a template to Pull Request with the following check boxes:

- [ ] Minor edit
- [x] Doc Update section is added (if not minor edit)
- [x] Checked if backport applies to maintenance/SLE12
- [x] Checked if backport applies to maintenance/SLE12SP1
- [x] Checked if backport applies to maintenance/SLE12SP2
- [x] Checked if backport applies to maintenance/SLE12SP3
- [x] Checked if backport applies to maintenance/SLE12SP4

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,14 @@
+### Description
+A few sentences describing the overall goals of the pull request's
+commits. Refer to the Bugzilla or FATE number if available.
+
+### Checks
+Check all of the items below.
+
+- [ ] Minor edit
+- [ ] Doc Update section is added (if not minor edit)
+- [ ] Backport required for maintenance/SLE12
+- [ ] Backport required for maintenance/SLE12SP1
+- [ ] Backport required for maintenance/SLE12SP2
+- [ ] Backport required for maintenance/SLE12SP3
+- [ ] Backport required for maintenance/SLE12SP4


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
+# Description
```
tomschr: Maybe you can add an additional line which references back to an issue; something like this:

```
This PR fixes #...
```

Apart from that, it looks great! :+1:

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+A few sentences describing the overall goals of the pull request's commits.
+
+# Checks
```
tomschr: Not sure if "Checks" is the right word here. Maybe "Backports"? "Requirements"?

svenseeberg: I literally meant it to be "Checks". Things the person opening the PR has to check. Renaming it to "I have checked if ..." would be possible.

.github/PULL_REQUEST_TEMPLATE.md
sknorr: * Mmh, shouldn't the ID of the issue fixed rather be in the commit message?

* Minor note, could you reduce the size of these headlines? I.e. `###` instead of`#`, the current size may be semantically correct but will just look jarring.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+Check all check marks that apply to this pull request. It is not
```
sknorr: "Check all items" mebbe?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,14 @@
...
+necessary check all of them.
+
+- [ ] Minor edit
+- [ ] Doc Update section is added, if not minor edit
+- [ ] Checked if backport applies to maintenance/SLE12
```
sknorr: I am not 100% sure how to interpret these boxes: mebbe use "Backport to ... needed" (if that is a the correct interpretation)?

.github/PULL_REQUEST_TEMPLATE.md
svenseeberg: The ID should be in the commit message, yes. But why not additionally mention it here? Makes it easier for reviewers.

PR #362, opened 2018-10-31
Changes to account for new directory tree

Gianvacca:
Modifications to Makefile and po-selector script to take into account the new directory tree.
Some useless legacy stuff was also wiped from the Makefile.


svenseeberg:
Thanks @Gianvacca . I amended the changes here to the commit in pr #358 . Additionally, I flattened all tabs to single white spaces. This seemed to be consistent with the overall indentation.

Gianvacca:
Hi @svenseeberg actually tabs have a precise meaning in Makefiles. They can't be substituted by white spaces. I used white spaces fore mere indentation (in functions or if blocks), but tabs are for recipes, [they have a functional role](https://www.gnu.org/software/make/manual/html_node/Recipe-Syntax.html).

PR #369, opened 2018-11-09
Nvdimm fix

lproven:
### Description
The (belated) technical review from jthumshirn@suse.de of the NVDIMM section in the SLE15 Admin Guide requested several alterations that also affect SLE12 SP4.
Amendment: replace "filesystem" with "file system" throughout, as per pre-existing edited version and style guide.

### Checks
Check items that apply.

- [x] Minor edit (does not required Doc Update)
- [ ] Doc Update section has been added in separate commit
- [ ] Backport required for maintenance/SLE12
- [ ] Backport required for maintenance/SLE12SP1
- [ ] Backport required for maintenance/SLE12SP2
- [ ] Backport required for maintenance/SLE12SP3
- [ ] Backport required for maintenance/SLE12SP4

### DIFF
--- a/xml/hardware_nvdimm.xml
+++ b/xml/hardware_nvdimm.xml
@@ -47,19 +47,16 @@
    However, some types of NVDIMMs already existed before the standard was
    defined. These can be used in the same way.
   </para>
-  
+
   <para>
    Different models use different forms of electronic storage medium, such as
-   Intel 3D XPoint, or a combination of NAND flash and DRAM, and in future,
-   memristor storage or other forthcoming types of non-volatile memory.
+   Intel 3D XPoint, or a combination of NAND flash and DRAM. New forms of
+   non-volatile RAM are also in development. This means that different vendors'
+   NVDIMMs offer different performance and durability characteristics.
+   Typically, they are slower than DRAM, but considerably faster than SSDs based
+   on NAND-flash technology.
   </para>
-
-  <para>
-   This means that different vendors' NVDIMMs offer different performance and
-   durability characteristics. Typically, they are slower than DRAM, but
-   considerably faster than SSDs based on NAND-flash technology.
-  </para>
-
+  
   <para>
    Because the storage technologies involved are in an early stage of
    development, different vendors' hardware may impose different limitations.
@@ -112,10 +109,10 @@
     <term>Namespace</term>
     <listitem>
      <para>
-      A single contiguously-address range of non-volatile storage, comparable to
-      those on NVM Express PCIe SSDs or to a SCSI Logical Unit (LUN). Each
-      NVDIMM usually appears in the server's <filename>/dev</filename> directory
-      as a separate block device. This storage must be assigned to one or more 
+      A single contiguously-address range of non-volatile storage, comparable
+      to those on NVM Express SSDs or to a SCSI Logical Unit (LUN). Each NVDIMM
+      usually appears in the server's <filename>/dev</filename> directory as a
+      separate block device. This storage must be assigned to one or more
       namespaces for it to be used. Depending on the method of access required,
       namespaces can either amalgamate storage from separate NVDIMMs into larger
       volumes, or allow it to be partitioned into smaller volumes, similarly to
@@ -146,7 +143,7 @@
        <term>sector</term>
        <listitem>
         <para>
-         For legacy file systems which do not checksum metadata. Suitable for
+         For legacy filesystems which do not checksum metadata. Suitable for
          small boot volumes. Compatible with other operating systems.
         </para>
        </listitem>
@@ -155,7 +152,7 @@
        <term>fsdax</term>
        <listitem>
         <para>
-         File system-DAX mode. Default if no other mode is specified. Creates a
+         Filesystem-DAX mode. Default if no other mode is specified. Creates a
          block device (<filename>/dev/pmem<replaceable>X</replaceable>
           [.<replaceable>Y</replaceable>]</filename>) which supports DAX for
          <literal>ext4</literal> or <literal>XFS</literal>.
@@ -168,7 +165,7 @@
         <para>
          Device-DAX mode. Creates a single-character device file (<filename>
           /dev/dax<replaceable>X</replaceable>.<replaceable>Y</replaceable>
-         </filename>). Does <emphasis>not</emphasis> require file system
+         </filename>). Does <emphasis>not</emphasis> require filesystem
          creation. Suitable to register PMEM for RDMA or for assigning it to
          virtual machines, or for mapping blocks too large to fit into RAM.
         </para>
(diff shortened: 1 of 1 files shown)


lproven:
OK, I have a "LGTM" from @jankara on this, so this should be good to go.

xml/hardware_nvdimm.xml
```diff
@@ -38,33 +39,29 @@
...
-   However, some types of NVDIMMs already existed before the standard was
-   defined. These can be used in the same way.
-  </para>
-  
-  <para>
-   Different models use different forms of electronic storage medium, such as
-   Intel 3D XPoint, or a combination of NAND flash and DRAM, and in future,
-   memristor storage or other forthcoming types of non-volatile memory.
+   However, NVDIMMs already existed before the standard was defined, meaning
+   that non-compliant hardware exists. These can be used in the same way.
```
taroth21: As the sentence before ends with 'non-compliant hardware' exists, I stumbled across the demonstrative pronoun  'These' on first reading, wondering what it refers to. Maybe rephrase the first or the second sentence to make the reference clear at first reading?

xml/hardware_nvdimm.xml
```diff
@@ -38,33 +39,29 @@
...
 
   <para>
-   This means that different vendors' NVDIMMs offer different performance and
-   durability characteristics. Typically, they are slower than DRAM, but
-   considerably faster than SSDs based on NAND-flash technology.
+   Different models use different forms of electronic storage medium, such as
+   Intel 3D XPoint, or a combination of NAND flash and DRAM. New forms of
+   non-volatile RAM are also in development. This means that different vendors'
+   NVDIMMs offer different performance and durability characteristics.
+   Typically, they are slower than DRAM, but considerably faster than SSDs based
```
taroth21: Same problem here: 'they' could either refer to the 'durability characteristics' or to 'NVDIMM'. For experts, this is obvious but maybe not for the translators. Rephrase?

xml/hardware_nvdimm.xml
```diff
@@ -38,33 +39,29 @@
...
+   on NAND-flash technology.
   </para>
-
+  
   <para>
    Because the storage technologies involved are in an early stage of
    development, different vendors' hardware may impose different limitations.
-   However, generally, writing data to NVDIMMs is slower than reading from
-   them, and the number of write cycles is generally limited.
+   However, generally, writing data to persistent memory is slower than reading
```
taroth21: Do you need both modifications here ('However' and 'generally')? How about the following: 'Generally, writing data to persistent memory is slower than reading from it, and the number of write cycles is limited.'

xml/hardware_nvdimm.xml
```diff
@@ -220,22 +214,20 @@
...
-       The <literal>ext2</literal> file system does not support DAX.
-      </para>
-     </note>
     </listitem>
    </varlistentry>
    <varlistentry>
     <term>Direct Access (DAX)</term>
     <listitem>
      <para>
-      Directly <literal>mmap()</literal> NV-RAM into a process' address space.
```
taroth21: Is ```<literal>mmap()</literal>``` used as verb here? If yes, it sounds like technical jargon to me (maybe I'm also lacking context here because I did not read the complete chapter). Could you perhaps also specify what   ```<literal>mmap()</literal>``` is? A function?

xml/hardware_nvdimm.xml
```diff
@@ -285,9 +277,9 @@
     </title> 
     <para>
      In cases where the server will be hosting an application that can directly
-     access large amounts of storage on a byte-by-byte basis, by 
-     <literal>mmap</literal>ing the storage directly into the application's
-     address space.
+     access large amounts of fast storage on a byte-by-byte basis, by 
+     <literal>mmap</literal>ing DAX storage directly into the application's
```
taroth21: Same here: ```<literal>mmap</literal>ing```-  I would avoid using this as a verb, because it might be hard to translate into other languages.

xml/hardware_nvdimm.xml
```diff
@@ -304,7 +296,8 @@
   <sect2>
    <title>PMEM with BTT</title>
    <para>
-    This is most useful when you just want to use NVDIMMs as very fast storage.
+    This is most useful when you just want to use the persistent memory on a set
```
taroth21: remove 'just' (style guide)

xml/hardware_nvdimm.xml
lproven: Good point. I have reworded and shortened this paragraph.

xml/hardware_nvdimm.xml
lproven: See comment above. I would value your comments on the new paragraph about performance characteristics.

xml/hardware_nvdimm.xml
lproven: A fair point. mmap() is a Unix system call. I have removed this reference and explained it in words instead.

xml/hardware_nvdimm.xml
lproven: Excellent point. I have rephrased the sentence to explain what mmap is and to remove the usage as a verb.

xml/hardware_nvdimm.xml
lproven: Fair point! I have simplified this sentence slightly.

xml/hardware_nvdimm.xml
```diff
@@ -38,33 +39,34 @@
...
    development, different vendors' hardware may impose different limitations.
-   However, generally, writing data to NVDIMMs is slower than reading from
-   them, and the number of write cycles is generally limited.
+   Thus, the following statements are generalizations.
+  </para>
+  
+  <para>
+   Persistent memory is up to ten times slower than DRAM, but around a thousand
+   times faster than flash storage. Like SSDs, writing is slower than reading,
+   and it can only handle a finite number of rewrite cycles. However, most forms
```
taroth21: The sentence 'Like SSDs, ...'  is a bit ambiguous. Does the limitation ' can only handle a finite number of review cycles' apply to both persistent memory *and* to SSDs? If yes, maybe rephrase the sentence to:
' Like (with?) SSDs, writing to persistent memory is slower than reading from it it, and both SSDs and persistent memory can only handle a finite number of rewrite cycles. ' 
(not sure about adding 'with' - up to you to decide as native speaker) :)

lproven: This one is more difficult. I have used the phrase "Like SSDs" 3 times, in 2 places. I did not realize that it could be ambiguous. I will rewrite both sections that compare the 2 forms of storage and attempt to remove any ambiguity.

lproven: This one is more difficult. I have used the phrase "Like SSDs" 3 times, in 2 places. I did not realize that it could be ambiguous. I will rewrite both sections that compare the 2 forms of storage and attempt to remove any ambiguity.

PR #387, opened 2019-01-14
Re-implement fix for BSC #1119551,

lproven:
Working against develop for easier back-porting (I did it against SLE12 SP4 last time, and messed it up.) 

I also noticed a missing parenthesis in the preceding paragraph, and fixed that and the wording around it.

### Description
Move the slash in the closing <self_update> tag into the correct position. As per:
https://bugzilla.suse.com/show_bug.cgi?id=1119551

### Checks

*Is a Doc Update section necessary and has it been created?*

- [ ] Minor edit without associated FATE/Bugzilla: no Doc Update required
- [ ] All other edits: Doc Update section has been added, in a separate commit

*Are backports required?*

- [x] To maintenance/SLE12SP3
- [x] To maintenance/SLE12SP4
- [x] To maintenance/SLE15SP0

### DIFF
--- a/xml/ay_bigfile.xml
+++ b/xml/ay_bigfile.xml
@@ -1083,7 +1083,7 @@ exit 0
     During the installation, &yast; can update itself to solve bugs in the
     installer that were discovered after the release. Refer to the &deploy;
     for further information about this feature. Use the following tags to
-    configure &yast; self-update:
+    configure the &yast; self-update:
    </para>
 
    <variablelist>
--- a/xml/ay_bigfile.xml
+++ b/xml/ay_bigfile.xml
@@ -1077,7 +1077,7 @@ exit 0
   </sect2>
 
   <sect2 xml:id="CreateProfile.General.self_update">
-   <title>The Self Update Section</title>
+   <title>The Self-Update Section</title>
 
    <para>
     During the installation, &yast; can update itself to solve bugs in the
@@ -1111,7 +1111,7 @@ exit 0
      </term>
      <listitem>
       <para>
-       Location of the update repository to use during &yast; self-update.
+       Location of the update repository to use during the &yast; self-update.
        For more information, refer to the &deploy;.
       </para>
       <important>
--- a/xml/ay_bigfile.xml
+++ b/xml/ay_bigfile.xml
@@ -1093,11 +1093,11 @@ exit 0
      <listitem>
       <para>
        This option allows to enable (set to <literal>true</literal>) or
-       disable (set to <literal>false</literal> the self-update. Setting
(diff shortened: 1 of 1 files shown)


xml/ay_bigfile.xml
```diff
@@ -1093,11 +1093,11 @@ exit 0
      <listitem>
       <para>
        This option allows to enable (set to <literal>true</literal>) or
-       disable (set to <literal>false</literal> the self-update. Setting
-       this value is optional. The default is <literal>true</literal>.
+       disable (set to <literal>false</literal>) &yast; from self-updating.
```
cwickert: I'm not a native speaker, but I wonder if "enable/disable from self-updating" is good English. Shouldn't it be "prevent from"? 

What's so bad about the current wording? We could elaborate it with "YaST self-update" or "installer self-update" instead of just "self-update", but unless there is a reason to change it, I would stick to the current wording with 'enable' and 'disable'.

xml/ay_bigfile.xml
```diff
@@ -1093,11 +1093,11 @@ exit 0
...
+       Setting this value is optional. The default is <literal>true</literal>.
       </para>
 <screen><![CDATA[<general>
- <self_update config:type="boolean">true<self_update/>
+ <self_update config:type="boolean">true</self_update>
```
cwickert: LGTM

xml/ay_bigfile.xml
lproven: OK... how about 
"enable (set to true) or disable (set to false) the YaST self-update?"
I though just saying "enable or disable self-update" was ambiguous.

xml/ay_bigfile.xml
```diff
@@ -1080,10 +1080,10 @@ exit 0
...
 
    <para>
-    During the installation, &yast; is able to update itself to solve bugs in
-    the installer that were discovered after the release. Refer to the &deploy;
-    for further information about
-    this feature. Use the following tags to configure the self update:
+    During the installation, &yast; can update itself to solve bugs in the
+    installer that were discovered after the release. Refer to the &deploy;
+    for further information about this feature. Use the following tags to
+    configure &yast; self-update:
```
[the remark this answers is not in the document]
lproven: OK, done.

cwickert:
There was still an occurrence of 'self update' (without hyphen), which I just fixed. Once travis finishes, I will squeeze everything.

xml/ay_bigfile.xml
cwickert: I don't really think it's ambiguous, but adding 'YaST' certainly won't hurt.

PR #400, opened 2019-02-14
Reword linuxrc coverage (BSC#1110431)

lproven:
Purpose is to make it explicit that syntax is not shared with autoyast

I have also formatted the name as a command where it is not capitalised or in a title, and made some minor copy edits, including addressing earlier remarks.

### Description

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE15SP0

### DIFF
--- a/xml/ay_bigfile.xml
+++ b/xml/ay_bigfile.xml
@@ -9996,10 +9996,10 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
 
      <para>
       It is possible to specify <literal>rootpassword</literal> in
-      <command>linuxrc</command> and also have a user section for the root user. If this
-      section is missing the password, then the password from <command>linuxrc</command>
-      will be used. Passwords in profiles take precedence over <command>linuxrc</command>
-      passwords.
+      <command>linuxrc</command> and also have a user section for the &rootuser;
+      user. If this section is missing the password, then the password from
+      <command>linuxrc</command> will be used. Passwords in profiles take
+      precedence over <command>linuxrc</command> passwords.
      </para>
     </note>
 
@@ -16208,8 +16208,10 @@ echo -n 100
     </para>
     <para>
      You can pass keywords to <command>linuxrc</command> using the kernel
-     command line, either interactively at boot time, when creating 
-     network-bootable images, or using a specially-configured DHCP server in
+     command line. This can be done in several ways. As is normal for kernel
+     parameters, you can specify them interactively at boot time. You can also
+     insert kernel parameters into custom network-bootable disk images. It is 
+     also possible to configure a DHCP server to pass kernel parameters in
      combination with Etherboot or PXE.
     </para>
     <para>
@@ -17161,23 +17163,25 @@ mv /var/run/zypp.sav /var/run/zypp.pid
   <title>Advanced Linuxrc Options</title>
   <info/>
   <para>
-   <command>Linuxrc</command> is a small program that runs after the kernel loads, but before &ay;
-   or other stages. It prepares the system for installation. It allows the user
-   to load modules, start an installed system or a rescue system, and to guide
-   the operation of &yast;.
+   <command>Linuxrc</command> is a small program that runs after the kernel has
+   loaded, but before &ay; or other stages. It prepares the system for
+   installation. It allows the user to load modules, start an installed system
+   or a rescue system, and to guide the operation of &yast;.
   </para>
   <note>
    <title>Linuxrc Settings Are Not The Same As &ay; Settings</title>
    <para>
     Some <command>linuxrc</command> settings have the same names as settings
-    used by &ay; in the <filename>autoyast.xml</filename> file. Beware that this
-    does not mean that they take the same parameters or function in the same
-    way. For example, &ay; takes a parameter <command>self_update</command>
-    and if this is set, another called <command>self_update_url</command>.
-    Linuxrc also has a <command>self_update</command> parameter, but this takes
-    either <literal>0</literal> or a URL. Attempting to pass &ay; parameters to
-    <command>linuxrc</command> will almost certainly not give the desired 
-    results.
+    used by &ay; in the <filename>autoyast.xml</filename> file. This does not
+    mean that they take the same parameters or function in the same way. For
+    example, &ay; takes a parameter <command>self_update</command>. If this
+    parameter is set, another one, <command>self_update_url</command> will be
+    read and followed. Linuxrc also has a <command>self_update</command> parameter, but this takes
+    either <literal>0</literal> or a URL.
+   </para>
+   <para>
+    Do not pass &ay; parameters to <command>linuxrc</command>, as this will
+    almost certainly not give the desired results.
    </para>
   </note>
   <para>
@@ -17766,7 +17770,7 @@ mv /var/run/zypp.sav /var/run/zypp.pid
     Even if parameters like <literal>hostip</literal>,
     <literal>nameserver</literal>, and <literal>gateway</literal> are passed to
     <command>linuxrc</command>, the network is only started when it is needed
-    (for example, when installing via SSH or VNC). Since <literal>autoyast</literal>
+    (for example, when installing via SSH or VNC). Because <literal>autoyast</literal>
     is not a <command>linuxrc</command> parameter (this parameter is ignored by
     <command>linuxrc</command> and only passed to &yast;), the network will
     <emphasis>not</emphasis> be started automatically when specifying a remote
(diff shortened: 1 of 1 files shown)


xml/ay_bigfile.xml
```diff
@@ -9923,7 +9923,7 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
     <para>
      A list of users can be defined in the <literal>&lt;users&gt;</literal>
      section. To be able to log in, make sure that either the &rootuser; users are
-     set up or rootpassword is specified as a linuxrc option.
+     set up or rootpassword is specified as a <command>linuxrc</command> option.
```
taroth21: is 'rootpassword' the option? then best tag it: ```<option>rootpassword</option>```

xml/ay_bigfile.xml
```diff
@@ -9996,9 +9996,9 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
 
      <para>
       It is possible to specify <literal>rootpassword</literal> in
-      linuxrc and also have a user section for the root user. If this
```
taroth21: we also have an entity for the user 'root' that tags the word 'root' as systemitem -> "for the &rootuser; user"
(unfortunately, it is a bit clumsy to use in the text as the entity as such is called 'rootuser' but only resolves to 'root' which makes it awkward in the source text because you have to add the word 'user' afterwards)

xml/ay_bigfile.xml
```diff
@@ -16196,24 +16196,21 @@ echo -n 100
...
     </para>
     <para>
-     All linuxrc keywords can be passed to linuxrc using the kernel command
-     line. The command line can also be set when creating network bootable
-     images or it can be passed to the kernel using a specially configured
-     DHCP server in combination with Etherboot or PXE.
-     <remark>emap 2011-11-07:Obscure paragraph. What is set in the command
-     line? And what is the 'it' that can be passed to the kernel? Probably not
-     the command line itself.</remark>
+     You can pass keywords to <command>linuxrc</command> using the kernel
```
taroth21: Hm, not sure I like the new version better, because it contains a lot of juxtapositions whose relationship to each other is not apparent. Could you try to split into two sentences (according to units of thoughts)?

xml/ay_bigfile.xml
```diff
@@ -17082,16 +17078,16 @@ mv /var/run/zypp.sav /var/run/zypp.pid
...
       signed</literal>. I need to manually interact.
      </para>
     </question>
     <answer>
      <para>
-      Linuxrc found some unsigned file (like a driver update). To
-      use an unsigned file, you can suppress that message by passing
-      <literal>insecure=1</literal> to the linuxrc parameter list (together
-      with the <literal>autoyast=...</literal> parameter).
+      <command>Linuxrc</command> found an unsigned file, such as a driver
```
taroth21: If tagged as 'command', it should be spelled/spelt in lowercase, even if it is located at the beginning of a sentence.

xml/ay_bigfile.xml
```diff
@@ -17165,44 +17161,65 @@ mv /var/run/zypp.sav /var/run/zypp.pid
   <title>Advanced Linuxrc Options</title>
   <info/>
   <para>
-   Linuxrc is a program used for setting up the kernel for installation
-   purposes. It allows the user to load modules, start an installed system,
-   a rescue system or an installation via &yast;.
+   <command>Linuxrc</command> is a small program that runs after the kernel loads, but before &ay;
```
taroth21: spelling of 'linuxrc')? see comment above...

In this case, I would prefer present perfect: "... that runs after the kernel has loaded..."

xml/ay_bigfile.xml
```diff
@@ -17165,44 +17161,65 @@ mv /var/run/zypp.sav /var/run/zypp.pid
...
+   or other stages. It prepares the system for installation. It allows the user
+   to load modules, start an installed system or a rescue system, and to guide
+   the operation of &yast;.
   </para>
+  <note>
+   <title>Linuxrc Settings Are Not The Same As &ay; Settings</title>
```
taroth21: spelling of linuxrc? I would also tag it as 'command' (even within a title)

xml/ay_bigfile.xml
```diff
@@ -17165,44 +17161,65 @@ mv /var/run/zypp.sav /var/run/zypp.pid
...
+   <para>
+    Some <command>linuxrc</command> settings have the same names as settings
+    used by &ay; in the <filename>autoyast.xml</filename> file. Beware that this
+    does not mean that they take the same parameters or function in the same
+    way. For example, &ay; takes a parameter <command>self_update</command>
+    and if this is set, another called <command>self_update_url</command>.
```
taroth21: another called... -> another one called (?)

xml/ay_bigfile.xml
```diff
@@ -17165,44 +17161,65 @@ mv /var/run/zypp.sav /var/run/zypp.pid
...
+    Linuxrc also has a <command>self_update</command> parameter, but this takes
+    either <literal>0</literal> or a URL. Attempting to pass &ay; parameters to
```
taroth21: I would make it a clear instruction: "Do not pass &ay; parameters to ```<command>linuxrc</command>``` as it will not give the desired results."

xml/ay_bigfile.xml
```diff
@@ -17165,44 +17161,65 @@ mv /var/run/zypp.sav /var/run/zypp.pid
...
+   <command>Linuxrc</command> is a small program that runs after the kernel loads, but before &ay;
+   or other stages. It prepares the system for installation. It allows the user
+   to load modules, start an installed system or a rescue system, and to guide
+   the operation of &yast;.
   </para>
+  <note>
+   <title>Linuxrc Settings Are Not The Same As &ay; Settings</title>
+   <para>
+    Some <command>linuxrc</command> settings have the same names as settings
+    used by &ay; in the <filename>autoyast.xml</filename> file. Beware that this
```
taroth21: Would be more concise (and contain the same information) if you shorten it to: "This does not mean...."

xml/ay_bigfile.xml
```diff
@@ -17165,44 +17161,65 @@ mv /var/run/zypp.sav /var/run/zypp.pid
...
+    because you need to specify a kernel parameter that <command>linuxrc</command>
+    recognizes as well), use <literal>linuxrc=nocmdline</literal>.
    </para>
 
    <para>
-    Linuxrc will always look for and parse a file
-    <filename>/linuxrc.config</filename>. Use this file to change default
-    values if you need to. In general, it is better to use the
-    <filename>info</filename> file instead. Note that
+    Linuxrc will always look for and parse a file called <filename>/linuxrc.config</filename>.
```
taroth21: spelling? ```<command/>``` tag?

xml/ay_bigfile.xml
```diff
@@ -17257,7 +17274,7 @@ mv /var/run/zypp.sav /var/run/zypp.pid
 
 
    <table>
-    <title>Advanced linuxrc keywords</title>
+    <title>Advanced Linuxrc Keywords</title>
```
taroth21: spelling? ```<command/>``` tag?

xml/ay_bigfile.xml
```diff
@@ -17748,16 +17765,17 @@ mv /var/run/zypp.sav /var/run/zypp.pid
    <para>
     Even if parameters like <literal>hostip</literal>,
     <literal>nameserver</literal>, and <literal>gateway</literal> are passed to
-    linuxrc, the network is only started when it is needed (for example, when
-    installing via SSH or VNC). Since <literal>autoyast</literal> is not a linuxrc
-    parameter (this parameter is ignored by linuxrc and only passed to &yast;),
-    the network will <emphasis>not</emphasis> be started automatically when
-    specifying a remote location for the &ay; profile.
+    <command>linuxrc</command>, the network is only started when it is needed
+    (for example, when installing via SSH or VNC). Since <literal>autoyast</literal>
```
taroth21: replace 'since' with 'because' or 'as'? ('since' can have multiple meanings, but you want the causal meaning here)

xml/ay_bigfile.xml
lproven: That is pre-existing text -- I probably just edited it for indentation. But a good catch all the same. Fixed.

xml/ay_bigfile.xml
lproven: Good idea. I have broken this down into one sentence per method.

PR #401, opened 2019-02-15
Feature/jsc#sle 4241 mv resolvconf

taroth21:
@cwickert: Thanks for the detailed review of PR#399! Please have another look if I did not miss any of your comments in this PR.


xml/net_yast.xml
```diff
@@ -783,7 +783,7 @@
      <para>
       In <guimenu>Modify DNS Configuration</guimenu>, select the way the DNS
       configuration (name servers, search list, the content of the
-      <filename>/etc/resolv.conf</filename> file) is modified.
+      <filename>/run/netconfig/resolv.conf</filename> file) is modified.
```
cwickert: Not related to your changes, but this sentence is hard to read due to the text in the parenthesis. The enumeration "name servers, search list, the content of the /etc/resolv.conf file" doesn't make sense anyway. The first too are the settings that are modified while the latter is the file to store them. I think we should just remove the file name here to improve readability.

taroth21: Thanks, changed now.

PR #403, opened 2019-02-21
[WIP] Feature/389 ds ldap server

taroth21:
### Description
PR for William Brown to review the current doc version with regards to **Fate#323362** (Make 389-ds primary ldap server). I added remarks to the following files for the sections where I suspect that changes are required (look for remarks with '@wbrown') . Please comment.

- ay_bigfile.xml
- security_auth.xml
- security_kerberos.xml
- security_ldap.xml 

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP4
- [x] To maintenance/SLE15SP0


xml/security_ldap.xml
```diff
@@ -74,41 +74,12 @@
...
    Unlike NIS, the LDAP service is not restricted to pure Unix networks.
-   Windows&trade; servers (starting with Windows 2000) support LDAP as a
-   directory service. The
-   application tasks mentioned above are additionally supported in non-Unix
-   systems.
-  </para>
-
-  <para>
-   The LDAP principle can be applied to any data structure that needs to be
+   Windows&trade; servers support LDAP as a directory service. The LDAP principle
```
Firstyear: This is a bit confusing as a sentence to me?

xml/security_ldap.xml
```diff
@@ -721,39 +639,10 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
-    store the OpenLDAP configuration anymore. In case of a system upgrade, a
-    copy of the original <filename>/etc/openldap/slapd.conf</filename> file
-    will get created as
-    <filename>/etc/openldap/slapd.conf.YaSTsave</filename>.
-   </para>
-  </tip>
-
   <para>
+   <remark>taroth 2019-02-21: @wbrown: the content of this section needs to be
+   revised for sure - could you please let me know what to remove, replace, or
```
Firstyear: Basically everything. 389-ds works really differently to openldap, so we probably need a completely different approach. Perhaps a good place to start is from my quickstart docs I have made? http://www.port389.org/docs/389ds/howto/quickstart.html

I also think perhaps it may be good to have a list of items we want to have documented. Probably at least:

* Setup/Install
* Basic admin (users, groups, setting up a client with sssd)
* Backup/Restore
* Setting up a replica
* Decommisioning a server?

Is there more? What points are covered here?

Firstyear:
Okay, I've started on some comments. What you have looks reasonable, it seems like a big "delete" of the openldap parts which is fine. I think next is to define what content dot-points/headings you want, and then I can help you fill it in :)

xml/security_ldap.xml
taroth21: Thanks, William! Sounds good to me. Also your quick start is a good basis to start from. :)

 Regarding the LDAP chapter, I could imagine the following structure:
- Structure of an LDAP Directory Tree (keep)
- Configuring the 389-ds LDAP Server (new)
    * Installing the Software
    * Setting Up the Server
    * Administering the Server (users, groups, setting up a client with sssd)
    * Backing Up and Restoring the Database
- Configuring an LDAP Client (new?/ adjust former section?)

Depending on which of these steps can also be done with YaST, we need to describe both approaches for those steps (manually & via YaST). 

Not sure about the topics 'replica' and 'decommissioning' yet, let's discuss this next week. 
I would try to focus on the basic tasks for the chapter. For further configuration tasks, we can probably also refer to the upstream docs.

Firstyear: For "structure", I'm happy for you to use content from my personal blog about this topic https://fy.blackhats.net.au/blog/html/pages/ldap_guide_part_1_foundations.html  (I often hear this is the "best" ldap primer on the internet :) )

Configuration sounds good, and seems pretty similar to the quickstart. We can go into more detail if we want.

I think the "ldap client" setup, is in the quickstart I linked too, and certainly we should expand on that too.

I'm not sure about yast setup (I haven't been involved in that part), but certainly the manual setup should be pretty easy.

I think we do this first, then we'll add replication later? Replication is our main selling point as an LDAP implementation, so it would be great to have something in the suse docs about this.

taroth21: @Firstyear : Actually, regarding the section 'Structure of an LDAP Directory Tree' my thought was to keep the section that we already have in the current chapter. If that is not good enough from your point of view, we can probably add information from your LDAP guide.

About replication: If this is the main selling point, we should certainly cover it in the docs (how soon is probably a question of resources).

xml/security_ldap.xml
```diff
@@ -389,290 +387,57 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
-     Activate other options as necessary:
-    </para>
-    <itemizedlist>
+  <variablelist>
+   <varlistentry>
+    <term><command>dsctl</command></term>
+    <listitem>
+     <para>
+      Manages a local instance and requires &rootuser; permissions. Used for
+      starting, stopping, backing up the database and more.</para>
```
Firstyear: I'd expand on this and say "this requires you to be connected to a terminal which is running the directory server instance".

xml/security_ldap.xml
```diff
@@ -389,290 +387,57 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
+    </listitem>
+   </varlistentry>
+    <varlistentry>
+     <term><command>dsconfig</command></term>
```
Firstyear: I think it's dsconf not dsconfig .

xml/security_ldap.xml
```diff
@@ -389,290 +387,57 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
      <listitem>
       <para>
-       You can <guimenu>Allow users to authenticate via LDAP</guimenu> and
-       <guimenu>Automatically Create Home Directories</guimenu> on the local
-       computer for them.
+       Manages a remote or local instance configuration. Requires
```
Firstyear: I think this is a bit unclear. The best way is to say "manages an instances configuration via it's external interfaces. This allows you to make configuration changes remotely on the instance"

xml/security_ldap.xml
```diff
@@ -389,290 +387,57 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
+    <varlistentry>
+     <term><command>dsidm</command></term>
      <listitem>
       <para>
-       Use <guimenu>Cache LDAP Entries For Faster Response</guimenu> to cache
-       LDAP entries locally. However, this bears the danger that entries can
-       be slightly out of date.
+       Manages content inside of a back-end. The permissions of this tool are
+       granted by access controls. <remark>taroth 2019-02-25: not sure what the
+       following  sentence is referring to? "and can even be used for some
```
Firstyear: So the dsidm tool manages things like users, groups, passwords etc. So some "self-service" actions would include a person being able to reset their own password or change details of their own account with access-controls permitting.

xml/security_ldap.xml
```diff
@@ -389,290 +387,57 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
+  </variablelist>
 
+  <!--taroth 2019-02-25: the following note and para are 'leftovers' of the former
+   section, check with wbrown if we need to keep them-->
   <note>
-<!-- bnc#911409 -->
    <title>LDAP Server Is Part of the Authentication Server</title>
    <para>
+    <remark>taroth 2019-02-25: @wbrown: is this note still valid?</remark>
     Basic LDAP Server initialization and configuration can be done within
```
Firstyear: I'm not sure. I would assume that yast if it can configure a 389-ds instance, would make this comment valid?

xml/security_ldap.xml
```diff
@@ -389,290 +387,57 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
-    <xref linkend="sec.security.ldap.data"/>, start the LDAP server as
-    &rootuser; by entering <command>sudo systemctl start
-    slapd</command>. To stop the server manually, enter the command
-    <command>sudo systemctl stop slapd</command>. Query the status of
-    the running LDAP server with <command>sudo systemctl status
-    slapd</command>.
-   </para>
    <para><remark>taroth 2019-02-21: @wbrown: is the following still true with
     389-ds?</remark>
     Use the &yast; <guimenu>&ycc_runlevel;</guimenu>, described in
```
Firstyear: I don't think so. dscreate will setup your systemd services for you, and yast if it calls dscreate will do the same. So this may not be necessary.

xml/security_ldap.xml
```diff
@@ -682,12 +447,21 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
+     create instance.inf
+     dscreate from-file etc.-->
+  </sect2>
  </sect1>
 <!-- ==================================================================== -->
  <sect1 xml:id="sec.security.ldap.data" os="sles;osuse">
   <title>Manually Administering LDAP Data</title>
-
+<!--taroth 2019-02-25: let's see where to move/integrate the content from this
+    section in the end-->
```
Firstyear: I suspect a lot of this will change as 389-ds supplies the dsidm command, which makes the need for ldapsearch/ldapmodify a rare-occurance. A personal goal in 389-ds is so I never have to see ldapmodify again as it's super hard to use and confusing.

xml/security_ldap.xml
```diff
@@ -889,7 +663,258 @@ telephoneNumber: +49 1234 567-10</screen>
...
   </sect2>
  </sect1>
- <sect1 xml:id="sec.security.ldap.info">
+
+<sect1 xml:id="sec.security.ldap.yast.usergr">
+  <title>Configuring LDAP Users and Groups in &yast;</title>
+<!--taroth 2019-02-25: let's see where to move/integrate the content from this
+    section in the end-->
+  <para>
+   The actual registration of user and group data differs only slightly from
```
Firstyear: This block is a bit confusing to me? I don't really understand what it's trying to say.

xml/security_ldap.xml
```diff
@@ -889,7 +663,258 @@ telephoneNumber: +49 1234 567-10</screen>
...
+   <step>
+    <para>
+     Access the &yast; user administration with <menuchoice>
+     <guimenu>Security and Users</guimenu><guimenu>User and Group
+     Management</guimenu></menuchoice>.
+    </para>
+   </step>
+   <step>
+    <para>
+     Use <guimenu>Set Filter</guimenu> to limit the view of users to the
```
Firstyear: This seems ambiguous. It depends on the pam module, but "limiting the view" could mean "limiting which users have the right to login to this system (authorisation)". I think this should be made clearer.

xml/security_ldap.xml
```diff
@@ -889,7 +663,258 @@ telephoneNumber: +49 1234 567-10</screen>
...
+     LDAP users and enter the password for Root DN.
```
Firstyear: There should never be a need to enter the root DN password, outside of an administration context. Perhaps this means the bind-dn and bind-password for pam/nsswitch?

Firstyear: Maybe I'm misunderstanding, is this yast changing LDAP server data? Or configuring a server to consume LDAP data?

xml/security_ldap.xml
```diff
@@ -889,7 +663,258 @@ telephoneNumber: +49 1234 567-10</screen>
...
+      </imageobject>
+      <imageobject role="fo">
+       <imagedata fileref="yast2_auth_client_ldap.png" width="65%" format="PNG"/>
+      </imageobject>
+     </mediaobject>
+    </informalfigure>
+   </step>
+   <step>
+    <para>
+     Specify one or more LDAP server URLs, host names, or IP addresses under
```
Firstyear: URLS or hostnames only. IP address would break TLS cn/san checking.

xml/security_ldap.xml
```diff
@@ -889,7 +663,258 @@ telephoneNumber: +49 1234 567-10</screen>
...
+     <guimenu>Enter LDAP server locations</guimenu>. When specifying multiple
+     addresses, separate them with spaces.
```
Firstyear: Recommend that the URLS be LDAPS:// only (startTLS and plain both have security issues)

xml/security_ldap.xml
```diff
@@ -889,7 +663,258 @@ telephoneNumber: +49 1234 567-10</screen>
...
+     be <literal>dc=example,dc=com</literal>.
+    </para>
+   </step>
+   <step>
+    <para>
+     If your LDAP server supports TLS encryption, choose the appropriate
+     security option under <guimenu>Secure LDAP Connection</guimenu>.
+    </para>
+    <para>
+     To first ask the server whether it supports TLS encryption and be able
```
Firstyear: LDAPS only here, StartTLS has security risks.

xml/security_ldap.xml
```diff
@@ -889,7 +663,258 @@ telephoneNumber: +49 1234 567-10</screen>
...
+      <para>
+       You can <guimenu>Allow users to authenticate via LDAP</guimenu> and
+       <guimenu>Automatically Create Home Directories</guimenu> on the local
+       computer for them.
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       Use <guimenu>Cache LDAP Entries For Faster Response</guimenu> to cache
+       LDAP entries locally. However, this bears the danger that entries can
```
Firstyear: I think this should be reworded to give an example of what the caching risk really is (and it depends on if they are using nscd or sssd).

xml/security_ldap.xml
```diff
@@ -889,7 +663,258 @@ telephoneNumber: +49 1234 567-10</screen>
...
+       boxes empty to bind anonymously to the server.
+      </para>
+      <warning>
+       <title>Authentication Without Encryption</title>
+       <para>
+        When using authentication without enabling transport encryption
+        using TLS or StartTLS, the password will be transmitted in the clear.
+       </para>
+       <remark>
+        Is the YaST dialog compatible with SASL in any way? - sknorr,
```
Firstyear: SASL is too complicated for a setup like this, not worth including.

xml/security_ldap.xml
```diff
@@ -889,7 +663,258 @@ telephoneNumber: +49 1234 567-10</screen>
...
+        2017-03-01
+       </remark>
+      </warning>
+     </listitem>
+    </itemizedlist>
+    <para>
+     Under <guimenu>Extended Options</guimenu>, you can additionally
+     configure timeouts for BIND operations.
```
Firstyear: This is a client timeout?

xml/security_ldap.xml
taroth21: It also was not entirely clear to me what the sentence was referring to, therefore I removed it, including the following lists and the next para.

xml/security_ldap.xml
taroth21: Added your reply to the remark for the records (and later reference).

xml/security_ldap.xml
taroth21: I think the sentence refers to the fact that you can administer a bunch of different users (local users, system users, NIS users, LDAP users, or custom users) with the same YaST dialog (you select which category to view and edit with 'Set Filter'). But as this information is not relevant for configuring LDAP users, I removed it. I also moved the information that groups can be administered in a similar way to the end of the procedure.

xml/security_ldap.xml
taroth21: AFAIK, you could configure LDAP users on the OpenLDAP server with YaST, yes. YaST is (open)SUSE's graphical tool for system administration tasks. So individual YaST modules are designed to access and alter configuration files, if required, otherwise they could not fulfill this task.

xml/security_ldap.xml
taroth21: That's exactly the question I had. c#13 in Fate#323362 mentions an integration with YaST and according to one of the project managers, yast-auth-server was updated in SLE 15 to handle 389-ds server. But I don't know who integrated this and if it works as expected/described in the docs. I assumed you would know about the YaST part of this feature as well, but I will check with Frederic then.

xml/security_ldap.xml
taroth21: OK, thanks, removed this para.

xml/security_ldap.xml
taroth21: Fixed, thanks.

xml/security_ldap.xml
taroth21: Done

xml/security_ldap.xml
taroth21: @Firstyear : Could you provide an example, please? I lack a bit of context here.

xml/security_ldap.xml
taroth21: Turned remark into a comment and added your answer, just for the records.

xml/security_ldap.xml
taroth21: @Firstyear : I don't know, we maybe need to check with someone from the YaST team who wrote or maintains this module.

xml/security_ldap.xml
taroth21: Added.

xml/security_ldap.xml
taroth21: Fixed typo.

xml/security_ldap.xml
taroth21: Added.

xml/security_ldap.xml
taroth21: Added.

xml/security_ldap.xml
taroth21: Tried to rephrase and added one step. Hopefully clearer now.

xml/security_ldap.xml
taroth21: Added this, thanks.

xml/security_ldap.xml
Firstyear: I think there are some improvements to make it it's clarity, it's super dense at the moment. But perhaps that's a future change? I just worry if this is someones first attempt to understand LDAP, would they read this and understand how it works?  The current content is okay for now (it took me a while to convince github to let me read it).

xml/security_ldap.xml
Firstyear: Yeah, I don't know about the yast side at all. For SUSE, my employment is for upstream focus, and packaging. I haven't used yast. I'd love to be in contact with the person who did the yast integration work though as I think 389-ds is changing a lot and so there may be some "rework" to do.

xml/security_ldap.xml
Firstyear: AHhhhh I see. Okay this makes more sense now if it's yast changing the users. In this case,  Ithink my comments don't mean much :)

xml/security_ldap.xml
Firstyear: Well, it depends what yast does .... anyway. If it's nscd, the cache risk is that when you define an authorisation rule (say, membersOf group admin) can login, and you "remove" a user from that group, the client cache will not see that change until the cache expires (or it refreshes). So there is a time window between the user remove and the affect on the server. The security risk here is that a revoked account can still use the login for a period of time later.

SSSD does not suffer this as it 'caches', but if online, it always checks the group memberships are still valid, so the reflection of the change is instant provided you are online. It's only if sssd is disconnected from the LDAP server that the cache risk exists.

xml/security_ldap.xml
Firstyear: Yep, sounds good.

Firstyear: PS: Maybe we could find them and email as a group because as mentioned, I'd like to talk to them too :)

xml/security_ldap.xml
taroth21: I'll keep this section on the list of ToDos then, but it has rather low priority, because first of all, we need to update the other content of the chapter that no longer applies. 

Regarding readability: As reading longer texts on GH is cumbersome, I will provide a draft PDF of the updated chapter as soon as we have reached a version that makes sense for you to read. :) Until then, you can also have a look at the  HTML output of the content in the develop branch, which was the starting point of this feature branch (see the links to docserv I sent you).

xml/security_ldap.xml
taroth21: Will send details via mail.

xml/security_ldap.xml
taroth21: Thanks for the explanation! Added this information as a warning about a potential security risks with caching.

xml/security_ldap.xml
taroth21: See your inbox. :)

xml/security_ldap.xml
Firstyear: Sounds good, I'll look at those links soon. I agree that we can leave this as low priority :)

xml/security_ldap.xml
Firstyear: Great, thank you!

xml/security_ldap.xml
```diff
@@ -405,11 +409,80 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
-     create instance.inf
-     dscreate from-file etc.-->
+ <procedure>
+  <title>Setting up the &ds389; Instance</title>
+    <para>
+     You create the instance with the <command>dscreate</command> command.
+     It can take a configuration file (<filename>*.inf</filename>) which
+     defines the instance configuration settings. Alternatively, it can be run
+     in an interactive mode. To create a template <filename>*.inf</filename> file,
+     use the <option>create-template</option> option. For more details, see the
```
Firstyear: Sounds good, you can probably add that the template is commented and pre-filled, so you can just read it and tweak variables. :)

xml/security_ldap.xml
```diff
@@ -405,11 +409,80 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
+     man page of <command>dscreate</command>.
+    </para>
+    <para>
+     If you want to set up a trial instance, you can use the following
+     example <filename>*.inf</filename> file as a starting point:
```
Firstyear: I'm worred about adding in a screenshot/file as we may change the content in the future ... but maybe that's a problem if we "get there".

xml/security_ldap.xml
```diff
@@ -405,11 +409,80 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
+
+[backend-userroot]
+sample_entries = yes
+suffix = dc=example,dc=com</screen>
+    </example>
+   <para>
+    Before you use this file in the following steps, edit the
+    <parameter>root_password</parameter> parameter. <remark>taroth 2019-02-27:
+    @wbrown: not sure which password needs to be set for 'root_passord' here
+    (I don't think it's the root password of the machine where this instance is
```
Firstyear: It's the password for the cn=Directory Manager user, aka the directory server root user. IT's a new password that's used for LDAP server administration only

xml/security_ldap.xml
```diff
@@ -405,11 +409,80 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
...
+   </para>
+   <step>
+    <para>
+     To create the &ds389; instance from the <filename>/tmp/instance.inf</filename>
+     above, run:
+    </para>
+    <screen>&prompt.sudo;dscreate from-file /tmp/instance.inf</screen>
+    <para>This creates a working LDAP server.</para>
+    <note>
+     <title>Special Option for Docker</title>
```
Firstyear: Better to not mention docker at all, since I actually just removed -c, and I think better would be if SUSE just a container to ship. Can we revisit docker once that image is produced?

xml/security_ldap.xml
```diff
@@ -820,10 +893,32 @@ Linux,ou=devel,dc=example,dc=com</screen>
...
+       <formalpara>
+        <title>nscd</title>
+        <para>
+         If you define an authorization rule (for example, members of group
+         <literal>admin</literal> can log in), and you remove a user from that
+         group, the client cache will not see that change until the cache expires
+         or refreshes. So a user whose account has been revoked can still log in
+         for a period of time later.
+        </para>
+       </formalpara>
```
Firstyear: Looks good! Thank you!

xml/security_ldap.xml
taroth21: Done.

xml/security_ldap.xml
taroth21: Thanks, rephrased the sentence accordingly.

xml/security_ldap.xml
taroth21: Nothing is carved in stone and I'm aware that updates might be required every now and then. (That's also why I chose to go with the 'basic' example from the upstream docs, because it does not contain too many details).

xml/security_ldap.xml
taroth21: I see, np. Removed the note.

Firstyear:
This all looks good to me. How about we merge this, and then we can continue to review and improve over the next few weeks? I think patches get hard to manage if they go for too long and too big. Does that sound good? 

I think maybe the next step is for me to look at the rendered content, and send some feedback that way to see it all in context.

xml/security_ldap.xml
```diff
@@ -41,15 +36,11 @@
...
    <para>
-    To make multiple concurrent reading accesses possible, the number of
-    updates is usually very low. The number of read and write accesses is
-    often limited to a few users with administrative privileges. In
-    contrast, conventional databases are optimized for accepting the largest
-    possible data volume in a short time.
+    The type of data stored in a directory tends to be long lived and changes
+    infrequently. This allows the service to be optimized for high performance
+    concurrent reads. In contrast, conventional databases are optimized for
+    accepting the largest possible data volume in a short time.
```
Firstyear: accepting a large number of writes to data in a short time?

xml/security_ldap.xml
```diff
@@ -298,6 +272,9 @@ MAY (userPassword $ searchGuide $ seeAlso $ businessCategory <co xml:id="co.ldap
    an example here.
   </para>
 
+  <para>
+   <remark>taroth 2019-03-05: ToDo - adjust calloutlist to example</remark>
```
Firstyear: I can help with this if there are questions of course

xml/security_ldap.xml
taroth21: Argh, overlooked that at first. Now in.

xml/security_ldap.xml
```diff
@@ -510,8 +510,11 @@ tls_cacertdir = /etc/dirsrv/slapd-localhost/ <co xml:id="co.ldap.server.dsrc.rem
     </callout>
     <callout arearefs="co.ldap.server.dsrc.remote.cert">
      <para>
-      <remark>taroth 2019-03-07: @wbrown: is the following correct?</remark>
-      For this to work, copy <filename>/etc/dirsrv/slapd-localhost/ca.crt</filename>
+      If you have configured a certificate yourself, provide the path to the
+      CA certificate here (for remote access with TLS). If you do not have
```
Firstyear: "If you have not configured" would be better english :)

xml/security_ldap.xml
```diff
@@ -510,8 +510,11 @@ tls_cacertdir = /etc/dirsrv/slapd-localhost/ <co xml:id="co.ldap.server.dsrc.rem
...
+      configured a certificate yourself, the server generates a self-signed
```
Firstyear: "the server" -> "dscreate"

xml/security_ldap.xml
```diff
@@ -510,8 +510,11 @@ tls_cacertdir = /etc/dirsrv/slapd-localhost/ <co xml:id="co.ldap.server.dsrc.rem
...
+      certificate which you can trust. In that case, specify
```
Firstyear: "in that case ..." could be better as "It can be found at `location`. Copy it to a readable location or to the client machine you are using ds*  from"

xml/security_ldap.xml
taroth21: Haha, I was wondering which way to put it when I wrote it, thanks!

taroth21:
@Firstyear : Even if this topic is still WIP, I have (manually) merged all the work we did so far into 'develop'  with the following 4 commits:

- https://github.com/SUSE/doc-sle/commit/2a93b09a1d9404bc564e1fc16f6b39f472d5c65a
- https://github.com/SUSE/doc-sle/commit/b2f5e57a671270a39e0b61ef38f4d9aa51936089
- https://github.com/SUSE/doc-sle/commit/f34b352af7bd8d9200e21fb65191dfb8dbaae1ac 
- https://github.com/SUSE/doc-sle/commit/c829b34c7fd31df6eb289aa51f338a1ab3f02494

As 'develop' is the branch in which we will work on the documentation for the (yet to come SLE 15 SP2), I considered it safe to include the changes (we still have enough time to adjust and finish them for SLE 15 SP2). 

However, as discussed before, I will only merge the changes into the maintenance/SLE15SP1 branch after we have finalized the docs (and if the software changes have been pushed to SLE 15 SP1 as a maintenance update).

PR #410, opened 2019-03-20
mention services view in yast sw_single

taroth21:
### Description
Implement Fate#320573: [POST-GM] [R] Add 'services' view to yast2 sw_single

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP4
- [x] To maintenance/SLE15SP0

### DIFF
--- a/xml/art_sle_modules_quick.xml
+++ b/xml/art_sle_modules_quick.xml
@@ -1152,8 +1152,16 @@ Package  Module or Repository
 emacs    Basesystem Module (sle-module-basesystem/15/x86_64)
 emacs    Available</screen>
   </example>
- </sect1>
 
+  <para>
+   If you do not want to search for a certain package, but want to view the list
+   of packages that belong to a module or extension, start the &yast;
+   <guimenu>Software Management</guimenu> module and click <menuchoice>
+    <guimenu>View</guimenu>
+    <guimenu>Services</guimenu>
+   </menuchoice>.
+  </para>
+ </sect1>
 
  <xi:include href="common_copyright_quick.xml"/>
  <xi:include href="common_gfdl1.2_i.xml"/>
--- a/xml/art_sle_modules_quick.xml
+++ b/xml/art_sle_modules_quick.xml
@@ -1154,12 +1154,11 @@ emacs    Available</screen>
   </example>
 
   <para>
-   If you do not want to search for a certain package, but want to view the list
-   of packages that belong to a module or extension, start the &yast;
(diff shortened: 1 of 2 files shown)


xml/yast2_sw.xml
```diff
@@ -275,6 +275,17 @@
       </para>
      </listitem>
     </varlistentry>
+    <varlistentry>
+     <term>Services</term>
+     <listitem>
+      <para>
+       Shows which packages belong to a certain service or module. Select the
```
cwickert: service or module -> module or extension

xml/art_sle_modules_quick.xml
```diff
@@ -1152,8 +1152,16 @@ Package  Module or Repository
 emacs    Basesystem Module (sle-module-basesystem/15/x86_64)
 emacs    Available</screen>
   </example>
- </sect1>
 
+  <para>
+   If you do not want to search for a certain package, but want to view the list
```
cwickert: How about: "Instead of searching for packages, you can also view the list..."?

xml/yast2_sw.xml
```diff
@@ -275,6 +275,17 @@
       </para>
      </listitem>
     </varlistentry>
+    <varlistentry>
+     <term>Services</term>
```
cwickert: Whoever came up with the term "services" for modules and extensions should be punished. :-/

xml/yast2_sw.xml
```diff
@@ -275,6 +275,17 @@
...
+     <listitem>
+      <para>
+       Shows which packages belong to a certain service or module. Select the 
+       module or extension (for example, <literal>Basesystem</literal> or
```
cwickert: If you apply the previous change, "module or extension" would be repeated. :-(
Change to "Select one to display"?

xml/art_sle_modules_quick.xml
taroth21: Much better, thanks.

xml/yast2_sw.xml
taroth21: Yes, adding yet another term does not really help here. ;)

xml/yast2_sw.xml
taroth21: Good catch, thanks.

xml/yast2_sw.xml
taroth21: I decided for 'Select one entry to display...'

PR #418, opened 2019-04-10
Add a NFS section to AutoYaST handbook

joseivanlopez:
### Description

This PR adds information about setting up a NFS drive for AutoYaST installation.

Relevant links:

* https://github.com/yast/yast-storage-ng/pull/896
* https://bugzilla.suse.com/show_bug.cgi?id=1130256

### Checklist
*Check all items that apply.*

*Are backports required?*

- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP4
- [x] To maintenance/SLE15SP0

### DIFF
--- a/xml/ay_bigfile.xml
+++ b/xml/ay_bigfile.xml
@@ -4705,15 +4705,20 @@ openssl x509 -noout -fingerprint -sha256
    <title>NFS Configuration</title>
 
    <para>
-    &ay; allows to install over a Network Filesystem (NFS). For that, you only have to create a drive with CT_NFS type and give the NFS share name (server:path) as device name. The information relative to the mount point is included as part of its first partition section. Note that for an NFS drive, only the first partition is taken into account.
+    &ay; allows to install over a <emphasis>Network File System</emphasis>
+    (NFS). For that, you have to create a drive with the
+    <literal>CT_NFS</literal> type and provide the NFS share name
+    (<replaceable>SERVER:PATH</replaceable>) as device name. The information relative
+     to the mount point is included as part of its first partition section. Note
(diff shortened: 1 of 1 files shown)


cwickert:
Hi @joseivanlopez, the content looks good, but needs some cosmetic fixes. I will take care of them, you shouldn't need to bother with DocBook tags etc.

I'm struggling with the first sentence though: Is this only for the installation or is the configured share available in the installed system, too? I mean, you are not installing **over** NFS (the installation sources are not on NFS) but onto an NFS share, right?

joseivanlopez:
> Hi @joseivanlopez, the content looks good, but needs some cosmetic fixes. I will take care of them, you shouldn't need to bother with DocBook tags etc.
> 

Ok, thanks a lot for taking care.

> I'm struggling with the first sentence though: Is this only for the installation or is the configured share available in the installed system, too? I mean, you are not installing **over** NFS (the installation sources are not on NFS) but onto an NFS share, right?

Yes, the idea is to use an NFS share as "destination" of the installation. Basically, your new system will be installed onto the NFS share instead of a local disk. Nothing special happens regarding the installation sources, they are the same as a regular installation.

joseivanlopez:
@cwickert I have "cherry-picked" your PR. Moreover, after our review meeting today, we have agreed to include some references between both sections related to NFS, see the last commit.

cwickert:
Thanks, these were excellent changes! I just rephrased two sentences a bit to confirm with the way we refer to additional information ("For more information about foo, refer to bar"), but that's was it.

Since I merged this manually, I'm now closing this PR. Thanks a lot for your input!

PR #422, opened 2019-04-29
Add full list of products, short names & supported platforms to Deply't Guide (BSC #1123063)

lproven:
### Description
Addresses https://bugzilla.suse.com/show_bug.cgi?id=1123063

### DIFF
--- a/xml/planning.xml
+++ b/xml/planning.xml
@@ -244,7 +244,7 @@
       <entry>&x86-64;</entry>
      </row>
      <row>
-      <entry>SUSE Manager for Retail</entry>
+      <entry>SUSE Manager Retail Branch Server</entry>
       <entry>SUMA_retail</entry>
       <entry>&x86-64;</entry>
      </row>
--- a/xml/planning.xml
+++ b/xml/planning.xml
@@ -9,8 +9,12 @@
  xmlns:xi="http://www.w3.org/2001/XInclude"
  xmlns:xlink="http://www.w3.org/1999/xlink">
  <title>Planning for &productname;</title>
- <info><abstract>
-   <para> This chapter describes some basic considerations before installing &productname;. </para>
+ <info>
+  <abstract>
+   <para>
+    This chapter describes some basic considerations before installing
+    &productname;.
+   </para>
   </abstract>
   <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
    <dm:bugtracker/>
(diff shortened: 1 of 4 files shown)


cjschroder:
IMO the real problem is using xpointers. For the CAP guides we use entities for repeated content, see https://github.com/SUSE/doc-cap/blob/develop/xml/repeated-content-decl.ent

cjschroder:
Though entities may not be the best choice for a single copy, they are great when you need to repeat the same content multiple times. OTOH I think anything is better than xpointers.

xml/planning.xml
```diff
@@ -163,4 +163,61 @@
...
  </sect1>
+ 
+ <sect1 xml:id="sec.planning.leanos">
+  <title>Changes in installation from &sls; version &product-ga;</title>
+  <para>
+   Starting with &productname; 15, all products are distributed using a single
+   set of installation media.
+  </para>
+  <para>
+   The products available on these unified media are as follows:
```
taroth21: The list of base products is also mentioned in the Installation Quick Start (see https://susedoc.github.io/doc-sle/develop/SLES-installquick/single-html/#sec.sle.installquick.install.lang): "With the Installer you can install all SUSE Linux Enterprise base products: [...]". Please match (for the future: let's move the itemizedlist to an entity and reuse in the Install Quick and Deployment Guide).

xml/planning.xml
```diff
@@ -163,4 +163,61 @@
...
+  </para>
+  <para>
+   To install any of these products, it is only necessary to download the
+   <filename>&slea; &productnumber; Installer DVD 1</filename> ISO image. 
+  </para>
+  <para>If you want to install without network access, additionally download
+   the other two ISO images: <filename>&slea; &productnumber; Installer DVD 2
+   </filename> and <filename>&slea; &productnumber; Installer DVD 3</filename>. 
+  </para>
+ </sect1>
```
taroth21: What should also be mentioned (according to bsc#1123063) is the 'Packages' image (use the &allmodules; entity to make it include the correct product version number). The Packages image is needed if you want to install without access to SCC or a registration server (RMT).

lproven: Good point. I found this: http://download.suse.de/install/SLE-15-Packages-GM/ (not for customer use, just as a reference).

It appears to confirm that there is 1 Packages ISO for each arch, *but* there seem to be 3 ISOs per arch: "SLE-15-Packages-x86_64-GM-DVD1.iso", "SLE-15-Packages-x86_64-GM-DVD2.iso", & ""SLE-15-Packages-x86_64-GM-DVD1.iso".

xml/planning.xml
lproven: That's a different list. I don't know which is authoritative, nor the shorthand names, which are useful in AutoYast and other places. Do you know of anywhere or anyone to ask?

taroth21: I would say that the one in the Install Quick is kind of 'authoritative', because it reflects what customers see in the 'Language, Keyboard and Product Selection' screen that appears during installation, see https://susedoc.github.io/doc-sle/develop/SLES-installquick/single-html/images/install_product_sles.png. Therefore I would advise to use it. However, your sentence about the restriction (not all products are supported on all architectures) makes sense. You can also add the shorthand names to the list if you have them (I agree that they are useful for AY, for example).

lproven: I have checked the list, including the official long and short package names, with Joachim Werner * Jiři Srain. They are now included. The list matches this one, but with more info.

xml/planning.xml
```diff
@@ -163,4 +163,61 @@
    For registering an already installed system, see <xref linkend="sec.yast-software.register"/>.
   </para>
  </sect1>
+ 
+ <sect1 xml:id="sec.planning.leanos">
+  <title>Changes in installation from &sls; version &product-ga;</title>
```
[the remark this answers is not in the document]
lproven: I think I have correctly reverted the changes @taroth21 describes but I would appreciate it if someone could check -- @cwickert or @cjschroder ?

lproven:
Any comments from @cwickert or @cjschroder ?

E.g. I am not sure if I should add new entities for SUMA etc., since we do not identify it by name anywhere else in the docs, as far as I can see.

xml/planning.xml
lproven: Received clarification on Bugzilla from Kai Dupke and Vincent Mouttoussamy -- Packages 2 medium is just source code; never needed for installation. Packages 3 is for beta/internal use only. Clarified role of Packages 2 in docs, and removed mention of Packages 3.

cjschroder:
Yes, you should use entities for any repeated content. (See http://docserv.nue.suse.com/documents/Style_Guide/docu_styleguide/single-html/#sec.entity)

Spelling and formatting look good, and it builds. Technical correctness, sorry, I do not know.

PR #424, opened 2019-05-02
An attempt at making KLP Guide name a little less circular

sknorr:
### Description
Not sure whether this option for the title... It is also not completely sure whether we actually still need this as a separate guide.

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE15SP0

PR #426, opened 2019-05-02
Restore updated chapter title

cjschroder:
Fixes error introduced to https://github.com/SUSE/doc-sle/pull/424/. This PR is a reminder to complete unfinished tasks, and for reviewers to make changes if necessary.

### Description

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP4
- [ X] To maintenance/SLE15SP0

### DIFF
--- a/xml/klp.xml
+++ b/xml/klp.xml
@@ -9,7 +9,7 @@
  xmlns:xi="http://www.w3.org/2001/XInclude"
  xmlns:xlink="http://www.w3.org/1999/xlink"
  xml:id="cha.klp" xml:lang="en">
- <title>Kernel Live Patching (&klpa;) Guide</title>
+ <title>Kernel Live Patching with &klpa;</title>
  <info>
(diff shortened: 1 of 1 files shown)


sknorr:
Fwiw, #424 only changed the book title and did not change the chapter title at all (albeit it did remove two comments that appeared just before that title).

Second, I am not sure that a chapter should be named "... Guide" because that implies it is a standalone document (which of course in one of two contents it is but in the other it is not).

cjschroder:
> Fwiw, #424 only changed the book title and did not change the chapter title at all (albeit it did remove two comments that appeared just before that title).
> 
> Second, I am not sure that a chapter should be named "... Guide" because that implies it is a standalone document (which of course in one of two contents it is but in the other it is not). Probably still better than the current title, so I am not wildly against it. If that helps. :/

I agree it should not be a guide, this turned into a bit of a mess & I am doing cleanup.

PR #441, opened 2019-06-06
[WIP] new transactional-update chapter

cjschroder:
https://jira.suse.com/browse/SLE-4315

### Description

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP4
- [x] To maintenance/SLE15SP0

### DIFF
--- a/xml/transactional_updates.xml
+++ b/xml/transactional_updates.xml
@@ -17,11 +17,10 @@
   <abstract>
    <para>
     Transactional updates are available in &productname; as a technology
-    preview. Transactional updates provide another mechanism for keeping your
-    system updated. Transactional updates are atomic (all updates are applied
-    only if all updates succeed), support rollbacks, and update both read-write
-    and read-only filesystems. It does not affect a running system as no
-    changes are applied until after the system is rebooted.
+    preview, for updating &slsa; when the root filesystem is read-only. 
+    Transactional updates are atomic (all updates are applied only if all 
+    updates succeed) and support rollbacks. It does not affect a running system 
+    as no changes are applied until after the system is rebooted.
    </para>
 
    <para>
@@ -31,7 +30,10 @@
     the root file system in the background, and then fetches updates from the
     release channels. After the new snapshot is completely updated, it is
     marked as active and will be the new default after the next reboot of the
-    system.
+    system. When <command>transactional-update</command> is set to run
+    automatically (which is the default behavior) it also reboots the system.
+    Both the time that the update runs and the reboot maintenance window are 
+    configurable.
    </para>
 
    <para>
@@ -45,19 +47,20 @@
    </para>
   </abstract>
  </info>
+ 
  <sect1 xml:id="sec.tu-limitations">
   <title>Limitations of Technical Preview</title>
 
   <para>
    As a technical preview, there are certain limitations in functionality. The
-   following packages will not work with transactional-update. In the
-   <filename>/srv</filename> directory:
+   following packages will not work with <command>transactional-update</command>. 
+   In the <filename>/srv</filename> directory:
   </para>
 
   <itemizedlist>
    <listitem>
     <para>
-     <package>nginx</package>
+     The <package>nginx</package> default index.html page may not be available
     </para>
    </listitem>
    <listitem>
@@ -81,13 +84,7 @@
   <para>
    In the <filename>/var</filename> directory:
   </para>
-
   <itemizedlist>
-   <listitem>
-    <para>
-     <package>update-alternatives</package>
-    </para>
-   </listitem>
    <listitem>
     <para>
      <package>mpi-selector</package>
@@ -95,22 +92,7 @@
    </listitem>
    <listitem>
     <para>
-     <package>emacs</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
-     <package>flatpak</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
-     <package>net-snmp</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
-     <package>clone-master-clean-up</package>
+     <package>emacs</package> games (Emacs still works)
     </para>
    </listitem>
    <listitem>
@@ -135,32 +117,7 @@
    </listitem>
    <listitem>
     <para>
-     <package>freeipmi</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
-     <package>GeoIP</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
      <package>iso_ent</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
-     <package>krb5-server</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
-     <package>libvpd2(/var/lib/lsvpd/run.vpdupdate)</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
-     <package>nfs-kernel-server</package>
     </para>
    </listitem>
    <listitem>
@@ -170,17 +127,7 @@
    </listitem>
    <listitem>
     <para>
-     <package>openldap2</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
      <package>opensp</package>
-    </para>
-   </listitem>
-   <listitem>
-    <para>
-     <package>PackageKit</package>
     </para>
    </listitem>
    <listitem>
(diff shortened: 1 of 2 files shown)


cjschroder:
A nice PDF for reviewers:
[cha.transactional-updates_color_en.pdf](https://github.com/SUSE/doc-sle/files/3262958/cha.transactional-updates_color_en.pdf)

xml/transactional_updates.xml
```diff
@@ -0,0 +1,517 @@
...
+   <package>transactional-update</package> package like any other package with
+   Zypper or &yast;.
+  </para>
+
+  <para>
+   You may also elect to install
+   <package>transactional-update-zypp-config</package>, which prevents Zypper
+   from accidentally removing <package>transaction-update</package>.
+  </para>
+ </sect1>
```
thkukuk: The whole paragraph is wrong. I have no clue why transactional-update is in Package Hub, I did never approve that. It will not work.
You have to make the decision during installation if you want to use transactional updates or not. If you want to use it, you have to:
1. Add the "Transactional Server" Module during registration/isntallation of the system.
2. Select the Transactional Server System Role

Installing any package from the Transactional Server Module later in a running system, is NOT supported and can break the system.
Changing the Subvolume-Layout of the root partition or make sub-directories/subvolumes of the root partition an own partition (except /home, /var, /srv, /opt) is not supported and will most likely break the system.

xml/transactional_updates.xml
```diff
@@ -0,0 +1,517 @@
...
+ <sect1>
+  <title>The <command>transactional-update</command> Command</title>
+
+  <important>
+   <title>Only use when requested by SUSE Support</title>
+   <para>
+    This reference for <command>transactional-update</command> should only be
+    used when requested by SUSE Support.
+   </para>
+  </important>
```
thkukuk: No clue where this is coming from, but the customer has to use transactional-update of course for daily work on the system. It's the only way for the customer to install/remove/update packages.

xml/transactional_updates.xml
```diff
@@ -0,0 +1,517 @@
...
+   update fails.
+  </para>
+
+  <para>
+   All changes become active only after reboot.
+  </para>
+
+  <para>
+   If the current root file system is identical to the active root file system
+   (after applying updates and reboot), remove all old snapshots:
```
thkukuk: The snapshots will not be removed! This command will tell snapper, that this snapshots can be deleted on the next cleanup phase. From the documentation:

           If the current root filesystem is identical to the active root
           filesystem (means after a reboot, before transactional-update
           creates a new snapshot with updates), all old snapshots without a
           cleanup algorithm get a cleanup algorithm set. This is to make
           sure, that old snapshots will be deleted by snapper. See the
           section about cleanup algorithms in snapper(8).

           Also removes all unreferenced (and thus unused) /etc overlay
           directories in /var/lib/overlay.

xml/transactional_updates.xml
```diff
@@ -0,0 +1,517 @@
...
+<screen>&prompt.root;<command>transactional-update --help</command></screen>
+    </listitem>
+   </varlistentry>
+  </variablelist>
+ </sect1>
+ <sect1 xml:id="sec.tu-disable">
+  <title>Disabling Transactional Updates</title>
+
+  <para>
+   You can disable transactional updates with this command:
```
thkukuk: I would rename the section to: Automatic Updates

and explain, that a systemd-timer will run once a day, applies all updates, and informs the rebootmgrd that the machine should be rebooted.
To diable automatic updates, run "systemctl ...". The time, when the update runs, can be adjusted as needed, see systemd-timer manual page. The maintenance window, in which rebootmgrd will reboot the machine, is also configureable, see rebootmgrd manual page.

xml/transactional_updates.xml
```diff
@@ -0,0 +1,517 @@
...
+<screen>&prompt.root;<command>systemctl --now disable transactional-update.timer</command></screen>
+
+  <note>
+   <title>Disabling Transaction Update Timer is Required During Upgrade</title>
+   <para>
+    You must disable transactional updates during the upgrade procedure from
+    one version of &productname; to the next.
+   </para>
+  </note>
+ </sect1>
```
thkukuk: Looks like this was copied from CaaSP. It is not true for SLES.

xml/transactional_updates.xml
```diff
@@ -0,0 +1,517 @@
...
+ <sect1 xml:id="sec.tu-troubleshooting">
+  <title>Troubleshooting</title>
+
+  <para>
+   If the upgrade fails, run <command>supportconfig</command> to collect log
+   data. Provide the resulting files, including
+   <filename>transactional-update-migration.log</filename> to &suse; Support.
```
thkukuk: Never heard of this log file. It's /var/log/transactional-update.log

xml/transactional_updates.xml
```diff
@@ -0,0 +1,517 @@
...
+   <dm:bugtracker></dm:bugtracker>
+   <dm:translation>yes</dm:translation>
+  </dm:docmanager>
+  <abstract>
+   <para>
+    Transactional updates are available in &productname; as a technology
+    preview. Transactional updates provide another mechanism for keeping your
+    system updated. Transactional updates are atomic (all updates are applied
+    only if all updates succeed), support rollbacks, and update both read-write
+    and read-only filesystems. It does not affect a running system as no
```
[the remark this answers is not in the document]
cjschroder: Thanks @thkukuk, this is a great review. A lot of the material came from the CaaSP manual. I appreciate your attention to detail & fast response.

xml/transactional_updates.xml
cjschroder: I found this during testing on my SLES 15 test system.

cjschroder: Which is a normal SLES 15 installation, not a read-only system

cjschroder:
Many thanks @thkukuk, this is extremely helpful. I will leave this open for two more days in case there are any more comments or corrections.
[cha.transactional-updates_color_en.pdf](https://github.com/SUSE/doc-sle/files/3273185/cha.transactional-updates_color_en.pdf)

xml/transactional_updates.xml
```diff
@@ -0,0 +1,444 @@
...
+   <dm:bugtracker></dm:bugtracker>
+   <dm:translation>yes</dm:translation>
+  </dm:docmanager>
+  <abstract>
+   <para>
+    Transactional updates are available in &productname; as a technology
+    preview, for updating &slsa; when the root filesystem is read-only. 
+    Transactional updates are atomic (all updates are applied only if all 
+    updates succeed) and support rollbacks. It does not affect a running system 
+    as no changes are applied until after the system is rebooted.
```
thkukuk: ...are applied until...
They are applied, they are only not visible and get only activated with the next system reboot.

xml/transactional_updates.xml
```diff
@@ -0,0 +1,444 @@
...
+ <title>Transactional Updates</title>
+ <info>
+  <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
+   <dm:bugtracker></dm:bugtracker>
+   <dm:translation>yes</dm:translation>
+  </dm:docmanager>
+  <abstract>
+   <para>
+    Transactional updates are available in &productname; as a technology
+    preview, for updating &slsa; when the root filesystem is read-only.
```
thkukuk: In general: transactional update should be used, if disturbing a running service is more expensive than a reboot. If a reboot is very expensive, transactional updates should not be used.

xml/transactional_updates.xml
```diff
@@ -0,0 +1,444 @@
...
+    as no changes are applied until after the system is rebooted.
+   </para>
+
+   <para>
+    Transactional updates are run daily by the
+    <command>transactional-update</command> script. The script checks for
+    available updates. If there are any updates, it creates a new snapshot of
+    the root file system in the background, and then fetches updates from the
+    release channels. After the new snapshot is completely updated, it is
+    marked as active and will be the new default after the next reboot of the
```
thkukuk: ...and will be the new default root filesystem after the next reboot...

xml/transactional_updates.xml
```diff
@@ -0,0 +1,444 @@
...
+  </abstract>
+ </info>
+ 
+ <sect1 xml:id="sec.tu-limitations">
+  <title>Limitations of Technical Preview</title>
+
+  <para>
+   As a technical preview, there are certain limitations in functionality. The
+   following packages will not work with <command>transactional-update</command>. 
+   In the <filename>/srv</filename> directory:
```
thkukuk: I would remove the /srv and /var directory separation and move the two sections together. This was for developers, that they know where the problem with their package is, but will only confuse customers.

xml/transactional_updates.xml
```diff
@@ -0,0 +1,444 @@
...
+  </para>
+  <itemizedlist>
+   <listitem>
+    <para>
+     <package>mpi-selector</package>
+    </para>
+   </listitem>
+   <listitem>
+    <para>
+     <package>emacs</package> games (Emacs still works)
```
thkukuk: Too cryptic, nobody will understand this. We should write, that emacs works except for the emacs games or so.

cjschroder:
> Some few more things to change.

Many thanks for your help!

cjschroder:
maintenance/SLE15SP0
e95660c886bdd

PR #450, opened 2019-07-12
Feature/docserv2 dc adjustments

taroth21:
This PR has two goals:

- Making sure the profiling info is the same in all DCs files for which the HTML is build from DC-SLE*-all.  

- Cleaning up the DC files (removing parameters that are no longer relevant, e.g. HTMLROOT).


sknorr:
Merci beaucoup!

taroth21:
BTW, for SLES, the profiling info was the same across all DCs, no changes required.

PR #459, opened 2019-07-31
Kill xpointers in Installation Chapters

taroth21:
### Description
The xpointers introduced by Sven in the installation chapters for the x86 and the aarch64 architecture were aimed at de-duplicating content. However, they turned out to be painful whenever adjustments to these chapters are needed. Therefore resolved the xpointers and re-added the content to the individual XML files, adding XML comments where the same content is used in different files. Also removed deplyoment_xpointers.xml  

### Checklist

*Are backports required?*

- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP4
- [x] To maintenance/SLE15SP0


taroth21:
Build a new PDF of the Deployment Guide before each commit and double-checked with 'pdfcompare' that the content did not differ from the original PDF. So should be safe to merge without review (as soon as the Travis run is finished).

PR #460, opened 2019-08-01
Update adm_sudo.xml

JeGoi:
### Description
Update exampleuser_plain for 'Starting a Shell' with -s as it is in para

### Checklist
* Check all items that apply.

*Are backports required?*
Hum, I don't think so


JeGoi:
No worries, it was a microscopic modification.
This doc is a very good! Thanks a lot.

PR #462, opened 2019-08-05
Add advice re storage space for RMT, note location of packages storage

cjschroder:
### Description
Add storage space recommendations and location of mirrored packages

### Checklist
* Check all items that apply.

*Are backports required?*

- [ x] To maintenance/SLE15SP1
- [ x] To maintenance/SLE15SP0

### DIFF
--- a/xml/rmt_install.xml
+++ b/xml/rmt_install.xml
@@ -5,10 +5,12 @@
     %entities;
 ]>
 
-<chapter xmlns="http://docbook.org/ns/docbook" xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:xlink="http://www.w3.org/1999/xlink" version="5.0" xml:id="cha-rmt-installation">
+<chapter xmlns="http://docbook.org/ns/docbook" 
+         xmlns:xi="http://www.w3.org/2001/XInclude" 
+         xmlns:xlink="http://www.w3.org/1999/xlink" 
+         version="5.0" xml:id="cha-rmt-installation">
 <!-- ==================================================================== -->
  <title>&rmt; Installation and Configuration</title>
-
  <para>
   &rmt; is included in &sls; starting with version 15. Install &rmt;
   directly during the installation of &productname; or install it on a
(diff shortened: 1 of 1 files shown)


xml/rmt_install.xml
```diff
@@ -31,7 +33,232 @@
...
+         The amount of storage your &rmt; server requires is dependent on several variables: the 
+         number of  repositories and architectures that you mirror, how many packages are 
+         downloaded, the  number of  extensions that are enabled, and the age of your server installation. 
+         Outdated  packages are not removed by &rmt;, but accumulate over time. When new versions are 
+         downloaded, the old versions are kept on disk but are not referenced by repository 
+         metadata, which makes them invisible to clients.
+     </para>
+     <para>
+         As a general guide, two to three times the total size of all mirrored repositories should be 
+         sufficient. If you are mirroring a single release, such as &slea; 15.1, 250 GB provides room
```
ikapelyukhin: `a single release` -- that seems confusing. It's not just SLES 15.1, it's SLES15.1 + all extensions (incl. PackageHub).

Also the figures seem rather arbitrary -- in principle there shouldn't be much difference between 15.0 and 15.1, in fact 15.1 repos are smaller. And yet the suggested figure for 15.1 is 120 GB higher than for 15.0.

`As a guidline, 200 GB is sufficient to mirror all repos for any SLES release with all extensions`?

xml/rmt_install.xml
```diff
@@ -31,7 +33,232 @@
...
+         for packages and updates. The total size of all x86_64 repositories for &slea; 12 SP4 is about 130 GB, and about   
+         155 GB for &slea; 15. There is not (yet) a tool for reporting repository sizes, so it is a good practice to monitor disk
```
ikapelyukhin: Reporting repo sizes -- I'm uncertain if anything like this will be ever officially supported. So again, not sure if it is worth mentioning.

xml/rmt_install.xml
```diff
@@ -31,7 +33,232 @@
...
+        <thead>
+     <row>
+      <entry>
+       <para>
+        Product
+       </para>
+      </entry>
+      <entry>
+       <para>
+        Size of default repositories in GB
```
ikapelyukhin: They are called `mandatory` rather than `default` elsewhere in the docs: https://github.com/SUSE/doc-sle/blob/30ba9ed801005014c5fe2bfa9340f89fc0cd69e5/xml/rmt_rmt-cli.xml#L125

xml/rmt_install.xml
```diff
@@ -31,7 +33,232 @@
...
+       <para>
+       4.79
+       </para>
+      </entry>
+     </row>   
+     
+            <row>
+      <entry>
+       <para>
+      sle-module-web-scripting/15/x86_64
```
ikapelyukhin: `sle-module-web-scripting` -- friendly names instead of identifiers? The name of the extension is `Web and Scripting Module 15 x86_64` -- that's what is displayed in SCC.

Same applies for all other products.

cjschroder:
That is great feedback. Now it is even more simplified and useful. Thank you!

PR #478, opened 2019-09-24
add interface attributes

jreidinger:
### Description
Goal of this pull request is to improve documentation for autoyast attributes for network interfaces.

### Checklist
* Check all items that apply.

*Are backports required?*

well, it is improvement, not bug fix, so I am not sure. Also some notes about deprecation may not apply.
- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

@teclator @imobachgs can you check it also?


xml/ay_bigfile.xml
```diff
@@ -6662,7 +6650,441 @@ config:type="boolean"&gt;true&lt;/keep_install_network&gt;</screen>
...
+       </row>
+       <row>
+        <entry>
+         <para>
+          bridge_ports
+         </para>
+        </entry>
+        <entry>
+         <para>
+           Comma separated list of bridge ports e.g. <literal>eth0 eth1</literal>.
```
taroth21: Hm, there's something fishy here, because both here and in the example above, the list is separated by spaces. So either the description or the examples need to be fixed.

imobachgs:
@taroth21 @jreidinger I've opened a PR against Josef's branch adding some content and including some format fixes. See https://github.com/jreidinger/doc-sle/pull/1.

xml/ay_bigfile.xml
jreidinger: you are right. It is space separated list. My fault.

jreidinger:
@imobachgs pr is now incorporated and also mistake with separator is fixed.

taroth21:
@jreidinger @imobachgs : Thanks! Sorry I could not make it today, will review first thing tomorrow morning and take care of merging. BTW, should this only go into (future) SLE 15 SP2? (just double-checking because there were no backports requested as per PR template above).

imobachgs:
@taroth21 Good point. I would say that we could backport this documentation to, at least, SLE 15 SP1. We should check whether some deprecation notice is still valid for SP1 (I think yes), but most of the documentation should be valid. @jreidinger What's your point of view?

taroth21:
@jreidinger @imobachgs : Thanks a lot for your work! :) I have opened https://github.com/SUSE/doc-sle/pull/480 as a follow-up on this one. The new PR contains your original commits plus some minor style fixes I applied. Will close this one here unmerged.

xml/ay_bigfile.xml
taroth21: No problem :).

jreidinger:
@imobachgs yes, I think that majority of description is valid. Of course deprecation, some notes about smart udev and also maybe some ignored attributes is not so ignored. Also I am not sure if driver options is not done in udev in old branch.

I do not check backport because this is not bugfix ( aka bug in documentation, just insufficient documentation ).

PR #480, opened 2019-09-27
autoyast attributes for network interfaces

taroth21:
### Description
Follow-up to https://github.com/SUSE/doc-sle/pull/478 after applying some minor style fixes. The goal of this pull request is to improve documentation for autoyast attributes for network interfaces.

### Checklist
* Check all items that apply.

*Are backports required?*

- [x] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3


taroth21:
@imobachgs @jreidinger : Please have a quick look if everything is OK and let me know if I can cherrypick to SLE 15 SP1. Based on Imobachs comment in https://github.com/SUSE/doc-sle/pull/478#issuecomment-535881133, most of the content changes should also apply to SLE 15 SP1. You could also comment here on the options that *do not* apply to older versions, then I would merge only the applicable changes manually, skipping the ones that are for SP2 only.

jreidinger:
@taroth21 your changes looks good to me. For backport, we probably need to retest it and recheck all elements, as it is possible that some elements are required in old code and no longer mandatory in new one. Sadly code is not so clear in old code as it goes through two layers and many helpers that can modify it.

PR #515, opened 2019-11-14
docker: Add clarifications about image building when connected to rmt

rhafer:
### Description

There is slight difference in the functionality of container-suseconnect on systems that are using RMT/SMT instead of SCC. Add a note about that to the containers guide

For details see also: SUSE/container-suseconnect#37

I am not entirely sure if I got the markup right so feel free to correct that if needed.

### Checklist
* Check all items that apply.

*Are backports required?*

- [x] To maintenance/SLE15SP1
- [x] To maintenance/SLE15SP0
- [x] To maintenance/SLE12SP5
- [x] To maintenance/SLE12SP4
- [x] To maintenance/SLE12SP3

### DIFF
--- a/xml/docker_building_images.xml
+++ b/xml/docker_building_images.xml
@@ -150,6 +150,25 @@
    </para>
   </note>
 
+  <note>
+   <title>Building images on systems registered with &rmt;</title>
+
+   <para>
+    When the host system used for building the docker image is registered
+    against &rmt; it is only possible to build containers for the same &slea; code
+    base that the host system is running on. I.e. if you docker host is a
+    &slea;&nbsp; 15 system you can only build &slea;&nbsp; 15 based images
+    on that host by default.
+
+    To build images for a different &slea; version than what is running on
+    the host machine credentials for the target release can be injected into
+    the container as outlined below.
+
+    When the host system is registered again the &scc; this
+    restriction does not apply.
+   </para>
+  </note>
+
   <para>
    To obtain the list of repositories, use the following command:
   </para>
--- a/xml/docker_building_images.xml
+++ b/xml/docker_building_images.xml
@@ -202,15 +202,11 @@ RUN zypper -n in vim</screen>
    <para>
     When the &docker; host machine is registered against an internal &rmt;
     server, the &docker; image requires the SSL certificate used by &rmt;:
-   <remark>
-    taroth 2018-08-08: not sure the content of this file is still correct, as
-    SMT has been replaced by RMT (Repository Mirroring Tool) meanwhile - @DEV:
-    please check</remark>
    </para>
 <screen>FROM registry.suse.com/suse/sles12sp4
 
-# Import the crt file of our private SMT server
-ADD http://smt.test.lan/smt.crt /etc/pki/trust/anchors/smt.crt
+# Import the crt file of our private RMT server
+ADD http://rmt.test.lan/rmt.crt /etc/pki/trust/anchors/rmt.crt
 RUN update-ca-certificates
 
 RUN zypper ref -s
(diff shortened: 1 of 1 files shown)


xml/docker_building_images.xml
```diff
@@ -150,6 +150,25 @@
...
 
+  <note>
+   <title>Building images on systems registered with &rmt;</title>
+
+   <para>
+    When the host system used for building the docker image is registered
+    against &rmt; it is only possible to build containers for the same &slea; code
+    base that the host system is running on. I.e. if you docker host is a
+    &slea;&nbsp; 15 system you can only build &slea;&nbsp; 15 based images
+    on that host by default.
```
ikapelyukhin: Generally speaking, that shouldn't be the case.

rhafer: @ikapelyukhin What exactly shouldn't be the case?

You mean it should be possible to build SLE12 containers on a SLE15 host when that host is registered against RMT? If yes, how so. AFAICS the RMT APIs are not providing the information to make that possible.

ikapelyukhin: @rhafer well, what information do you need? :slightly_smiling_face:

rhafer: @ikapelyukhin In order to fetch the repos for a given product `container-suseconnect` uses the `/connect/subscriptions/products` API for that to work it needs the regcode. The tool currently queries that via `/connect/systems/subscriptions` (using the credentials from the SCCcredentials file on the host). This all works fine when connecting to SCC directly.

However when connecting to RMT (or SMT) the `/connect/systems/subscriptions` API doesn't exist and  `/connect/subscriptions/products` will give a 401 (unauthorized) as it needs a regcode. In that case `container-suseconnect` will fallback (at least when using https://github.com/SUSE/container-suseconnect/pull/38)  to `/connect/systems/products` which just returns the repos for the product(s) that the host's credentials have been registered for.

So if you know a way to workaround that I am happy to learn about that.

ikapelyukhin: @rhafer you can do the following experiment: see in how many places in SUSEConnect's code base `subscriptions/products` is being accessed.

It isn't used by SUSEConnect at all. I believe `subscriptions/products` endpoint is only used by SUMA to do subscription matching. You should never use `subscriptions/products` (probably shouldn't use `systems/subscriptions` either) -- always use `systems/products` instead.

On public cloud RMTs you are likely to run into other authentication issues, but outside of the public cloud RMT's authentication mechanisms are much less restrictive than that of SCC.

rhafer: > It isn't used by SUSEConnect at all. I believe `subscriptions/products` endpoint is only used by SUMA to do subscription matching. You should never use `subscriptions/products` (probably shouldn't use `systems/subscriptions` either) -- always use `systems/products` instead.

@ikapelyukhin Note, that the original code in container-suseconnect isn't from me, but I am pretty sure that it's implement the way it is intentionally.  E.g. to specifically allow building e.g. SLE12 images on SLE15 hosts.

Moving to `systems/products` in all cases would "break" what I think is quite a handy feature. Even when using SCC.

What's wrong with the current approach in https://github.com/SUSE/container-suseconnect/pull/38? It actually works for SCC and RMT (with the restriction mention in here)

Maybe @flavio can shed some light on this as he wrote much of the original code I believe.

(Somehow I think this discussion should have move to https://github.com/SUSE/container-suseconnect/pull/38 a couple of comments ago, but it's to late for that now, ;-) )

cjschroder:
hello @rhafer, thank you for the PR! We :heart: contributors. You're welcome to discuss here :) You should add [WIP] to the title until you resolve your question.

xml/docker_building_images.xml
ikapelyukhin: @rhafer ah, OK -- `container-suseconnect` implements a zypper plugin and `systems/subscriptions` is used to get repo access tokens from SCC. Fair enough, this works for now. I've sent out an email with further details about this issue in general, see if for further details about the build process in public clouds.

rhafer:
> hello @rhafer, thank you for the PR! We heart contributors. You're welcome to discuss here :) I have marked this [WIP] until you resolve your question.

@cjschroder I think we've clarified the remaining issues now (via mail). And the change should be good to go in as is for now. So I'll remove the WIP marker again.

As @ikapelyukhin pointed out will we need to do some further adjustments on `container-suseconnect` in the near future. I'll submit separate documentation PRs for those changes as needed.

cjschroder:
SLE15SP1: 25f785d9174f2223
SLE15SP0: 917dd60fafe8669f

SMT is SLE 12 only, and RMT is SLE 15 only, see https://documentation.suse.com/sles/15-SP1/html/SLES-all/cha-rmt-migrate.html#sec-rmt-migrate-notes-features

PR #528, opened 2019-12-10
12 SP5: Change product from SLED to SLE-WE

sknorr:
### Description

This is a minimal fix, not quite sure if this will lead to issues in some profiled parts of the documents. Looks ok at first glance though.

Also no clue if this should be applied to the translated documents and if so, whether that is easy to apply. What do you think?

### Checklist

*Are backports required?*

No.


taroth21:
We discussed this for a while but there are more implications than we originally thought, therefore rejecting.

PR #531, opened 2019-12-17
Feature/adm sudo revision

dmpop:
### Description

A complete revision of the sudo chapter.

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

### DIFF
--- a/xml/adm_sudo.xml
+++ b/xml/adm_sudo.xml
@@ -607,43 +607,50 @@ users
   </sect2>
 
   <sect2 xml:id="sec-sudo-usecases-xorg">
-   <title>Using &sudo; with &xvendor; Applications</title>
+   <title>Running Graphical Applications with &sudo;</title>
    <para>
-    When starting graphical applications with &sudo;, you will encounter the
-    following error:
+    By default, unprivileged users are not allowed to run graphical
+    applications using &sudo;. Doing this usually results in an error message,
+    for example:
    </para>
 <screen>
 &prompt.sudo;xterm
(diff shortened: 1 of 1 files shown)


xml/adm_sudo.xml
```diff
@@ -555,7 +539,7 @@ User_List Host_List = [(User_List)] [NOPASSWD:|PASSWD:] Cmnd_List
      </para>
      <para>
       Depending on whether you want users to re-enter their passwords,
-      uncomment the specific line in <filename>/etc/sudoers</filename> and
+      uncomment the appropriate line in <filename>/etc/sudoers</filename> and
       comment out the default rule.
      </para>
```
tomschr: I'm not sure if you should change `/etc/sudoers` directly. In a way, it contradicts your own recommendation. :wink: 
Couldn't you add these two comments into the newly created `/etc/sudoers.d/userpw` file?
I think, that would avoid to edit the main file and you can see it as a holistic solution.

tomschr:
Sorry to bother again, but I think, I have another idea which might be helpful for this chapter.

For the "sudo Use Case" section, you could add one or more additional subsections:

* Giving access to a specific command that needs root privileges (`Cmd_Alias`).
* Add specific users to an "admin" group that are trusted and allowed root privileges (`User_Alias`)

PR #532, opened 2019-12-17
Add warning to partitioning § describing GB/GiB difference (BSC#1155315)

lproven:
### Description
This adds a Warning to the partitioning section of the Quick Install guide with a note about binary/decimal units, as per https://bugzilla.suse.com/show_bug.cgi?id=1155315 

If the working/formatting looks OK then I will insert the same warning in the various other partitioning section.
### Checklist
* Check all items that apply.

*Are backports required?*

- [x] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [x] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

### DIFF
--- a/xml/art_sle_install_quick.xml
+++ b/xml/art_sle_install_quick.xml
@@ -570,6 +570,45 @@
        </listitem>
       </varlistentry>
      </variablelist>
+     <warning>
+      <title>Disk Space Units</title>
+      <para>
+       Note that for partitioning purposes, disk space is measured in binary
+       units, rather than in decimal units. For example, if you enter sizes of
+       <literal>1GB</literal>, <literal>1GiB</literal> or <literal>1G</literal>,
+       all signify 1&nbsp;GiB (Gibibyte), as opposed to 1&nbsp;GB (Gigabyte).
+      </para>
+      <variablelist>
+       <varlistentry>
+        <term>Binary</term>
+        <listitem>
+         <para>
+          1&nbsp;GiB = 1&nbsp;073&nbsp;741&nbsp;824 bytes.
+         </para>
+        </listitem>
+       </varlistentry>
+       <varlistentry>
+        <term>
+         Decimal
+        </term>
+        <listitem>
+         <para>
+          1&nbsp;GB = 1&nbsp;000&nbsp;000&nbsp;000 bytes.
+         </para>
+        </listitem>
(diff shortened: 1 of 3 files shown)


xml/art_sle_install_quick.xml
```diff
@@ -570,6 +570,45 @@
        </listitem>
       </varlistentry>
      </variablelist>
+     <warning>
+      <title>Disk Space Units</title>
+      <para>
+       Note that for partitioning purposes, disk space is measured in binary
+       units, rather than in decimal units. For example, if you enter sizes of
+       <literal>1GB</literal>, <literal>1GiB</literal> or <literal>1G</literal>,
+       all signify 1&nbsp;GiB (Gibibyte), as opposed to 1&nbsp;GB (Gigabyte).
```
dmpop:
```suggestion
       they all signify 1&nbsp;GiB (Gibibyte), as opposed to 1&nbsp;GB (Gigabyte).
```

xml/deployment_expert_partitioner.xml
```diff
@@ -29,6 +29,45 @@
...
+   units, rather than in decimal units. For example, if you enter sizes of
+   <literal>1GB</literal>, <literal>1GiB</literal> or <literal>1G</literal>,
+   they all signify 1&nbsp;GiB (Gibibyte), as opposed to 1&nbsp;GB (Gigabyte).
+  </para>
+  <variablelist>
+   <varlistentry>
+    <term>Binary</term>
+    <listitem>
+     <para>
+      1&nbsp;GiB = 1&nbsp;073&nbsp;741&nbsp;824 bytes.
```
cjschroder: Good stuff, this is one of those pitfalls that trap the unwary.

Why use spaces instead of commas?

lproven: Good question. This is a standard "Brit abroad" precautionary measure. 

In non-English-speaking Europe, notably including Germany & Czechia, they use commas for the decimal marker, and sometimes, periods as the three-digit separator. So if I put in an expenses claim for €12.34 it will say €12,34 etc.  This confuses English speakers but our way confuses them, so it's common to replace the 3-digit separator with a non-breaking space.

I am not aware of house style but I should be -- I will find out.

PR #545, opened 2020-01-20
YaST-related parts for 389-ds ldap server

taroth21:
### Description
Working with @Firstyear to finish the YaST parts of the 389 ds server documentation.  

### Checklist

*Are backports required?*

- [x] To maintenance/SLE15SP1
- [x] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3


xml/security_ldap.xml
```diff
@@ -784,7 +784,95 @@ systemctl start sssd</screen>
...
+   </informalfigure>
+  </step>
+  <step>
+   <para>
+    If all setting are according to your wishes, confirm with <guimenu>OK</guimenu>.
+   </para>
+  </step>
+ </procedure>
+ <para>
+  <remark>taroth 2020-02-20: @firstyear: does another window follow if I click 'OK'?
```
Firstyear: There are no following windows, it just installs

xml/security_ldap.xml
```diff
@@ -784,7 +784,95 @@ systemctl start sssd</screen>
...
+   I could not test it  as I do not have test certificates (and one cannot click
+   'OK' without filling all fields, which makes sense)</remark>
```
Firstyear: That must be the old version, the new version doesn't require certificates ....

xml/security_ldap.xml
```diff
@@ -784,7 +784,95 @@ systemctl start sssd</screen>
...
+  &yast; creates a directory server instance with the settings you have specified.
+  <remark>taroth 2020-01-20: @firstyear: what happens in the background?
```
Firstyear: Uhhhhh .... some things can be changed, some can't? I think instance name can never be changed post install. Suffix can be changed, but it's a big effort. Everything else is easy to change.

xml/security_ldap.xml
taroth21: Yes, you are right - I tried on SLE 15 SP1 with the latest updates for 389-ds installed. I installed the latest snapshot for SLE 15 SP2 today and could finish the setup in YaST without certificates.

xml/security_ldap.xml
taroth21: Changed that, thanks.

PR #546, opened 2020-01-21
adapted keyboard valid entries

schubi2:
### Description
The regarding keymap code entry is also a valid entry.

### Checklist
* Check all items that apply.

*Are backports required?*
No, head only


cjschroder:
Thank you, @schubi2! Merging.

PR #579, opened 2020-04-13
Update links to documentation.suse.com

taroth21:
Based on the following guidelines:

*  avoid entities for URLs (entities for parts of the URL make it harder to
    copy/paste/replace complete URLs, plus they will be resolved during
    translation anyway)
-> this way the URLs are the same in the English versions and translation
    branches and can replaced the same way if needed
 *  using upcoming 15-SP2 URLs 
    (they will only work after FCS, but it prevents us from updating them
     in the translation branches after FCS)
 * use links to html instead of single-html (shorter download times, SEO)
 * use descriptive text for links to mention book name
    (in case deeplink is broken, plus better accessibility)
based on the following guidelines:

*  avoid entities for URLs (entities for parts of the URL make it harder to
    copy/paste/replace complete URLs, plus they will be resolved during
    translation anyway)
-> this way the URLs are the same in the English versions and translation
    branches and can replaced the same way if needed
 *  using upcoming 15-SP2 URLs 
    (they will only work after FCS, but it prevents us from updating them
     in the translation branches after FCS)
 * use links to html instead of single-html (shorter download times, SEO)
 * use descriptive text for links to mention book name
    (in case deeplink is broken, plus better accessibility)

### DIFF
--- a/xml/adm_support.xml
+++ b/xml/adm_support.xml
@@ -1631,7 +1631,7 @@ setup-sca</screen>
       <step>
        <para>
         Check the &suse; Knowledge base
-        (<link xlink:href="http://www.suse.com/support/kb/"/>) for results that
+        (<link xlink:href="https://www.suse.com/support/kb/"/>) for results that
         directly relate to the problem identified by SCA. Work at resolving
         them.
        </para>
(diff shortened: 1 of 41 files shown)


taroth21:
The files ```art_sle_install_quick.xml``` and ```deployment_yast_installer.xml``` had the same list of links. To avoid that the links interrupt the sentence flow (they had been insertions before), I moved them to the end of the sentence for better readability and translatablity.

sknorr:
> avoid entities for URLs (entities for parts of the URL make it harder to
> copy/paste/replace complete URLs, plus they will be resolved during
> translation anyway)
> -> this way the URLs are the same in the English versions and translation
> branches and can replaced the same way if needed

To be honest, the translation argument makes no sense to me -- the English documents in the `trans/` branches includes entities just like the normal version. The translated documents have a ton of translated context and don't match line numbers, so updating URLs in these documents would be a bigger upheaval in any case.

fsundermeyer:
Links that have part entity and part link are really hard to read and are error prone (forgetting slashes or having double slashes, etc. and therefore Tanja and me agreed to drop them.

PR #580, opened 2020-04-14
Update the Gnome fix

dmpop:
### Description
Replaced outdated info with a new fix for broken Gnome.

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

### DIFF
--- a/xml/troubleshooting.xml
+++ b/xml/troubleshooting.xml
@@ -1079,64 +1079,25 @@
   </sect2>
 
   <sect2 xml:id="sec-trouble-login-dk">
-   <title>Login Successful but &gnome; Desktop Fails</title>
+   <title>&gnome; Desktop Has Issues</title>
    <para>
-    If this is the case, it is likely that your &gnome; configuration files
-    have become corrupted. Some symptoms may include the keyboard failing to
-    work, the screen geometry becoming distorted, or even the screen coming up
-    as a bare gray field. The important distinction is that if another user
-    logs in, the machine works normally. It is then likely that the problem can
-    be fixed relatively quickly by simply moving the user's &gnome;
-    configuration directory to a new location, which causes &gnome; to
-    initialize a new one. Although the user is forced to reconfigure &gnome;,
-    no data is lost.
+     If you are experiencing issues with the &gnome; desktop (for example,
+     incorrect screen geometry, blank desktop, etc.), try resetting &gnome; settings.
+     Before you do this, back up the existing configuration files using the following command:
    </para>
-   <procedure xml:id="proc-trouble-login-dk">
-    <step>
-     <para>
-      Switch to a text console by pressing <keycombo>
-      <keycap function="control"/> <keycap function="alt"/> <keycap>F1</keycap>
-      </keycombo>.
-     </para>
-    </step>
-    <step>
-     <para>
-      Log in with your user name.
-     </para>
-    </step>
-    <step>
-     <para>
-      Move the user's &gnome; configuration directories to a temporary
-      location:
-     </para>
-<screen>&prompt.user;mv .gconf  .gconf-ORIG-RECOVER
-&prompt.user;mv .gnome2 .gnome2-ORIG-RECOVER</screen>
-    </step>
-    <step>
-     <para>
-      Log out.
-     </para>
-    </step>
-    <step>
-     <para>
-      Log in again, but do not run any applications.
-     </para>
-    </step>
-    <step>
-     <para>
-      Recover your individual application configuration data (including the
-      Evolution e-mail client data) by copying the
-      <filename>~/.gconf-ORIG-RECOVER/apps/</filename> directory back into the
-      new <filename>~/.gconf </filename> directory as follows:
-     </para>
-<screen>&prompt.user;cp -a .gconf-ORIG-RECOVER/apps .gconf/</screen>
-     <para>
-      If this causes the login problems, attempt to recover only the critical
-      application data and reconfigure the remainder of the applications.
-     </para>
-    </step>
-   </procedure>
+<screen>mkdir ./old-gnome-config/ &amp;&amp; mv ./.gnome* ./old-gnome-config/ &amp;&amp; mv .gconf* ./old-gnome-config/ &amp;&amp; mv ./.cache ./old-gnome-config/ &amp;&amp; mv ./dbus ./old-gnome-config/ &amp;&amp; mv ./.dmrc ./old-gnome-config/ &amp;&amp; mv ~/.config/dconf/* ./old-gnome-config/</screen>
+   <para>
+     Run the <command>dconf reset -f /org/gnome/</command>, then log out and try to log in again.
+   </para>
+   <para>
+     If this fixed the desktop issues, restore the desired configuration files.
+   </para>
+   <para>
+     In case you still have problems with the desktop, try the following command:
+   </para>
+<screen>mv .config/dconf/user .config/dconf/user.bak &amp;&amp; sudo reboot</screen>
   </sect2>
+
  </sect1>
  <sect1 xml:id="sec-trouble-netconfig">
   <title>Network Problems</title>


xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,25 @@
...
-      new <filename>~/.gconf </filename> directory as follows:
-     </para>
-<screen>&prompt.user;cp -a .gconf-ORIG-RECOVER/apps .gconf/</screen>
-     <para>
-      If this causes the login problems, attempt to recover only the critical
-      application data and reconfigure the remainder of the applications.
-     </para>
-    </step>
-   </procedure>
+<screen>mkdir ./old-gnome-config/ &amp;&amp; mv ./.gnome* ./old-gnome-config/ &amp;&amp; mv .gconf* ./old-gnome-config/ &amp;&amp; mv ./.cache ./old-gnome-config/ &amp;&amp; mv ./dbus ./old-gnome-config/ &amp;&amp; mv ./.dmrc ./old-gnome-config/ &amp;&amp; mv ~/.config/dconf/* ./old-gnome-config/</screen>
```
sknorr, line 1088: 1. The command is really long and hard to read. Please check such things in the (PDF) output.
2. You can combine multiple moves like this, with the output dir being the last argument: `mv ./.gnome* .gconf* ./old-gnome-config/`. (However, if any of those dirs does not exist, I think the entire command will fail. So, doing individual moves may ultimately be safer.)
3. Look at the consistency of your command, you use `./` sometimes in front of the dir name, sometimes you don't.
4. I would suggest making sure somehow that the user is in `~`. Either prefix all the dir names in your command or add a (slightly redundant looking) `cd ~` at the beginning.
5. I am a bit at issue with the dirs/files that are being moved:
   * `.dmrc` -> I have _never_ seen that file on any machine afaicr. I'd suggest dropping it from the list of things.
   * `.gnome*` (aka `.gnome`, `.gnome2`, `.gnome2_private`) -> these may still used by some old GTK2 _applications_, they won't help with any desktop issues. (Anyone who has actual content in these dirs has most likely been migrating their user dir for 10+ years without ever cleaning up. For me, these are just empty dirs.)
   * `.gconf*` -> This seems like it's the old dir name, GConf config has moved into `.config/gconf` now. Not sure if any configuration in the old dir name is still used if it's present but I almost can't imagine it.
   *  `.cache` -> This one, as it's name implies is only cached data: web cache, file browser thumbnails, music cover art. It tends to be a very large directory full of things that can be easily generated or downloaded again. I'd vouch for its complete removal.
   * `./dbus` -> Afaics, this dir only contains a reference to the current session's DBus. Can probably be removed entirely, but not 100% sure here.
   * `~/.config/dconf/*` -> Moving this dir instead of copying seems dangerous! dconf's ability to restore the default values relies on applications having set up dconf schemas. Not all applications do this, in particular Gnome Shell extensions regularly fail to install a schema. Hence, moving instead of copying the dconf dir may mean you are losing the entire configuration for some applications with sometimes no way of restoring to default properly and you may run into more trouble than there was before.

   So... with that, some additions:
   * GNOME Shell _and_ modern GNOME applications write their data to the following trifecta of dirs: `.cache` (for cache data), `.config` (for configuration data), `.local/share` (for user data/history and app launcher). In addition, Flatpak applications use `.var` for all of their data.
   * As explained above, I would just delete `.cache`.
   * I am not sure what to do about `.local/share`,  moving the entire dir means you probably have to restore some parts of it later, e.g. to get your emails and chat logs back. However, it also contains GNOME Shell extensions.
   * `.config` is basically the same deal, you probably want most of it back later.
   * `.var` should be irrelevant for the current use case 
   * So, with `.local`/`.config`, maybe it would be best to pick a few specific dirs that are relevant e.g. `gnome-shell` and `gnome-settings-daemon`, `xorg` instead of just moving everything and then moving it back later..? You might need some assistance from someone with more desktop knowledge to know what's really relevant here.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,25 @@
...
-    logs in, the machine works normally. It is then likely that the problem can
-    be fixed relatively quickly by simply moving the user's &gnome;
-    configuration directory to a new location, which causes &gnome; to
-    initialize a new one. Although the user is forced to reconfigure &gnome;,
-    no data is lost.
+     If you are experiencing issues with the &gnome; desktop (for example,
+     incorrect screen geometry, blank desktop, etc.), try resetting &gnome; settings.
+     Before you do this, back up the existing configuration files using the following command:
    </para>
-   <procedure xml:id="proc-trouble-login-dk">
```
sknorr, line 1094: Why did you not keep the procedure structure here? Moving to paras seems like a step back.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,25 @@
...
-      application data and reconfigure the remainder of the applications.
-     </para>
-    </step>
-   </procedure>
+<screen>mkdir ./old-gnome-config/ &amp;&amp; mv ./.gnome* ./old-gnome-config/ &amp;&amp; mv .gconf* ./old-gnome-config/ &amp;&amp; mv ./.cache ./old-gnome-config/ &amp;&amp; mv ./dbus ./old-gnome-config/ &amp;&amp; mv ./.dmrc ./old-gnome-config/ &amp;&amp; mv ~/.config/dconf/* ./old-gnome-config/</screen>
+   <para>
+     Run the <command>dconf reset -f /org/gnome/</command>, then log out and try to log in again.
+   </para>
+   <para>
+     If this fixed the desktop issues, restore the desired configuration files.
```
sknorr, line 1093: Finding out which ones are actually "desired" might be a bit troubling for Mr Joe Random User. Even I don't _really_ know.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,25 @@
...
+   </para>
+   <para>
+     In case you still have problems with the desktop, try the following command:
+   </para>
+<screen>mv .config/dconf/user .config/dconf/user.bak &amp;&amp; sudo reboot</screen>
```
sknorr, line 1098: Why reboot for user config? A fresh log in should be enough.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,25 @@
...
-    work, the screen geometry becoming distorted, or even the screen coming up
-    as a bare gray field. The important distinction is that if another user
-    logs in, the machine works normally. It is then likely that the problem can
-    be fixed relatively quickly by simply moving the user's &gnome;
-    configuration directory to a new location, which causes &gnome; to
-    initialize a new one. Although the user is forced to reconfigure &gnome;,
-    no data is lost.
+     If you are experiencing issues with the &gnome; desktop (for example,
+     incorrect screen geometry, blank desktop, etc.), try resetting &gnome; settings.
+     Before you do this, back up the existing configuration files using the following command:
```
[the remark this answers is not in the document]
dmpop: Good points! I'll update the text. Thanks, Stefan!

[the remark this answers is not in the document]
dmpop: Come to think of it: actually the`mv` command is exactly what is needed, as it removes potentially problematic/corrupted config file and forces Gnome to recreate them.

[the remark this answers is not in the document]
sknorr: Right which is why I tried to suggest keeping the command below but updating the wording here.

xml/troubleshooting.xml
yifanjiang: True, it is troubling. I usually create a fresh new user to exclude all possibly configuration issues in the first place. If that troubleshoots the issue, I add the problematic $HOME directory items to the new $HOME incrementally and find out which one caused the issue. Then it is probably easier to find out the directory of interests. Using this method, I would not be bothered by losing data in the problematic $HOME (I majorly use cp/rsync instead mv now).

However I can barely find a good place to insert this idea nicely in the guide and it may over lengthen the guide. So I am just putting it here for your reference in case it is useful.

xml/troubleshooting.xml
yifanjiang: I agree with Stefan's comments. With some additional thoughts:

1. am I right ./dbus is actually ./.dbus (with a dot prefix)?

2. .local/share and .config map to  $XDG_DATA_HOME and $XDG_CONFIG_HOME respectively in freedesktop standard: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

Their usage is quite widely adapted in GNOME environment that the former is used for "user-specific data", and the latter is used for "user-specific configuration". So I could hardly find a way to detach selective parts in the directories for general troubleshooting.

Theoretically, for example, an arbitrary faulty application which automatically launches (as a part of /etc/xdg/autostart or systemd user session) could contribute to prohibit a successful desktop login. However we could hardly know what the application is when writing the guide.

Maybe a batch move is the effective choice by now at cost of efficiency.

3. @sknorr Just out of curiosity, I read it many times, but have trouble to understand the reasoning of ~/.config/dconf/* part, do we have some examples for a failure case when we need to restore from a backup .config/dconf/* ?

dmpop: Thanks for your input, Stefan and Yifan! OK, to sum up:

- Omit `.dmrc`
- Omit `.gnome*` because removing them won't help solving Gnome issues
- Omit `.config/gconf` because it doesn't do anything
- Delete `.cache`

1. What should we do with `~/.config/dconf/*` if moving it is not an option?
2. We are not sure what to do with `.local/share` Perhaps instead of moving the diretory, we should suggest disabling Gnome extensions?
3. Regarding picking specific dirs from `.local/.config` Are we in agreement that we the directories  that need to be backed up and then restored are `gnome-shell`, `gnome-settings-daemon`, and `xorg`? @yifanjiang Could you confirm that, please?

> Their usage is quite widely adapted in GNOME environment that the former is used for "user-specific data", and the latter is used for "user-specific configuration". So I could hardly find a way to detach selective parts in the directories for general troubleshooting.

So what should we do with these directories?

xml/troubleshooting.xml
dmpop: Maybe this is actually the safest and most efficient way to troubleshoot the problem in the first place? Create a new user, and then "add the problematic $HOME directory items to the new $HOME incrementally and find out which one caused the issue."

dmpop:
I'm going to create a new PR describing a better procedure based on the received feedback.

xml/troubleshooting.xml
sknorr: Even with that approach, I think it might be an issue that dconf is rather hard to separate into different pieces. Also not sure whether all your old config will work again when drop it in piece-meal, there may be dependencies between different dirs. But maybe my criticism are just an example for the perfect being the enemy of the good. :)

PR #584, opened 2020-04-22
Automatic Assignment of PR-Reviews

taroth21:
Add auto_assign.yml. As some PRs probably need a review by a second person, not sure if we really should handle it this way. 

@fsundermeyer @sknorr @tomschr : What do you think?


taroth21:
Discussed with Frank- it does not necessarily make sense for SLE.

PR #597, opened 2020-05-27
NM connection editor not installed by default (bsc#1169867)

tbazant:
### Description
NM connection editor is no longer installed by default in SLE15SP2. This update includes a hint
to install it manually in case you need it.

### DIFF
--- a/xml/nm.xml
+++ b/xml/nm.xml
@@ -146,8 +146,8 @@ http://www.heise.de/netze/artikel/WLAN-und-LAN-sichern-mit-IEEE-802-1X-und-Radiu
    <title>&nm; Connection Editor</title>
    <para>
     In previous &productname; releases, network connections were configured
-    using the application called <emphasis>&nm; Connection Editor</emphasis>.
-    It is no longer installed by default because <emphasis>&gnome; Control
+    using an application called <emphasis>&nm; Connection Editor</emphasis>.
+    This is no longer installed by default, because <emphasis>&gnome; Control
      Center</emphasis> has fully replaced its configuration capabilities.
    </para>
    <para>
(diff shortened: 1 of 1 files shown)


xml/nm.xml
```diff
@@ -142,6 +142,22 @@ http://www.heise.de/netze/artikel/WLAN-und-LAN-sichern-mit-IEEE-802-1X-und-Radiu
    connections.
   </para>
 
+  <tip>
+   <title>&nm; Connection Editor</title>
+   <para>
+    In previous &productname; releases, network connections were configured
+    using the application called <emphasis>&nm; Connection Editor</emphasis>.
```
lproven, line 149:
```suggestion
    using an application called <emphasis>&nm; Connection Editor</emphasis>.
```

xml/nm.xml
```diff
@@ -142,6 +142,22 @@ http://www.heise.de/netze/artikel/WLAN-und-LAN-sichern-mit-IEEE-802-1X-und-Radiu
...
+    It is no longer installed by default because <emphasis>&gnome; Control
```
lproven, line 150:
```suggestion
    This is no longer installed by default, because <emphasis>&gnome; Control
```

xml/nm.xml
```diff
@@ -142,6 +142,22 @@ http://www.heise.de/netze/artikel/WLAN-und-LAN-sichern-mit-IEEE-802-1X-und-Radiu
...
+     Center</emphasis> has fully replaced its configuration capabilities.
+   </para>
+   <para>
+    If you still need to use &nm; Connection Editor to configure network
```
lproven, line 154:
```suggestion
    If you still need to use the &nm; Connection Editor to configure network
```

lproven: Question rather than suggestion: would it not be better to say "if you still _want_ to use..." rather than "need to"? If the Control Center really does everything then strictly it is not a need, right?

tbazant: If they refuse to use GNOME and prefer for example XFCE instead, they need to use it. And exactly for such people we're mentioning the connection editor.

PR #598, opened 2020-05-27
Add new section on configuring the TasksMax settings in systemd

cjschroder:
Resolves Jira #SLE-10123

Please review especially for technical correctness. Not sure that the section on configuring individual users is correct.

### Description

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

### DIFF
--- a/xml/tuning_cgroups.xml
+++ b/xml/tuning_cgroups.xml
@@ -236,9 +236,10 @@
   <para>
       However, it soon became apparent that there is not a single
       default that applies to all use cases. 512 is not low enough
-      to prevent a runaway process from crashing a system, and not
-      high enough for processes that create a lot of threads, such
-      as databases. In &systemd; 234 the default was changed to 15%, 
+      to prevent a runaway process from crashing a system, especially 
+      when other resources such as CPU and RAM are not restricted, 
+      and not high enough for processes that create a lot of threads, 
+      such as databases. In &systemd; 234 the default was changed to 15%, 
       which is 4915 tasks (15% of the kernel limit of 32768, 
       see <command>cat /proc/sys/kernel/pid_max</command>).
   </para>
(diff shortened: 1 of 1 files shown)


xml/tuning_cgroups.xml
```diff
@@ -221,8 +220,130 @@
...
+       Verify this the same way as for system services:
+   </para>
+   <screen>&prompt.user;systemctl show -p TasksMax user-1000.slice
+TasksMax=18446744073709551615
+   </screen>
+   <para>
+       Change this for <literal>user-1000.slice</literal> to 256:
+   </para>
+   <screen>&prompt.sudo;systemctl edit user@1001.service
+[Service]
```
fbuihuu, line 336: I think the example you gave is a bit misleading because it only set a limits on the number of processes spawned by the systemd user instance.

In order to limit the number of *all* processes spawned by a user (which includes all processes started from user sessions) you need to set the limit on the user's slice instead. For example:

`cat /etc/systemd/system/user-1000.slice.d/50-tasksmax.conf
[Slice]
TasksMax=256`

xml/tuning_cgroups.xml
```diff
@@ -221,8 +220,130 @@
...
+       This is the same as having no limit. It is not a requirement to change the 
+       default, but setting some limits may help to prevent system crashes
+       from runaway processes.
+   </para>
+   <para>
+       Change the global default in
+       <filename>/etc/systemd/system.conf</filename> by uncommenting the 
+       <literal>DefaultTasksMax</literal> line, and changing the value either
+       to a number of tasks, or a percentage, like the following examples:
+   </para>
```
fbuihuu, line 271: Are you sure this works ?

IIRC drop-in files wherever they're located (could be in /usr or /run or ...) take precedence over their main configuration files wherever it's located. This part is debatable but it's how it's implemented so far so we have to deal with it.

So to change the SUSE default, the drop-in itself must be overridden:

```
cat >/etc/systemd/system.conf.d/50-tasksmax.conf
[Manager]
DefaultTasksMax=128
```

or all SUSE specific defaults should be masked and hence the defaults are restored to their upstream values: 

`ln -s /dev/null /etc/systemd/system.conf.d/20-suse-defaults.conf`

xml/tuning_cgroups.xml
fbuihuu: Also please note that this only works with SLE15 or later.

cjschroder:
thank you @fbuihuu, much appreciate your help. Please look at this again-- all the steps seem to work on my test systems. I think it is not necessary to disable the SUSE defaults, but it doesn't hurt anything, and it shows how to do it.

cjschroder:
Here is nice PDF for reviewing: 
[sec-tuning-cgroups-tasksmax_color_en.pdf](https://github.com/SUSE/doc-sle/files/4697269/sec-tuning-cgroups-tasksmax_color_en.pdf)

xml/tuning_cgroups.xml
```diff
@@ -221,8 +220,177 @@
...
+      system resources.
+  </para>
+  <para>
+      However, it soon became apparent that there is not a single
+      default that applies to all use cases. 512 is not low enough
+      to prevent a runaway process from crashing a system, especially 
+      when other resources such as CPU and RAM are not restricted, 
+      and not high enough for processes that create a lot of threads, 
+      such as databases. In &systemd; 234 the default was changed to 15%, 
+      which is 4915 tasks (15% of the kernel limit of 32768,
```
lproven, line 243:
```suggestion
      which is 4915 tasks (15% of the kernel limit of 32768; 
```

lproven: Comma splice.

fbuihuu:
>  I think it is not necessary to disable the SUSE defaults, but it doesn't hurt anything, and it shows how to do it.

It's not necessary indeed and it should only be done if one wishes to restore the upstream defaults.

So I would suggest to differentiate this case with the other one which consists in overriding the SUSE defaults.

xml/tuning_cgroups.xml
```diff
@@ -221,8 +220,177 @@
...
+   Status: "Taking your SQL requests now..."
+    Tasks: 30 (limit: 256)
+   CGroup: /system.slice/mariadb.service
+           └─11845 /usr/sbin/mysqld --defaults-file=/etc/my.cnf --user=mysql
+   </screen>
+   <para>
+       The Tasks line shows that MariaDB currently has 30 tasks running, and has
+       an upper limit of 256. The following example demonstrates how to raise
+       MariaDB's limit to 8096. Create a new override file,
+       <filename>/etc/systemd/system.conf.d/30-mariadb.conf</filename>,
```
fbuihuu, line 317: To override the limit for MariaDB only, the drop-in file needs to be located in `/etc/systemd/system/mariadb.service.d`.

fbuihuu:
I would also suggest to create 4 different sub-sections for more clarity:

1. Override the default limit for system services
2. Override the limit for a specific system service (mariadb in your case)
3. Override the default limit set for users
4. Override the limit for a given user

xml/tuning_cgroups.xml
```diff
@@ -221,8 +220,177 @@
...
+       The new value takes effect at the next user login.
+   </para>
+   <para>
+       How do you know what values to use? This varies according to your workloads,
+       system resources, and other resource configurations. When your user TasksMax 
+       value is too low, you will see error messages like "Failed to fork (Resources
+       temporarily unavailable)". Test this by setting the user TasksMax value
+       to 128, then open some applications. To recover, close all apps 
+       (except your terminal) and change the setting.
+   </para>
```
Werkov, line 379: The ultimate hint for troubleshooting is checking dmesg/syslog for messages from kernel that tells which unit (cgroup) hit the limit:

> [7835271.819498] cgroup: fork rejected by pids controller in /system.slice/run-r61c17f6d334540b599c366514bc1d492.service

xml/tuning_cgroups.xml
```diff
@@ -221,8 +220,177 @@
...
+   <para>
+       The Tasks line shows that MariaDB currently has 30 tasks running, and has
+       an upper limit of 256. The following example demonstrates how to raise
+       MariaDB's limit to 8096. Create a new override file,
+       <filename>/etc/systemd/system.conf.d/30-mariadb.conf</filename>, 
+       with the following lines, then reload:
+   </para>
+   <screen>
+[Service]
+TasksMax=8096
```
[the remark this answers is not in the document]
cjschroder: Yes, I always mix that up with 4096.

PR #601, opened 2020-06-02
Gnome troubleshooting procedure

dmpop:
### Description
Rewrite of the Gnome troubleshooting procedure based on  feedback from Yifan Jiang.

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

### DIFF
--- a/xml/troubleshooting.xml
+++ b/xml/troubleshooting.xml
@@ -1108,16 +1108,18 @@
      <step>
        <para>
 	 Log out and log in as the new user. This gives with a fresh
-	 installation of &gnome;.
+	  &gnome; environment.
        </para>
      </step>
      <step>
        <para>
-	 Copy individual &gnome; &gnome;-specific configuration files from the
-	 <filename>.local/.config/</filename> directory of the old
-	 user account to the new one. Log out and log in again as the
-	 new user after every copy operation to check whether &gnome;
-	 still works correctly.
+	 Copy individual &gnome; &gnome;-specific configuration files
+	 from the <filename>~/.local/</filename> and
+	 <filename>~/.config/</filename> directories of the old user
+	 account to the respective directories of the new user
+	 account. Log out and log in again as the new user after every
+	 copy operation to check whether &gnome; still works
+	 correctly.
        </para>
      </step>
      <step>
@@ -1128,8 +1130,9 @@
      </step>
      <step>
        <para>
-	 Log in as the old user, and delete the offending configuration file. Log
-	 out and log in again as the old user.
+	 Log in as the old user, and move the offending configuration
+	 file to a different location. Log out and log in again as the
+	 old user.
        </para>
      </step>
      <step>
--- a/xml/troubleshooting.xml
+++ b/xml/troubleshooting.xml
@@ -1081,21 +1081,63 @@
   <sect2 xml:id="sec-trouble-login-dk">
    <title>&gnome; Desktop Has Issues</title>
    <para>
-     If you are experiencing issues with the &gnome; desktop (for example,
-     incorrect screen geometry, blank desktop, etc.), try resetting &gnome; settings.
-     Before you do this, back up the existing configuration files using the following command:
+     
+     If you are experiencing issues with the &gnome; desktop (for
+     example, incorrect screen geometry, blank desktop, etc.), there
+     are several ways to troubleshoot the misbehaving graphical
+     desktop environment. The recommended procedure described below
+     offers the safest option to fix broken &gnome;.
    </para>
-<screen>mkdir ./old-gnome-config/ &amp;&amp; mv ./.gnome* ./old-gnome-config/ &amp;&amp; mv .gconf* ./old-gnome-config/ &amp;&amp; mv ./.cache ./old-gnome-config/ &amp;&amp; mv ./dbus ./old-gnome-config/ &amp;&amp; mv ./.dmrc ./old-gnome-config/ &amp;&amp; mv ~/.config/dconf/* ./old-gnome-config/</screen>
-   <para>
-     Run the <command>dconf reset -f /org/gnome/</command>, then log out and try to log in again.
-   </para>
-   <para>
-     If this fixed the desktop issues, restore the desired configuration files.
-   </para>
-   <para>
-     In case you still have problems with the desktop, try the following command:
-   </para>
-<screen>mv .config/dconf/user .config/dconf/user.bak &amp;&amp; sudo reboot</screen>
+   <procedure>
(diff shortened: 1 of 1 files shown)


xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,67 @@
...
+     </step>
+     <step>
+       <para>
+	 Repeat the previous step until you find the configuration file that
+	 breaks &gnome;.
+       </para>
+     </step>
+     <step>
+       <para>
+	 Log in as the old user, and delete the offending configuration file. Log
```
yifanjiang, line 1131: "delete the offending configuration file" sounds dangerous :-) Can we have something like moving the configuration file/directory somewhere else for further analysis (eg. there might be a bug the user wants to report)?

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,67 @@
...
+     <step>
+       <para>
+	 Log out and log in as the new user. This gives with a fresh
+	 installation of &gnome;.
+       </para>
+     </step>
+     <step>
+       <para>
+	 Copy individual &gnome; &gnome;-specific configuration files from the
+	 <filename>.local/.config/</filename> directory of the old
```
yifanjiang, line 1117: As I understood what it implies here, perhaps you meant .local and .config directories under the user's home? So they are two different directories. The format in line 1117 looks like they are one single directory.

Besides, I remember Stefan mentioned similar thing in an earlier PR, we'd better to guide the user to the corresponding path (eg. cd $HOME) before using the relative path to explain things. Please ignore this if
it has been mentioned in the context around the changing segments.

xml/troubleshooting.xml
dmpop: Good idea! I'll update the text.

xml/troubleshooting.xml
dmpop: My bad, these are indeed two different directories.

> Besides, I remember Stefan mentioned similar thing in an earlier PR, we'd better to guide the user to the corresponding path (eg. cd $HOME)

True, but the reader needs to copy files from the old user account to the new one. So using $HOME is not practical here. But I'll see if I can do something about it.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,67 @@
...
+     </step>
+     <step>
+       <para>
+	 Fill out the required fields and press <guimenu>OK</guimenu> to create a new user.
+       </para>
+     </step>
+     <step>
+       <para>
+	 Log out and log in as the new user. This gives with a fresh
+	 installation of &gnome;.
```
yifanjiang, line 1111: More precisely, instead of providing a "fresh installation", log-in as a new user provides "a fresh GNOME user environment" or "a fresh user GNOME session".

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,70 @@
...
-    work, the screen geometry becoming distorted, or even the screen coming up
-    as a bare gray field. The important distinction is that if another user
-    logs in, the machine works normally. It is then likely that the problem can
-    be fixed relatively quickly by simply moving the user's &gnome;
-    configuration directory to a new location, which causes &gnome; to
-    initialize a new one. Although the user is forced to reconfigure &gnome;,
-    no data is lost.
+     
+     If you are experiencing issues with the &gnome; desktop (for
+     example, incorrect screen geometry, blank desktop, etc.), there
```
sknorr, line 1086: As noted in the previous PR already:

1. please remove "incorrect screen geometry", extremely unlikely to be relevant in the modern era
2. please stick to our standard 1-space indentation -- this PR completely disregards it

Also, a point newly noticed:
* "for example ... etc." is redundant.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,70 @@
...
+     are several ways to troubleshoot the misbehaving graphical
+     desktop environment. The recommended procedure described below
+     offers the safest option to fix broken &gnome;.
```
sknorr, line 1089: "...a broken \&gnome; desktop." sounds better to me.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,70 @@
...
+   <procedure>
+     <title>Troubleshooting &gnome;</title>
+     <step>
+       <para>
+	 Launch &yast; and switch to <guimenu>Security and Users</guimenu>.
+       </para>
+     </step>
+     <step>
+       <para>
+	 Open the <guimenu>User and Group Management</guimenu> dialog and press <guimenu>Add</guimenu>.
```
sknorr, line 1100: "click," not "press" because this is a GUI button and not a physical keyboard key or mouse button.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,70 @@
...
+       </para>
+     </step>
+     <step>
+       <para>
+	 Fill out the required fields and press <guimenu>OK</guimenu> to create a new user.
```
sknorr, line 1105: Same here.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,70 @@
...
+       </para>
+     </step>
+     <step>
+       <para>
+	 Log out and log in as the new user. This gives with a fresh
+	  &gnome; environment.
```
sknorr, line 1111: "This gives with" sounds broken.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,70 @@
...
+       </para>
+     </step>
+     <step>
+       <para>
+	 Copy individual &gnome; &gnome;-specific configuration files
```
sknorr, line 1116: * "\&gnome; \&gnome;"?
* content-wise: I'd use something more like "Copy individual subdirectories of the .local/share and .config directories." The spiel about GNOME-specific dirs is not really helpful or specific.

xml/troubleshooting.xml
```diff
@@ -1079,64 +1079,70 @@
...
+	 from the <filename>~/.local/</filename> and
+	 <filename>~/.config/</filename> directories of the old user
+	 account to the respective directories of the new user
+	 account. Log out and log in again as the new user after every
+	 copy operation to check whether &gnome; still works
+	 correctly.
```
sknorr, line 1122: Last sentence should probably be a new step, but even if you want to keep it as one step, it should definitely have its own para.

PR #602, opened 2020-06-03
AY scripts: improve and adapt to SP3 changes

jreidinger:
### Description
Adapt to the latest changes in autoyast - PR https://github.com/yast/yast-autoinstallation/pull/612

### Checklist
* Check all items that apply.

*Are backports required?*

no, new changes for SP3

### DIFF
--- a/xml/ay_custom_user_scripts.xml
+++ b/xml/ay_custom_user_scripts.xml
@@ -321,6 +321,16 @@ echo "Testing the init script" &gt;
      described above, except for the <literal>chrooted</literal> element,
      which can only be used in chroot scripts.
     </para>
+    <important>
+      <title>Deprecated Elements</title>
+      <para>
+       <literal>debug</literal> is deprecated element and can be removed in future
+       releases. To adapt use it as parameter in <literal>interpreter</literal>.
+       E.g. instead of "&lt;interpreter&gt;shell&lt;/interpreter&gt; use
+       &lt;interpreter&gt;/bin/sh -x&lt;/interpreter&gt; and result is same
+       as having enabled debug flag.
+      </para>
+    </important>
     <table>
      <title>Script XML Representation</title>
      <tgroup cols="3">
@@ -405,14 +415,16 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
         <entry>
          <para>
           Specify the interpreter that must be used for the script.
-          Any interpreter available at given environment is possible
-          to specify. Full path even with parameters is possible.
+          Any interpreter available in the given environment can be specified.
+          It is possible to provide a full path to the interpreter, including parameters.
+          There are also deprecated keywords interpreter "shell", "perl" and "python" that are
+          supported by <literal>debug</literal> flag.
          </para>
 <screen>&lt;interpreter&gt;/bin/bash -x&lt;/interpreter&gt;</screen>
         </entry>
         <entry>
          <para>
-          Optional; default is <literal>/bin/sh</literal>.
+          Optional; default is <literal>shell</literal>.
          </para>
         </entry>
        </row>
@@ -433,9 +445,9 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
          <para>
           Optional; default is the type of the script (pre-scripts in this
           case). If you have more than one script, you should define
-          different names for each script. If it is not defined and
-          <literal>location</literal> is defined, it will use filename from
-          location path.
+          different names for each script. If <literal>filename</literal>
+          is not defined and <literal>location</literal> is defined,
+          filename from location path will be used.
          </para>
         </entry>
        </row>
@@ -489,13 +501,15 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
          <para>
           If this is <literal>true</literal>, every single line of a shell
           script is logged. Perl scripts are run with warnings turned on.
-          For other interpreters use directly parameter in interpreter field.
+          It works only for deprecated keywords <literal>interpreter</literal>.
+          For others use any interpreter modification as parameters in
+          <literal>interpreter</literal> value like e.g. "&lt;interpreter&gt;ruby -w&lt;/interpreter&gt;".
          </para>
 <screen>&lt;debug config:type="boolean"&gt;true&lt;/debug&gt;</screen>
         </entry>
         <entry>
          <para>
-          Optional; default is <literal>false</literal>.
+          Optional; default is <literal>true</literal>.
          </para>
         </entry>
        </row>
@@ -612,7 +626,6 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
       &lt;filename&gt;chroot-post.sh&lt;/filename&gt;
       &lt;interpreter&gt;shell&lt;/interpreter&gt;
       &lt;source&gt;&lt;![CDATA[
-#!/bin/sh
 echo "Testing chroot (chrooted) scripts"
 ls
 ]]&gt;
@@ -622,7 +635,6 @@ ls
       &lt;filename&gt;chroot-pre.sh&lt;/filename&gt;
         &lt;interpreter&gt;/bin/bash -x&lt;/interpreter&gt;
         &lt;source&gt;&lt;![CDATA[
-#!/bin/bash
 echo "Testing chroot scripts"
 df
 cd /mnt
@@ -636,8 +648,6 @@ ls
         &lt;filename&gt;post.sh&lt;/filename&gt;
         &lt;interpreter&gt;shell&lt;/interpreter&gt;
         &lt;source&gt;&lt;![CDATA[
-#!/bin/sh
-
 echo "Running Post-install script"
 systemctl start portmap
 mount -a 192.168.1.1:/local /mnt
@@ -650,9 +660,7 @@ umount /mnt
         &lt;filename&gt;post.pl&lt;/filename&gt;
         &lt;interpreter&gt;perl&lt;/interpreter&gt;
         &lt;source&gt;&lt;![CDATA[
-#!/usr/bin/perl
 print "Running Post-install script";
-
 ]]&gt;
         &lt;/source&gt;
       &lt;/script&gt;
@@ -666,7 +674,6 @@ print "Running Post-install script";
         &lt;filename&gt;pre.sh&lt;/filename&gt;
         &lt;interpreter&gt;shell&lt;/interpreter&gt;
         &lt;source&gt;&lt;![CDATA[
-#!/bin/sh
 echo "Running pre-install script"
 ]]&gt;
         &lt;/source&gt;
--- a/xml/ay_custom_user_scripts.xml
+++ b/xml/ay_custom_user_scripts.xml
@@ -696,13 +696,13 @@ echo Hi
     </para>
     <para>
      The log consists of the output produced when executing the
-     scripts containing both standart output and standard error output
-     together.
+     scripts, containing both the standard output and the standard
+     error output combined.
     </para>
     <para>
-     If the script ends up with non zero exit code then warning will
-     be shown with logs content unless <literal>feedback</literal> option
-     is provided.
+     If the script ends with a non-zero exit code, then a warning will
+     be shown with the content of the logs, unless the <literal>feedback</literal>
+     option was provided.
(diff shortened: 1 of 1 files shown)


xml/ay_custom_user_scripts.xml
```diff
@@ -402,14 +405,14 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
         <entry>
          <para>
           Specify the interpreter that must be used for the script.
-          Supported options are <literal>shell</literal> and
-          <literal>perl</literal>.
+          Any interpreter available at given environment is possible
+          to specify. Full path even with parameters is possible.
```
mvidner, line 409: You are losing the documentation of the `shell` and `perl` special values. Also, a `python` special value has been supported by the code.

lproven, line 409: This needs to be re-phrased.

I would suggest:

"Any interpreter available in the given environment can be specified. It is possible to provide a full path to the interpreter, including parameters."

xml/ay_custom_user_scripts.xml
```diff
@@ -430,7 +433,9 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
          <para>
           Optional; default is the type of the script (pre-scripts in this
           case). If you have more than one script, you should define
-          different names for each script.
+          different names for each script. If it is not defined and
```
lproven, line 436: > If it is not defined and

If **what** is not defined?

xml/ay_custom_user_scripts.xml
```diff
@@ -430,7 +433,9 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
...
+          <literal>location</literal> is defined, it will use filename from
```
lproven, line 437: > it will use filename from

What will use (the) filename? Filename of what? What is (the) "location path"?

xml/ay_custom_user_scripts.xml
```diff
@@ -484,12 +489,13 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
          <para>
           If this is <literal>true</literal>, every single line of a shell
           script is logged. Perl scripts are run with warnings turned on.
+          For other interpreters use directly parameter in interpreter field.
```
lproven, line 492: > For other interpreters use directly parameter in interpreter field.

What parameter? What field? Suffixed onto the name of the binary of the language?

xml/ay_custom_user_scripts.xml
```diff
@@ -484,12 +489,13 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
...
          </para>
 <screen>&lt;debug config:type="boolean"&gt;true&lt;/debug&gt;</screen>
         </entry>
         <entry>
          <para>
-          Optional; default is <literal>true</literal>.
+          Optional; default is <literal>false</literal>.
```
mvidner, line 498: I've checked the code and indeed you are changing a default that the users may rely on, IMHO totally unnecessarily.

xml/ay_custom_user_scripts.xml
```diff
@@ -689,9 +695,14 @@ echo Hi
      and the output logs in the <filename>log</filename> directory.
     </para>
     <para>
-     The log consists of the output produced when executing the shell
-     scripts using the following command:
+     The log consists of the output produced when executing the
+     scripts containing both standart output and standard error output
+     together.
```
lproven, line 700:
```suggestion
     The log consists of the output produced when executing the
     scripts, containing both the standard output and the standard
     error output combined.
```

xml/ay_custom_user_scripts.xml
```diff
@@ -689,9 +695,14 @@ echo Hi
...
+    </para>
+    <para>
+     If the script ends up with non zero exit code then warning will
+     be shown with logs content unless <literal>feedback</literal> option
+     is provided.
```
lproven, line 705:
```suggestion
     If the script ends with a non-zero exit code, then a warning will
     be shown with the content of the logs, unless the <literal>feedback</literal>
     option was provided.
```

xml/ay_custom_user_scripts.xml
```diff
@@ -613,10 +619,10 @@ ls
       &lt;/source&gt;
     &lt;/script&gt;
     &lt;script&gt;
-      &lt;filename&gt;chroot.sh&lt;/filename&gt;
-        &lt;interpreter&gt;shell&lt;/interpreter&gt;
+      &lt;filename&gt;chroot-pre.sh&lt;/filename&gt;
+        &lt;interpreter&gt;/bin/bash -x&lt;/interpreter&gt;
         &lt;source&gt;&lt;![CDATA[
-#!/bin/sh
+#!/bin/bash
```
mvidner, line 625: Note the CDATA content starts with a newline, which makes the `#!` useless. Also, YaST would ignore it even if it were in the right place. Let's delete it.

(not specific to this \<source>, all sources in this example have this problem)

xml/ay_custom_user_scripts.xml
jreidinger: sounds good.

jreidinger: well, I see it as a bit deprecated less powerful options, but I can add back those three specific values and mention that it is supported by debug flag?

xml/ay_custom_user_scripts.xml
jreidinger: the filename option. I am also not sure how it will look like when para is so long.

xml/ay_custom_user_scripts.xml
jreidinger: @lproven let me show you example - 

You have script test.sh at ftp://neser-vr.suse.cz/test.sh
in autoyast profile you do *not* specify filename and has only <location>ftp://neser-vr.suse.cz/test.sh</location> . So autoyast use as filename "test.sh" ( basically path from parsed URL and apply on it basename so even when it is scripts/test.sh it will use just test.sh )

xml/ay_custom_user_scripts.xml
jreidinger: parameter depends on interpreter. e.g. -x for bash, -w for ruby, etc.
By field I should probably use parameter, but it is confusing with script parameter. Maybe I should talk about XML element?

xml/ay_custom_user_scripts.xml
jreidinger: well, for me it is really strange to have debugging enabled by default. It will also quite mess output. To be honest as autoyast is non-interactive I expect not much customers show any feedback and just let script run hidden.

xml/ay_custom_user_scripts.xml
jreidinger: agreed

xml/ay_custom_user_scripts.xml
jreidinger: sounds good to me. Feel free to improve phrasing as much as you want :wink:

xml/ay_custom_user_scripts.xml
jreidinger: sounds good

xml/ay_custom_user_scripts.xml
jreidinger: also consider now that you can specify in interpreter "sh" or "shell" and if you do not specify debug, it means that sh run just as sh and shell runs with -x...For perl situation is even more strange as "perl" run with -w and "/bin/perl" runs without it.

jreidinger: maybe we can even drop that debug flag and ignore it? and just for backward compatible convert "shell" interpreter to "sh" and kill all keywords which makes it just confusing?

xml/ay_custom_user_scripts.xml
imobachgs: As they are valid, I guess we should keep them in the documentation. Perhaps we can describe them in a `<important/>` section as deprecated.

xml/ay_custom_user_scripts.xml
imobachgs: If we decide to keep the new behavior, we should document that change.

jreidinger: @lproven Liam we discuss it and will revert back to keep same behavior, but we would like to mark this element and also keywords interpreters as deprecated that will be removed in SLE16. Is there way how to do it in docbook?

lproven: I don't know any specific markup for something being deprecated, but we could just add a <info> or <warning> block mentioning it. Would that be enough?

xml/ay_custom_user_scripts.xml
lproven: OK. 

What I was asking was what "it" meant in all these contexts. If "it" usually means "autoyast" then that is how I will edit it to read.

Repetition is more permissible in documentation than in normal writing. :-)

jreidinger:
@lproven @mvidner pull request updated with your notes, please re-review.
Thanks

xml/ay_custom_user_scripts.xml
```diff
@@ -357,6 +367,9 @@ echo "Testing the init script" &gt;
          <para>
           Define a location from where the script gets fetched. Locations
           can be the same as for the control file (HTTP, FTP, NFS, etc.).
+          Additionally relative url can be used that defines path relative to
+          directory where is control file with
+          syntax <literal>relurl://script.sh</literal>.
```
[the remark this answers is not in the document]
jreidinger: yep, we are good at reinventing wheel

xml/ay_custom_user_scripts.xml
```diff
@@ -430,7 +445,9 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
          <para>
           Optional; default is the type of the script (pre-scripts in this
           case). If you have more than one script, you should define
-          different names for each script.
+          different names for each script. If <literal>filename</literal>
+          is not defined and <literal>location</literal> is defined,
+          filename from location path will be used.
```
lproven, line 450:
```suggestion
          the filename from the location path will be used.
```

xml/ay_custom_user_scripts.xml
```diff
@@ -484,6 +501,9 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
          <para>
           If this is <literal>true</literal>, every single line of a shell
           script is logged. Perl scripts are run with warnings turned on.
+          It works only for deprecated keywords <literal>interpreter</literal>.
```
lproven, line 504:
```suggestion
          This only works for the deprecated keyword <literal>interpreter</literal>.
```

xml/ay_custom_user_scripts.xml
```diff
@@ -484,6 +501,9 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
...
+          For others use any interpreter modification as parameters in
```
lproven, line 505:
```suggestion
          For other languages, give the path to the interpreter as a parameter in the 
```

xml/ay_custom_user_scripts.xml
```diff
@@ -484,6 +501,9 @@ echo "Testing the pre script" &gt; /tmp/pre-script_out.txt
...
+          <literal>interpreter</literal> value like e.g. "&lt;interpreter&gt;ruby -w&lt;/interpreter&gt;".
```
lproven, line 506:
```suggestion
          <literal>interpreter</literal> value, for example "&lt;interpreter&gt;ruby -w&lt;/interpreter&gt;".
```

xml/ay_custom_user_scripts.xml
```diff
@@ -689,9 +702,14 @@ echo Hi
      and the output logs in the <filename>log</filename> directory.
     </para>
     <para>
-     The log consists of the output produced when executing the shell
-     scripts using the following command:
+     The log consists of the output produced when executing the
+     scripts, containing both the standard output and the standard
+     error output combined.
```
lproven, line 707:
```suggestion
     scripts, containing a combination of both the standard output
     and the standard error output.
```

jreidinger:
@lproven thanks, all suggestions applied

taroth21:
Thanks everybody! As we have adjusted our merge policy, I have removed the SP3 labels and have merged the changes now.

PR #605, opened 2020-06-08
Integrate Dublin proofing changes to Deployment Guide

lproven:
### Description
Integrate all the changes requested by Dublin proofreading of the DepGuide.

One is possibly controversial: we are not sure if "DHCPv6 DHCP Unique Identifier" is necessary. It seems tautological. As requested, I have removed the 2nd occurrence to leave it as "DHCPv6 Unique Identifier". If the longer version is required because it is formally correct, it is trivial to reverse.

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

### DIFF
--- a/xml/deployment_image.xml
+++ b/xml/deployment_image.xml
@@ -56,7 +56,7 @@
    up and clones can be used immediately.
   </para>
   <para>
-   The <command>clone-master-clean-up</command> command, for example, removes:
+   For example, the <command>clone-master-clean-up</command> command removes:
   </para>
   <itemizedlist>
    <listitem>
--- a/xml/deployment_image.xml
+++ b/xml/deployment_image.xml
@@ -14,7 +14,7 @@
   <abstract>
    <para>
     This chapter describes how to use cloned images for installing &productname;.
-    This is mostly used in virtualized environments.
+    This process is mostly used in virtualized environments.
    </para>
   </abstract>
   <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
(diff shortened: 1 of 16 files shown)


xml/deployment_image.xml
```diff
@@ -56,7 +56,7 @@
    up and clones can be used immediately.
   </para>
   <para>
-   The <command>clone-master-clean-up</command> command for example removes:
```
taroth21, line 59: To prevent interrupting the train of thought, I would move 'For example,' to the beginning of the sentence.

lproven:
> one file (`deployment_prep_power.xml`) has a very large diff, could you please check again if you accidentally reformatted it completely while integrating the proofing corrections?

That was intentional -- I think it got reflowed automatically in the past and the indentation was a mess, so I did it with ``daps-xmlformat``.

taroth21:
> > one file (`deployment_prep_power.xml`) has a very large diff, could you please check again if you accidentally reformatted it completely while integrating the proofing corrections?
> 
> That was intentional -- I think it got reflowed automatically in the past and the indentation was a mess, so I did it with `daps-xmlformat`.

I see - but for the future, please use a separate commit (outside of the PR) for reformatting a complete file (as already suggested by Stefan in another case). Otherwise it's hard to identify the proofing changes as such and impossible to cherry-pick any proofing commits, if required.

lproven:
> I see - but for the future, please use a separate commit (outside of the PR) for reformatting a complete file

Oh. OK then.

> (as already suggested by Stefan in another case). Otherwise it's hard to identify the proofing changes as such and impossible to cherry-pick any proofing commits, if required.

Fair point!

PR #608, opened 2020-06-11
Adding jeos tip for installing Product Patches after firstboot

vmoutoussamy:
### Description
Adding jeos tip for installing Product Patches after firstboot


cjschroder:
Thank you @vmoutoussamy! There was a markup error (you can see these in the Travis checks if you don't build locally), a 'Tip' cannot go inside a 'Procedure'. I corrected it & will merge after the Travis checks are completed.

PR #610, opened 2020-06-16
vt doc: Improve libvirt network section

jfehlig:
Add a warning about enabling forwarding in the VM host when enabling
a libvirt virtual network. While at it, make some small improvements
in the description of libvirt's virtual networks.

*Are backports required?*

- [x ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [x ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

### DIFF
--- a/xml/libvirt_networking.xml
+++ b/xml/libvirt_networking.xml
@@ -8,12 +8,11 @@
  <title>Managing Networks</title>
  <info>
   <abstract>
-   <para>
-    This chapter introduces common networking configurations supported by
-    &libvirt;. It does not depend on the hypervisor used. It is valid for all
-    hypervisors supported by &libvirt;, such as &kvm; or &xen;. These setups
-    can be achieved using both the graphical interface of &vmm; and the command
-    line tool <command>virsh</command>.
+    <para>
+     This chapter describes common network configurations for a &vmhost;,
+     including those supported natively by the &vmhost; and &libvirt;.
+     The configurations are valid for all hypervisors supported by &productname;,
+     such as &kvm; or &xen;.
    </para>
   </abstract>
   <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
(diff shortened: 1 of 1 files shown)


jfehlig:
BTW, this change was inspired by https://bugzilla.suse.com/show_bug.cgi?id=1166570

jfehlig:
After some conversations with Marius, I've force pushed some improvements to the improvements.

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,957 @@
...
   </abstract>
   <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
    <dm:bugtracker></dm:bugtracker>
   </dm:docmanager>
  </info>
  <para>
-  There are two common network setups to provide a &vmguest; with a network
-  connection:
+  There are two common network configurations to provide a &vmguest; with a
+  network connection:
```
hw-claudio, line 24: a guest network connection from where to where?

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,957 @@
...
  </para>
  <itemizedlist>
   <listitem>
    <para>
-    A <emphasis>virtual network</emphasis> for the guest
+    A <emphasis>network bridge</emphasis> acting as a Layer 2 switch
```
hw-claudio, line 29: from the user perspective: when would I choose this?

For me it should be:

network bridge: (who is the subject?) acting as a layer 2 switch:

this configuration is best suited for XXX

Configuring this kind of networking for guest VMs means that VMs will be able to connect from (blabla) and connections from xxx will be able to reach the VM.

Consider these consequences when choosing this network type: (XXX notable consequences and side effects)

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,957 @@
...
    </para>
   </listitem>
   <listitem>
    <para>
-    A <emphasis>network bridge</emphasis> over a host's physical network
-    interface that the guest can use
+    A <emphasis>virtual network</emphasis> managed by &libvirt; with Layer 3
+    forwarding enabled
```
hw-claudio, line 35: Virtual network: libvirt manages the network, relying on layer 3 forwarding

this configuration is best suited for YYY

Configuring this kind of networking for guest VMs means that VMs will be able to (blabla) and hosts located in network ZZZ will be able to reach the VM.

Consider these consequences of choosing this networking type for your guest VMs (QQQ notable consequences and side effects)

xml/libvirt_networking.xml
jfehlig: I don't think "where to where" is necessary. We don't generally say "I need a network connection from my server's ethernet port to a switch". It is enough to say I need a network connection for my server, no?

jfehlig:
Ok, another attempt and another force push. Claudio mentioned some pictures would help, e.g. similar to the ones on the libvirt wiki

https://wiki.libvirt.org/page/VirtualNetworking

What do the doc folks think about that?

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
+     <calloutlist>
+      <callout arearefs="vnet-xml-name">
+       <para>
+        The name of the new virtual network.
+       </para>
+      </callout>
+      <callout arearefs="vnet-xml-bridge">
+       <para>
+        The name of the bridge device used to construct the virtual network.
+        When defining a new network with a &lt;forward> mode of "nat" or
```
lproven, line 769:
```suggestion
        When defining a new network with a &lt;forward&gt; mode of "nat" or
```

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
+        "route" (or an isolated network with no &lt;forward> element),
```
lproven, line 770:
```suggestion
        "route" (or an isolated network with no &lt;forward&gt; element),
```

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
+        &libvirt; will automatically generate a unique name for the bridge
+        device if none is given.
+       </para>
+      </callout>
+      <callout arearefs="vnet-xml-forward">
+       <para>
+        Inclusion of the &lt;forward> element indicates that the virtual
```
lproven, line 777:
```suggestion
        Inclusion of the &lt;forward&gt; element indicates that the virtual
```

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
+        network will be connected to the physical LAN. The
+        <literal>mode</literal> attribute specifies the forwarding method. The
+        most common modes are "nat" (default, network address translation),
+        "route" (direct forwarding to the physical network, no address
+        translation), and "bridge" (network bridge configured outside of
+        &libvirt;). If the &lt;forward> element is not specified, the virtual
```
lproven, line 783:
```suggestion
        &libvirt;). If the &lt;forward&gt; element is not specified, the virtual
```

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
+      <callout arearefs="vnet-xml-dhcp">
+       <para>
+        Enable DHCP server for the virtual network, offering IP addresses
+        ranging from the specified <literal>start</literal> and
+        <literal>end</literal> attribute.
+       </para>
+      </callout>
+      <callout arearefs="vnet-xml-dhcp-host">
+       <para>
+        The optional &lt;host> elements specify hosts that will be given names
```
lproven, line 803:
```suggestion
        The optional &lt;host&gt; elements specify hosts that will be given names
```

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
+      </callout>
+     </calloutlist>
+    </example>
+    <example>
+     <title>Isolated Network</title>
      <para>
-      Select the desired method for assigning an IP address to the VLAN device.
+      This configuration provides a completely isolated private network. The
+      guests can talk to each other, and to &vmhost;, but cannot reach any
+      other machines on the LAN, as the &lt;forward> element is missing in the
```
lproven, line 853:
```suggestion
      other machines on the LAN, as the &lt;forward&gt; element is missing in the
```

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
-    desired feature especially if &vmhost;-only networks should be connected.
-    In this case, use the special address <literal>0.0.0.0</literal> with
-    netmask <literal>255.255.255.255</literal>. The system scripts handle this
-    address as no IP address set.
-   </para>
+<screen>&lt;network>
+        &lt;name>host-bridge&lt;/name>
+        &lt;forward mode="bridge"/>
+        &lt;bridge name="br0"/>
+&lt;/network>
```
lproven, line 880:
```suggestion
<screen>&lt;network&gt;
        &lt;name&gt;host-bridge&lt;/name&gt;
        &lt;forward mode="bridge"/&gt;
        &lt;bridge name="br0"/&gt;
&lt;/network&gt;
```

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
+ <sect1 xml:id="libvirt-networks-bridged">
+  <title>Network Bridge</title>
 
   <para>
-   On the other hand, guests inside a <emphasis>forwarded</emphasis> (NAT,
-   network address translation) virtual network can make any outgoing network
-   connection they request. Incoming connections are allowed from &vmhost;, and
-   from other guests connected to the same virtual network. All other incoming
-   connections are blocked by <systemitem>iptables</systemitem> rules.
+   The network bridge configuration provides a layer 2 switch for &vmguest;s,
```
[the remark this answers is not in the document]
jfehlig: I made all your suggested changes (unless otherwise noted) and force pushed the updated patch.

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
+       <guimenu>Add</guimenu>.
       </para>
      </step>
      <step>
       <para>
-       Specify a name for the new virtual network and click
-       <guimenu>Forward</guimenu>.
+       Select <guimenu>Bridge</guimenu> from the <guimenu>Device Type</guimenu>
+       list and enter the bridge device interface name in the
+       <guimenu>Configuration Name</guimenu> entry. Proceed with
```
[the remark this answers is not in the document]
jfehlig: I didn't review most of the stuff I moved, so thanks for doing that! Might as well fix up other issues while in here.

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
-   outgoing network access.
+   A standard &libvirt; installation on &productname; already comes with a
+   predefined virtual network named "default" that provides DHCP and DNS
+   services for the network along with connectivity to the &vmhost;'s physical
+   network using the network address translation (NAT) forwarding mode.
+   Although predefined, the "default" virtual network must be explicitly enabled
+   by the administrator. For more information on the forwarding modes supported
+   by &libvirt; see the <citetitle>Connectivity</citetitle> section of the
+   <citetitle>Network XML format</citetitle> documentation at
+   <link xlink:href="https://libvirt.org/formatnetwork.html#elementsConnect"/>.
```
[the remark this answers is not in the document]
jfehlig: I also added <literal> around the 'default' network name, similar to your earlier suggestion.

xml/libvirt_networking.xml
jfehlig: There are several instances of '>' vs '&gt;' in a similar context, e.g. lines 745-760, 830-839 and 858-866. Should those be changed as well?

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
   </para>
 
   <para>
-   If there is a network bridge present on &vmhost;, you can connect a
-   &vmguest; to it directly. This provides the &vmguest; with full incoming and
-   outgoing network access.
+   A standard &libvirt; installation on &productname; already comes with a
+   predefined virtual network named "default" that provides DHCP and DNS
+   services for the network along with connectivity to the &vmhost;'s physical
+   network using the network address translation (NAT) forwarding mode.
```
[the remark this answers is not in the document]
jfehlig: I think it is best to keep "forwarding mode", since NAT is one of the modes available with the forwarding element

https://libvirt.org/formatnetwork.html#elementsConnect

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,976 @@
...
-     </step>
-     <step>
       <para>
-       To enable static routing for the new virtual network, activate the
-       relevant option and type the network and gateway addresses.
+       If you intend to create a virtual bridge that has no connection to a
+       real Ethernet device, use <guimenu>Statically assigned IP
+       Address</guimenu>. In this case, it is a good idea to use addresses from
+       the private IP address ranges, for example,
+       <literal>192.168.x.x</literal> or <literal>10.x.x.x</literal>.
```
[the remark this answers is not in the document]
jfehlig: I changed the x to 0 and included the mask for each class, which IMO is more standard notation.

xml/libvirt_networking.xml
lproven: Yes, I think they should.

jfehlig: Thanks. I've changed all occurrences and force pushed the latest patch. I think I've covered all review comments.

xml/libvirt_networking.xml
hw-claudio: Hi Jim, I don't think that from a user perspective it is clear what kind of use case each network configuration allows, and which limitations it has.

As an example, with out of the box upstream qemu user networking, there are a lot of limitations for the VM.

The VM can initiate a connection to the external network, but the VM is not reachable from the external connections, so it can not run a reachable service this way. Also ICMP is limited etc.

With these two network configurations, what are, if any, the limitations for the user? Is the VM going to be visible externally, and how? Any limitations to the traffic type? Any performance consideration of the one versus the other for the user?

xml/libvirt_networking.xml
```diff
@@ -8,958 +8,978 @@
...
 
   <para>
-   A standard libvirt installation on &productname; already comes with a
-   predefined virtual network providing DHCP server and network address
-   translation (NAT) named "default".
+   The network bridge configuration is the default configuration of
+   &productname; when configured as a &kvm; or &xen; hypervisor. It is the
+   preferred configuration when you simply want to connect &vmguest;s to the
+   &vhhost;'s LAN.
   </para>
```
hw-claudio, line 57: I like this change, from a user perspective it is much clearer what the goal of the configuration is.

xml/libvirt_networking.xml
hw-claudio: For inspiration we could look at something like:
https://jamielinux.com/docs/libvirt-networking-handbook/

We do not need to include all options like there, just the two we care about and support.

But rather than an in-detail description of everything under the sun (which can still come later), the entry point for the user could be: how do I choose the right networking for me? What will it allow me to do, and what will it instead be limited? Does it fit my requirements?

jfehlig: I added a bit more content to the generic description of bridged network vs libvirt virtual network and will force push shortly. E.g. I mentioned that layer 3 forwarding "introduces additional packet processing on the &vmhost; as compared to a Layer 2 bridged network." IMO this PR is a huge improvement over the existing text. Further improvements can be explored when/if we have a "doc day".

xml/libvirt_networking.xml
jfehlig: Thanks. I think my latest tweak provides another small, incremental improvement.

cwickert:
Umm, this actually broke the build because of a typo in an entity:

```
/usr/src/app/xml/libvirt_networking.xml:56: parser error : Entity 'vhhost' not defined

   &vhhost;'s LAN.
```

You can see this in travis, but due to [a bug](https://github.com/openSUSE/daps/issues/539), daps exited 0 nevertheless.

Fixed validation in d1ba31cf1d09e38102db284ebd607989530d3a93

jfehlig:
I thought there were gate checks to verify a PR could be built. Such trivial and common mistakes could be automatically detected well before merging.

lproven:
> I thought there were gate checks to verify a PR could be built. Such trivial and common mistakes could be automatically detected well before merging.

There are. Or, rather, there are meant to be. This highlighted a bug in DAPS which caused it to return the wrong exit code.

PR #443, opened 2019-06-10
Add a multi-device Btrfs section to AutoYaST handbook

joseivanlopez:
### Description

This PR adds information about how to configure a multi-device Btrfs drive in AutoYaST.

Relevant links:

* https://trello.com/c/FsEzGg70/975-3-multi-device-btrfs-support-in-autoyast
* https://jira.suse.de/browse/SLE-3877

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE15SP0

PR #619, opened 2020-06-30
Import @joseivanlopez' Multi-device Btrfs § from PR 443 (JIRA SLE-3877)

lproven:
### Description

This PR adds information about how to configure a multi-device Btrfs drive in AutoYaST.

(This is new functionality in SLE 15 SP2, which is why this change has been on hold since July 2019. It was in openSUSE long ago.)

Relevant links:

 * https://trello.com/c/FsEzGg70/975-3-multi-device-btrfs-support-in-autoyast
 * https://jira.suse.de/browse/SLE-3877

I have reformatted @joseivanlopez' text from https://github.com/SUSE/doc-sle/pull/443
Changes from PR #443:
 * inserted into the same position in the new, multi-file AutoYast guide
 * fixed indentation in the example
 * removed `raid5` and `raid6` as per @ancorgs's comment as they are not yet supported

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

### DIFF
--- a/xml/ay_partitioning.xml
+++ b/xml/ay_partitioning.xml
@@ -2566,7 +2566,7 @@ size=40G features='0' hwhandler='0' wp=rw
    </informaltable>
   </sect2>
 
-  <sect2 xml:id="ay.multidevice_btrfs">
+  <sect2 xml:id="ay-multidevice-btrfs">
    <title>Multi-device Btrfs Configuration</title>
 
    <para>
--- a/xml/ay_partitioning.xml
+++ b/xml/ay_partitioning.xml
@@ -2570,10 +2570,18 @@ size=40G features='0' hwhandler='0' wp=rw
    <title>Multi-device Btrfs Configuration</title>
 
    <para>
-    A <emphasis>Btrfs file system</emphasis> can be created on top of many
-    devices, offering similar features to a RAID device. With &ay;, a
-    <emphasis>multi-device Btrfs</emphasis> can be configured by specifying a
-    drive with the <literal>CT_BTRFS</literal> type. The <literal>device</literal>
+    Btrfs supports creating a single volume that spans more than one
+    storage device, offering similar features to software RAID implementations
+    such as the Linux kernel's built-in <command>mdraid</command> subsystem.
+    <emphasis>Multi-device Btrfs</emphasis> offers advantages over some other
+    RAID implementations. For example, you can dynamically migrate a
+    multi-device Btrfs volume from one RAID level to another, RAID levels can
+    be set on a per-file basis, and more. However, not all of these features are
+    fully supported yet in &productname; &product-ga; &product-sp;.
+   </para>
+    
+   <para>With &ay;, a multi-device Btrfs can be configured by specifying a drive
+    with the <literal>CT_BTRFS</literal> type. The <literal>device</literal>
     property is used as an arbitrary name to identify each multi-device Btrfs.
    </para>
 
--- a/xml/ay_partitioning.xml
+++ b/xml/ay_partitioning.xml
@@ -2646,7 +2646,7 @@ as per ancorgs' comment on Github-->
     system:
    </para>
 
-   <itemizedlist mark="bullet" spacing="normal">
+   <itemizedlist>
     <listitem>
      <para>
       Devices need to indicate the <literal>btrfs_name</literal> property to be
--- a/xml/ay_partitioning.xml
+++ b/xml/ay_partitioning.xml
@@ -2577,7 +2577,7 @@ size=40G features='0' hwhandler='0' wp=rw
     RAID implementations. For example, you can dynamically migrate a
     multi-device Btrfs volume from one RAID level to another, RAID levels can
     be set on a per-file basis, and more. However, not all of these features are
-    fully supported yet in &productname; &product-ga; &product-sp;.
+    fully supported yet in &productname; &product-ga; SP &product-sp;.
    </para>
     
    <para>With &ay;, a multi-device Btrfs can be configured by specifying a drive
--- a/xml/ay_partitioning.xml
+++ b/xml/ay_partitioning.xml
@@ -2566,6 +2566,104 @@ size=40G features='0' hwhandler='0' wp=rw
    </informaltable>
   </sect2>
 
+  <sect2 xml:id="ay.multidevice_btrfs">
+   <title>Multi-device Btrfs Configuration</title>
+
+   <para>
+    A <emphasis>Btrfs file system</emphasis> can be created on top of many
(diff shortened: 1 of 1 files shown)


xml/ay_partitioning.xml
```diff
@@ -2566,6 +2566,104 @@ size=40G features='0' hwhandler='0' wp=rw
...
+    to create over such block devices.
+   </para>
+
+   <para>
+    The following example shows a simple multi-device Btrfs configuration:
+   </para>
+
+   <example>
+    <title>Multi-device Btrfs configuration</title>
+      <screen>&lt;partitioning config:type="list"&gt;
```
tomschr, line 2592: Just to let you know: You don't have to mask the `&gt;` at the end, it can also be a literal `>`. Only the `<` needs to be masked as `&lt;` inside `<screen>`.

It may or may not help/please your eyes. :wink:

xml/ay_partitioning.xml
```diff
@@ -2566,6 +2566,104 @@ size=40G features='0' hwhandler='0' wp=rw
...
 
+  <sect2 xml:id="ay-multidevice-btrfs">
+   <title>Multi-device Btrfs Configuration</title>
+
+   <para>
+    A <emphasis>Btrfs file system</emphasis> can be created on top of many
+    devices, offering similar features to a RAID device. With &ay;, a
+    <emphasis>multi-device Btrfs</emphasis> can be configured by specifying a
+    drive with the <literal>CT_BTRFS</literal> type. The <literal>device</literal>
+    property is used as an arbitrary name to identify each multi-device Btrfs.
```
tomschr, line 2577: Why do I want that? What is the benefit of having a "multi-device Btrfs" over, let's say, a software RAID?
(Maybe it's beyond the scope of this little section to compare all possible (RAID)-technologies, but it would be nice to have at least one compelling argument for that.)

xml/ay_partitioning.xml
```diff
@@ -2566,6 +2566,104 @@ size=40G features='0' hwhandler='0' wp=rw
...
+   
+<!--lproven 2020-06-30: omitting raid5 & raid6 for now as unsupported in 15SP2
+as per ancorgs' comment on Github-->
+ 
+   <para>
+    Keep the following in mind when configuring a multi-device Btrfs file
+    system:
+   </para>
+
+   <itemizedlist mark="bullet" spacing="normal">
```
tomschr, line 2649: You don't need the attributes, because these are the defaults:

```suggestion
   <itemizedlist>
```

xml/ay_partitioning.xml
```diff
@@ -2566,6 +2566,104 @@ size=40G features='0' hwhandler='0' wp=rw
...
+    LVM logical volumes, etc.) and assign them to the Btrfs file system you want
+    to create over such block devices.
+   </para>
+
+   <para>
+    The following example shows a simple multi-device Btrfs configuration:
+   </para>
+
+   <example>
+    <title>Multi-device Btrfs configuration</title>
```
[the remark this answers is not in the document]
lproven: Why XML? This is part of the AutoYast docs -- I think it is a given that it's in AutoYast config file format, no?

xml/ay_partitioning.xml
lproven: I did not know that! Thank you. I think it's more elegant to match them, though, and maybe less prone to incorrectly remembering which one you need to mask and which one you don't...?

xml/ay_partitioning.xml
lproven: I defer to @joseivanlopez and @ancorgs on that. I will email them and ask.

xml/ay_partitioning.xml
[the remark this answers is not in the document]
tomschr: Isn't it XML?`Probably it's a given. But would it hurt to add this information? I think no. :-)
I don't think it's that important, decide what you like.

xml/ay_partitioning.xml
joseivanlopez: Good question. To be honest, I don't know the real benefits of having native Btrfs RAID over classical RAID. The only thing I would point to is that having RAID in the file system itself allows you to use everything that Btrfs offers (snapshots, quotas, etc) and also RAID behavior at the same time (without mixing several technologies). Moreover, Btrfs RAID is more "versatile" because (theoretically) it supports migration from one RAID level to another and RAID levels can also be specified per file.

In my opinion, all these features are not that mature yet. For example, booting from a Btrfs RAID could be tricky. See this link for some discussion about the topic: https://unix.stackexchange.com/questions/480391/btrfs-raid-1-vs-mdadm-raid-1.

Regarding the documentation, I am not sure whether to compare RAID vs Btrfs RAID is a good idea. Maybe only commenting something about the possibility of raid level migration and raid level per file would be enough. But take into account that these advanced features are not supported by YaST/AutoYaST yet, and I don't know the plans about it.

tomschr: > To be honest, I don't know the real benefits of having native Btrfs RAID over classical RAID.

To play devils advocate here: why then including this little snippet? :joy_cat: 

> [...] Regarding the documentation, I am not sure whether to compare RAID vs Btrfs RAID is a good idea.

IMHO this is not really needed here and would probably beyond the scope of this little section. However, you answered already the question:

> The only thing I would point to is that having RAID in the file system itself allows you to use everything that Btrfs offers (snapshots, quotas, etc) and also RAID behavior at the same time (without mixing several technologies).

I think a small sentence about this would be completely sufficient.

lproven: OK, I have added a short explanation:

>  Btrfs supports creating a single volume that spans more than one storage device, offering similar features to software RAID implementations such as the Linux kernel's built-in `mdraid` subsystem. *Multi-device Btrfs* offers advantages over some other RAID implementations. For example, you can dynamically migrate a multi-device Btrfs volume from one RAID level to another, RAID levels can be set on a per-file basis, and more. However, not all of these features are fully supported yet in SUSE Linux Enterprise Server 15 SP 2.

> With AutoYaST, a multi-device Btrfs can be configured by specifying a drive with the `CT_BTRFS` type. The `device` property is used as an arbitrary name to identify each multi-device Btrfs.

lproven:
I think this is good to go now. @ancorgs, @joseivanlopez, @tomschr -- any comments?

If not, @tomschr, would you be so kind as to merge it next week while I am away? Thank you!

tomschr:
@lproven Sure, no problem. :+1: I'll take care of that.

PR #621, opened 2020-07-13
Added a note about certificate location for RMT

tmuntaner:
### Description

Some customers would like to use a custom CA certificate to set up their RMT servers. With this note, we can help point them in the right direction without detailing a process for custom certificates.

### Checklist

* Check all items that apply.

*Are backports required?*

No


cjschroder:
Thank you @tmuntaner! Nice work.

tmuntaner:
Thanks @cjschroder

PR #624, opened 2020-07-24
Add SLED 15 SP2 translations

sknorr:
### Description
 Adds SLED 15 SP2 translations.

### Checklist
* Check all items that apply.

*Are backports required?*

- [x] no


sknorr:
We merged Frank's version of this instead. This can be deleted<---.

PR #627, opened 2020-07-27
vCPU scheduling (jsc#SLE-10167)

tbazant:
### Description
The granularity of vCPUs scheduling in Xen can be specified at boot time

### DIFF
--- a/xml/xen_config_options.xml
+++ b/xml/xen_config_options.xml
@@ -1052,14 +1052,15 @@ alice                             4     1     3   -b-       2.8 2-3</screen>
  <sect1 xml:id="sec-xen-config-scheduling">
   <title>Virtual CPU Scheduling</title>
   <para>
-   The &xen; hypervisor schedules virtual CPUs individually across physical CPUs.
-   With modern CPUs supporting multiple threads on each core, virtual CPUs can
-   run on the same core in different threads and thus influence each other.
-    The performance of a virtual CPU running in one thread can be noticeably affected
-    by what other virtual CPUs in other threads are doing. When these virtual
-   CPUs belong to different guest systems, these guests can influence each
-   other. The effect can vary, from variations in the guest CPU time accounting
-   to worse scenarios such as <emphasis>side channel attack</emphasis>.
+   The &xen; hypervisor schedules virtual CPUs individually across physical
+   CPUs.  With modern CPUs supporting multiple threads on each core, virtual
+   CPUs can run on the same core in different threads and thus influence each
+   other.  The performance of a virtual CPU running in one thread can be
+   noticeably affected by what other virtual CPUs in other threads are doing.
+   When these virtual CPUs belong to different guest systems, these guests can
+   influence each other. The effect can vary, from variations in the guest CPU
+   time accounting to worse scenarios such as <emphasis>side channel
+    attack</emphasis>.
   </para>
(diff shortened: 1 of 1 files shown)


jfehlig:
I'm don't have enough knowledge about this topic to comment. Dario Faggioli would be a good candidate for reviewing it.

xml/xen_config_options.xml
```diff
@@ -1049,4 +1049,58 @@ alice                             4     1     3   -b-       2.8 2-3</screen>
...
+    <term>core</term>
+    <listitem>
+     <para>
+      Virtual CPUs of a virtual core are always scheduled together on one
+      physical core. It never happens that two or more virtual CPUs from
+      different virtual cores are scheduled on the same physical core.
+      Therefore, some physical cores may have some of their CPUs left idle,
+      even if there are virtual CPUs wanting to run. This may slightly impact
+      performance, but it is important to achieve the improved isolation and
+      consistency.
```
dfaggioli, line 1092: About the last sentence. I think "slightly" might be a bit vague, or even incorrect, depending on the actual circumstances. In fact, the impact on the performance depends on the benchmark or workload. I'd propose something along the lines of:

"The impact on performance will depend on the actual workload being run inside of the guest systems. In most of the analyzed cases, the observed performance degradation, especially if under considerable load, was smaller than disabling HyperThreading (and hence leaving all the cores with just one thread)."

dfaggioli: And if we go for (something like) what I proposed, I'd say we could even add a mention/reference to the 'smt' boot parameter, i.e.: https://xenbits.xen.org/docs/unstable/misc/xen-command-line.html#smt-x86

But do we have it documented? I quickly checked, but have not seen it...

xml/xen_config_options.xml
```diff
@@ -1049,4 +1049,58 @@ alice                             4     1     3   -b-       2.8 2-3</screen>
...
   </sect2>
  </sect1>
+ <sect1 xml:id="sec-xen-config-scheduling">
+  <title>Virtual CPU Scheduling</title>
+  <para>
+   &xen; hypervisor schedules virtual CPUs individually across physical CPUs.
+   On modern CPUs that support multiple threads on one core, virtual CPUs can
+   run on the same core in different threads and thus influence each other.
+   Such virtual CPUs can observe performance variations resulting from the
+   actions of a virtual CPU running on the other thread.  When these virtual
```
lproven, line 1059:
```suggestion
    The performance of a virtual CPU running in one thread can be noticeably affected
    by what other virtual CPUs in other threads are doing. When these virtual
```

xml/xen_config_options.xml
```diff
@@ -1049,4 +1049,58 @@ alice                             4     1     3   -b-       2.8 2-3</screen>
...
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry>
+    <term>core</term>
+    <listitem>
+     <para>
+      Virtual CPUs of a virtual core are always scheduled together on one
+      physical core. It never happens that two or more virtual CPUs from
+      different virtual cores are scheduled on the same physical core.
```
lproven, line 1088:
```suggestion
      physical core. Two or more virtual CPUs from different virtual cores
      will never be scheduled on the same physical core.
```

PR #629, opened 2020-07-29
Firewall configuration via AutoYaST

mchf:
### Description
Describing new feature. Firewall configuration is going to be applied in first stage whenever possible.

### Checklist
* Check all items that apply.

*Are backports required?*

- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3

### DIFF
--- a/xml/ay_firewall.xml
+++ b/xml/ay_firewall.xml
@@ -356,38 +356,38 @@
       <itemizedlist>
           <listitem>
               <para>
-                  You are running &ay; with a &firewalld; section, and not 
-                  installing &slea; over SSH or VNC. The firewall is configured in the 
+                  You are running &ay; with a &firewalld; section, and not
+                  installing &slea; over SSH or VNC. The firewall is configured in the
                   first stage.
               </para>
           </listitem>
           <listitem>
               <para>
-                  You are running &ay; with a &firewalld; section, installing 
-                  &slea; over SSH or VNC, and no second stage is required. The 
+                  You are running &ay; with a &firewalld; section, installing
+                  &slea; over SSH or VNC, and no second stage is required. The
                   firewall is configured in the first stage.
               </para>
           </listitem>
           <listitem>
               <para>
-                  You are running &ay; with a &firewalld; section, installing 
-                  &slea; over SSH or VNC, and the second stage is required. The 
+                  You are running &ay; with a &firewalld; section, installing
+                  &slea; over SSH or VNC, and the second stage is required. The
                   firewall is configured in the second stage.
               </para>
-          </listitem>          
+          </listitem>
           <listitem>
               <para>
-                  You are running &ay; without a &firewalld; section. The 
-                  firewall is configured in the first stage according to the 
+                  You are running &ay; without a &firewalld; section. The
+                  firewall is configured in the first stage according to the
                   default product proposals.
               </para>
           </listitem>
           <listitem>
               <para>
-                  You are running &ay; with or without a firewall section, 
+                  You are running &ay; with or without a firewall section,
                   together with custom script which requires network access. The
-                  firewall is configured in the first stage either according to 
-                  the profile or the product proposals, and the firewall 
+                  firewall is configured in the first stage either according to
+                  the profile or the product proposals, and the firewall
                   configuration must be adapted so that the custom script has
                   network access as needed.
               </para>
--- a/xml/ay_firewall.xml
+++ b/xml/ay_firewall.xml
@@ -345,7 +345,6 @@
       <title>Installation Stages When the &firewalld; Profile is Applied</title>
 
       <para>
-          <!-- TODO replace product name and version with entity -->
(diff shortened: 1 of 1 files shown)


xml/ay_firewall.xml
```diff
@@ -340,6 +340,21 @@
...
 
+   <sect2 xml:id="CreateProfile-firewall-stage">
+      <title>When the Firewall Profile is Applied</title>
+
+      <para>
+       The profile is completely applied at the end of the first stage of installation, except when 
+       the installation runs over the network, for example over SSH or VNC. In this case we have 
+       to guarantee that the installed system stays accessible for the whole installation process. For 
+       that reason the firewall configuration is postponed to the second stage. If you need the network 
+       for some reason to fulfill installation successfully, you have to take care of proper firewall
```
cjschroder, line 351: hello @mchf, thank for you this PR! Very much appreciated. I have a question or two-- "you have to take care of proper firewall configuration" I am not sure what this means. Does this mean the admin should set up their firewall configuration after installation? Or does this mean the admin must be sure that their profile is correct?

xml/ay_firewall.xml
```diff
@@ -340,6 +340,21 @@
...
+
+      <para>
+       The profile is completely applied at the end of the first stage of installation, except when 
+       the installation runs over the network, for example over SSH or VNC. In this case we have 
+       to guarantee that the installed system stays accessible for the whole installation process. For 
+       that reason the firewall configuration is postponed to the second stage. If you need the network 
+       for some reason to fulfill installation successfully, you have to take care of proper firewall 
+       configuration. Last note is for cases when firewall configuration is not present in 
+       the profile at all. Firewall configuration is done according to the product settings as in common
+       installation in such case.
```
cjschroder, line 354: "Firewall configuration is done according to the product settings as in common       installation in such case." Does this mean that in the absence of a profile, there are default firewall settings that are applied?

thanks!

mchf: Not only in case of absence of a profile, but always when firewall section is missing.

xml/ay_firewall.xml
mchf: It means that if for some reason - e.g. in a  custom script - is network required to fulfill installation, then it is responsibility of user to configure firewall accordingly - we are not able to detect that as we do in case of ssh / vnc installation.

cjschroder:
@mchf all right, I think I get it now. Thank you for the fast reply! Please review my changes, and if they look correct I will merge. 

which SLE releases does this apply to? Currently, the master branch is SLE 15 SP2.

xml/ay_firewall.xml
```diff
@@ -341,17 +341,26 @@
...
-       the installation runs over the network, for example over SSH or VNC. In this case we have 
-       to guarantee that the installed system stays accessible for the whole installation process. For 
-       that reason the firewall configuration is postponed to the second stage. If you need the network 
-       for some reason to fulfill installation successfully, you have to take care of proper firewall 
-       configuration. Last note is for cases when firewall configuration is not present in 
-       the profile at all. Firewall configuration is done according to the product settings as in common
-       installation in such case.
+       The &firewalld; profile is applied at the end of the first 
+       stage of installation. This is an exception to the normal behavior of
+       system configuration occurring in the second stage (see
```
mchf, line 349: it is not exception. We're moving autoyast configuration into first stage for many (core) modules.   On the other hand, in case of firewall exception is when configuration is postponed to second stage - it happens only if running ssh / vnc installation with second stage enabled (which has to be done explicitly in the profile).

mchf: So what can happen is:
1) running AY with firewall section, no SSH / VNC -> firewall is configured at the end of 1st stage
2) running AY with firewall section over SSH / VNC and second stage is required -> firewall is configured in the second stage
3) running AY with firewall section over SSH / VNC and no second stage is required -> firewall is configured at the first stage
4) running AY without firewall section -> firewall is configured in the first stage according to product proposals

mchf:
It is documentation for changes done for sle-15-sp3 and is not planned to be (back)ported to sp2 and / or older products.

xml/ay_firewall.xml
mchf: and of course:
5. running AY with or without firewall section together with custom script which requires network access -> firewall is configured in the first stage either according to the profile or the product proposals but the firewall configuration has to be adapted to reflect the script needs (if needed)

cjschroder: Thanks for all your work @mchf. Sometimes getting the explanations right is the hardest part of documention. Let me know if this version is correct.

lproven (approved):

xml/ay_firewall.xml
mchf: sounds good. thanks for giving it right shape.

taroth21:
Thanks everybody. As we move forward with SP3-specific PRs, merging this now.

PR #631, opened 2020-07-30
Resolves bsc#1174687, but does not build

cjschroder:
Adding this to SLE12SP5 throws a big glob of validation errors, presumably because NovDoc

### Description
This chapter needs to be added to  SLE12SP5 & up. But it does not validate. I tried this and that, and need help.

### Checklist
* Check all items that apply.

*Are backports required?*

- [x] To maintenance/SLE15SP1
- [x] To maintenance/SLE15SP0
- [x] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3


xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!DOCTYPE chapter PUBLIC "-//Novell//DTD NovDoc XML V1.0//EN" "novdocx.dtd"
+[
+  <!ENTITY % NOVDOC.DEACTIVATE.IDREF "INCLUDE">
+  <!ENTITY % entities SYSTEM "entity-decl.ent">
+  %entities;
+]>
```
tomschr, line 7: This should be just a minimal DOCTYPE without the reference to Novdoc. Something like this:

```suggestion
<!DOCTYPE chapter 
[
  <!ENTITY % entities SYSTEM "entity-decl.ent">
  %entities;
]>
```

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+<chapter id="cha.deployment_firstboot">
```
tomschr, line 8: This looks like a DocBook 4 document, but we need DocBook 5. The latter contains the DB namespace, `xml:id` instead of `id` and more.

You need to convert this document into DocBook 5.

1. Preserve your entities first:

        $ sed -i -f /usr/share/daps/lib/entities.preserve.sed xml/deployment_firstboot.xml

2. Convert the DocBook 4 into DocBook 5:

        $ xsltproc --output deployment_firstboot.xml \
           /usr/share/daps/daps-xslt/migrate/suse-upgrade.xsl xml/deployment_firstboot.xml
        $ mv deployment_firstboot.xml xml/

3. Recover your entities:

        $ sed -i -f /usr/share/daps/lib/entities.recover.sed xml/deployment_firstboot.xml

4. Add the above minimal DOCTYPE in the header.

5. Try again. If it works, dance and be happy! :wink:

sknorr: Generally, Toms is right, the script should work. However, this is a relatively simple document and using the script won't be quicker than doing it manually. So in this case, I'd use this alternative:

1. I'd steal the header of another document (say, https://github.com/SUSE/doc-sle/blob/master/xml/apparmor_repositories.xml#L1,L7 -- just as Toms showed in his first comment but including the \<chapter\> tag as well because you need the xmlns:... definitions in your new document.
2. Edit the xml:id in the header -- should be the same as the id of the old document
3. Replace ` id=` -> ` xml:id=` across the document using search/replace.


There is a final step to get it to work with our CI that Toms has not mentioned and that is replacing the `.` characters in _all_ xml:id/linkend/arearefs attributes with `-`. Either do so manually or with this script https://github.com/sknorr/daps-migrate-tools/blob/master/id-dash-change-script
To use the script:

1. Clone the daps-migrate-tools repo
2. Go to your doc-sle repo
3. Run `../daps-migrate-tools/id-dash-change-script xml/deployment_firstboot.xml` without options (fix the path, needs to point at the right repo)

tomschr: Hmn... :thinking: 

> using the script won't be quicker than doing it manually

I'd friendly disagree. :-) I would be cautious to do it manually. Even for "simple" documents.

It would take Carla 1min to copy-and-paste the above commands. As a result, she gets a document which is _DocBook5-ready_. No need to worry about `id` -> `xml:id` changes, missed `ulink` to `link`, DocBook, XInclude, or XLink namespaces etc. That's all done by the stylesheet. So why should anyone do that manually when we already have a working procedure?

I agree with Stefan, two parts are missing: the DOCTYPE declaration and the ID values. Both aren't created automatically (yet). If it would help, I could add these features to the stylesheet too. As a parameter that can be switched on and off. I assume, this is something which would be helpful for future conversions.

So from my perspective, the only thing that Carla has to do after the above procedure is to call Stefan's script.

Maybe this is something we should add to the Hitchhiker's Guide, if not already the case. I haven't checked it yet, still recovering from yesterday.

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+  <step>
+   <para>
+    Prepare the master machine whose disk needs to be cloned to the client
+    machines. For more information, refer to
+    <xref linkend="sec.fb.master"/>.
+   </para>
+  </step>
+  <step>
+   <para>
+    Customize the firstboot workflow. For more information, refer to
```
lproven, line 34:
```suggestion
    Customize the Firstboot workflow. For more information, refer to
```

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+    Have the end user personalize the instance of &productname;. For more
+    information, refer to <xref linkend="sec.fb.personalize"/>.
+   </para>
+  </step>
+ </procedure>
+ <sect1 id="sec.fb.master">
+  <title>Preparing the Master Machine</title>
+
+  <para>
+   To prepare a master machine for a firstboot workflow, proceed as follows:
```
lproven, line 56:
```suggestion
   To prepare a master machine for a Firstboot workflow, proceed as follows:
```

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+   </step>
+   <step id="st.master2">
+    <para>
+     Boot the machine.
+    </para>
+   </step>
+   <step id="st.master3">
+    <para>
+     Perform a normal installation including all necessary configuration
+     steps and wait for the installed machine to boot. Also install the
```
lproven, line 73: "Also install"? Surely it must boot first, unless I misunderstand? In which case, I would suggest "Then, install..."

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+    <para>
+     To define your own workflow of &yast; configuration steps for the end
+     user or to add your own &yast; modules to this workflow, proceed to
+     <xref linkend="sec.fb.customize"/>. Otherwise proceed directly to
+     <xref linkend="st.master5"/>.
+    </para>
+   </step>
+   <step id="st.master5">
+    <para>
+     Enable firstboot as &rootuser;:
```
lproven, line 87:
```suggestion
     Enable Firstboot as &rootuser;:
```

lproven: OK, this uncapitalized usage is consistent. In which case ISTM that we have a choice to make. Either it is the _name_ of a technology or tool, in which case it should be capitalized; or, it's a command or other tool, in which case I feel it should be formatted as &lt;command&gt;firstboot&lt;/command&gt; or something like that. Either way, "firstboot" is not an English word.

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+    </para>
+    <para>
+     Create an empty file
+     <filename>/var/lib/YaST2/reconfig_system</filename> to trigger
+     firstboot's execution. This file will be deleted after the firstboot
```
lproven, line 92: ... also here, twice...

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+     Proceed to <xref linkend="sec.fb.clone"/>.
+    </para>
+   </step>
+  </procedure>
+ </sect1>
+ <sect1 id="sec.fb.customize">
+  <title>Customizing the Firstboot Installation</title>
+
+  <para>
+   Customizing the firstboot installation workflow may involve several
```
lproven, line 109: ... and here...

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+    <listitem>
+     <para>
+      Finish Setup
+     </para>
+    </listitem>
+   </itemizedlist>
+   <para>
+    But bear in mind that this workflow is just a template. You may adjust
+    it properly, editing manually the firstboot configuration file
+    <filename>/etc/YaST2/firstboot.xml</filename>. This XML file is a subset
```
lproven, line 400:
```suggestion
    Bear in mind that this workflow is just a template. You may adjust
    it properly, manually editing the firstboot configuration file
    <filename>/etc/YaST2/firstboot.xml</filename>. This XML file is a subset
```

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+    <step>
+     <para>
+      Apply your settings and close the configuration file.
+     </para>
+    </step>
+   </procedure>
+   <tip>
+    <title>Finding Connected Network Interface For Auto-Configuration</title>
+    <para>
+     If the target hardware may feature more than one network interface add
```
lproven, line 742:
```suggestion
     If the target hardware may feature more than one network interface, add
```

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+     the <systemitem>network-autoconfig</systemitem> package to the
+     application image. <systemitem>network-autoconfig</systemitem> makes
+     sure that during firstboot all available Ethernet interfaces are cycled
+     until one is successfully configured with DHCP.
```
lproven, line 746: Passive voice here. **What** is cycling through them? Yast, AutoYast? In which case, I would suggest:

"makes sure that during Fastboot, Yast cycles through all available Ethernet interfaces until..."

xml/deployment_firstboot.xml
```diff
@@ -0,0 +1,870 @@
...
+   </tip>
+  </sect2>
+ </sect1>
+ <sect1 id="sec.fb.clone">
+  <title>Cloning the Master Installation</title>
+
+  <para>
+   Clone the master machine's disk using any of the imaging mechanisms
+   available to you, and roll these images out to the target machines. For
+   more information about imaging see
```
lproven, line 851:
```suggestion
   more information about imaging, see
```

xml/deployment_firstboot.xml
sknorr: Additional suggestion for simplicity: "may feature" -> "may have"

xml/deployment_firstboot.xml
sknorr: Suggestion to not open this can of content worms here now -- this is not original writing, just a copy of an older document. Better merge it first and then try to get @\dgdavid to clarify.

xml/deployment_firstboot.xml
cjschroder: Leaving as-is. I searched the docs, and 'firstboot'  is mostly lower-case. Maybe it should be capitalized, but I don't want to deal with it right now.

xml/deployment_firstboot.xml
cjschroder: fixxored.

xml/deployment_firstboot.xml
cjschroder: agreed, but as I said in an earlier comment I don't want to deal with it right now.

xml/deployment_firstboot.xml
cjschroder: :+1:

xml/deployment_firstboot.xml
cjschroder: This document is editing :laughing: 
Fixed.

cjschroder:
Thanks @tomschr @sknorr @lproven much appreciate the help.

PR #632, opened 2020-08-02
Btrfs zstd

ddiss:
zstd documentation is missing from the Btrfs compression section.
It can be used with SLES 15sp1+.

### DIFF
--- a/xml/storage_filesystems.xml
+++ b/xml/storage_filesystems.xml
@@ -217,9 +217,9 @@
    <warning>
     <title>Support for Rollbacks</title>
     <para>
-     Rollbacks are only supported by the &suse; support if you do not remove
-     any of the preconfigured subvolumes. You may, however, add subvolumes
-     using the &yast; Partitioner.
+     Rollbacks are only supported by &suse; if you do not remove any of the
+     preconfigured subvolumes. You may, however, add subvolumes using the
+     &yast; Partitioner.
     </para>
    </warning>
    <sect3 xml:id="sec-filesystems-major-btrfs-compress">
--- a/xml/storage_filesystems.xml
+++ b/xml/storage_filesystems.xml
@@ -233,16 +233,22 @@
      </para>
     </note>
     <para>
-     Since &slea;12 SP1, compression for Btrfs file systems is supported. Use
-     the <option>compress</option> or <option>compress-force</option> option
-     and select the compression algorithm, <literal>lzo</literal> or
-     <literal>zlib</literal> (the default). The zlib compression has a higher
-     compression ratio while lzo is faster and takes less CPU load.
+     The Btrfs file system supports transparent compression. While enabled,
+     Btrfs compresses file data when written and uncompresses file data when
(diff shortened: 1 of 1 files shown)


ddiss:
Hmm, I just checked (with openSUSE 15.2) and found that zstd compression isn't supported by GRUB (it's supported upstream via http://git.savannah.gnu.org/cgit/grub.git/commit/?id=461f1d8af1820623dfc1e1e62c6f9f24c33793ba), so zstd will also need to be covered by the existing `Note: GRUB 2 and LZO Compressed Root` caveat.

ddiss:
Ping - any feedback here?

xml/storage_filesystems.xml
```diff
@@ -217,32 +217,39 @@
...
+     Btrfs compresses file data when written and uncompresses file data when
+     read.
+    </para>
+    <para>
+     Use the <option>compress</option> or <option>compress-force</option>
+     mount option and select the compression algorithm, <literal>zstd</literal>,
+     <literal>lzo</literal> or <literal>zlib</literal> (the default). zlib
+     compression has a higher compression ratio while lzo is faster and takes
+     less CPU load. The zstd algorithm offers a modern compromise, with
+     preformance close to lzo and compression ratios similar to zlib.
```
taroth21, line 247:
```suggestion
     performance close to lzo and compression ratios similar to zlib.
```

taroth21 (changes requested):
@ddiss : David, many thanks - just a typo fix from my side.

taroth21:
@ddiss : It's a bit hard to follow to which product version this fix applies... (PR description mentions 15SP1+ but then https://github.com/SUSE/doc-sle/pull/632/commits/bca52d703ae4c mentions it is not supported for 15SP2 out of the box), that's probably why Stefan flagged it with 'on hold: merge later' and 'SLE15SP3' only. 

Could you clarify please?

FYI: If the fix applies to SLE15SP3 only, it is on hold for now, we will merge the SP3-only changes into master when we start to update the product documentation for SP3 (~Nov. 2020).

xml/storage_filesystems.xml
ddiss: Fixed, thanks

ddiss:
Thanks for the feedback Tanja...

> @ddiss : It's a bit hard to follow to which product version this fix applies... (PR description mentions 15SP1+ but then bca52d7 mentions it is not supported for 15SP2 out of the box), that's probably why Stefan flagged it with 'on hold: merge later' and 'SLE15SP3' only.
> 
> Could you clarify please?

This documentation applies to SLES15-SP1 and later. The GRUB commit explicitly mentions 15SP2 because that's where I checked it; prior releases are expected to also lack bootloader zstd support.

taroth21 (approved):

taroth21:
@ddiss :+1: Merged and backported to the following branches:

- maintenance/SLE15SP2
- maintenance/SLE15SP1

PR #637, opened 2020-08-11
fix errors regarding mounting SMB1 shares

aaptel:
### Description

The SLE15+ docs have errors regarding mounting SMB1 shares.

no

- [ ] This PR only applies to SLE 15 SP3 or higher
- [x] This PR applies to older releases as well.

### Are backports required?

- [x] To maintenance/SLE15SP2
- [x] To maintenance/SLE15SP1
- [x] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP2

### DIFF
--- a/xml/net_samba.xml
+++ b/xml/net_samba.xml
@@ -765,7 +765,7 @@
    <para>
     However, this change only affects <command>mount</command> and
     mounting via <filename>/etc/fstab</filename>. SMB1 is still
-    available by explicitely requiring it when using the following:
+    available by explicitly requiring it. Use the following command:
    </para>
    <itemizedlist>
     <listitem>
--- a/xml/net_samba.xml
+++ b/xml/net_samba.xml
@@ -763,9 +763,9 @@
     by default, namely SMB 2.1, SMB 3.0, or SMB 3.02.
    </para>
    <para>
-    However, this change only affects <command>mount</command> and mounting via
-    <filename>/etc/fstab</filename>. SMB1 is still available by default when
-    using the following:
+    However, this change only affects <command>mount</command> and
+    mounting via <filename>/etc/fstab</filename>. SMB1 is still
+    available by explicitely requiring it when using the following:
    </para>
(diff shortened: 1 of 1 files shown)


pcacjr:
+1

aaptel:
is there anything preventing this to be merged?

xml/net_samba.xml
```diff
@@ -763,9 +763,9 @@
     by default, namely SMB 2.1, SMB 3.0, or SMB 3.02.
    </para>
    <para>
-    However, this change only affects <command>mount</command> and mounting via
-    <filename>/etc/fstab</filename>. SMB1 is still available by default when
-    using the following:
+    However, this change only affects <command>mount</command> and
+    mounting via <filename>/etc/fstab</filename>. SMB1 is still
+    available by explicitely requiring it when using the following:
```
sknorr, line 768: typo fix:
```suggestion
    available by explicitly requiring it when using the following:
```

sknorr:
Thank you!

sknorr:
@aaptel Since this bit me at home already, my personal notes say to add to `/etc/samba/smb.conf`:

```
    [global]
    [...]
            client min protocol = CORE
``

Is that an overall worse solution? Should that be documented as well?

aaptel:
We can document it but keep in mind it's insecure. It would be best to raise the max protocol of the server instead of lowerning the min of the client.

sknorr:
Thanks for the feedback! I added this now - cf. f074cd50f. Since there was a warning already, I left that part as-is.

(My original reason for using the global config file was that I wanted to connect to the share via Nautilus which does not really seem to offer any way to slip the `,vers=1.0` into the command line it uses. On server updates: in my personal use case, I can't (easily) update my router, though I am considering replacing it with a real NAS.)

PR #639, opened 2020-08-20
hardening guide: incorporate topics from SAP HANA hardening guide

mgerstner:
### Description

This adds two new sections to the SLE hardening guide that originate from the SAP HANA hardening guide. The topics aren't specific to SAP. A customer complained about lacking aspects in these two sections. The concerns have been addressed and the content updated and adjusted to the more general audience of the SLE hardening guide.

* bsc#1174501

No

- [ ] This PR only applies to SLE 15 SP3 or higher
- [X] This PR applies to older releases as well.

### Are backports required?

- [ ] To maintenance/SLE15SP2
- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [x] To maintenance/SLE12SP5
- [X] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP2

### DIFF
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -3554,7 +3554,7 @@ tty1</screen>
     <important>
      <para>
       These configuration changes will also cause root logins on serial
-      consoles like <filename>/dev/ttyS0</filename> to be denied. In case you
+      consoles such as <filename>/dev/ttyS0</filename> to be denied. In case you
       require such use cases you need to list the respective tty devices
       additionally in the <filename>/etc/securetty</filename> file.
      </para>
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -3450,22 +3450,24 @@ Account Expires: Never</screen>
 
    <para>
     By default, the &rootuser; user is assigned a password
-    and can login via various methods like on any local terminal, in a
-    graphical session, or remotely via SSH. These methods should be restricted
-    as far as possible. Shared usage of the root account should be avoided,
-    instead individual administrators should use tools like
-    <literal>su(1)</literal> or <literal>sudo(8)</literal> to elevate
-    privileges. This allows to associate &rootuser; logins with particular
-    users. This also adds another layer of security, because not only the root
-    password but also the password of an administrator's account needs to be
-    compromised to gain full root access.  This section explains how to limit
-    direct root logins on the different levels of the system.
+    and can log in using various methods&mdash;for example, on a local
+    terminal, in a graphical session, or remotely via SSH. These methods
+    should be restricted as far as possible. Shared usage of the root account
+    should be avoided. Instead, individual administrators should use tools such as
+    <literal>su</literal> or <literal>sudo</literal> (for more information,
+    type <command>man 1 su</command> or <command>man 8 sudo</command>) to obtain elevated
+    privileges. This allows associating &rootuser; logins with particular
+    users. This also adds another layer of security; not only the &rootuser;
+    password, but both the &rootuser; <emphasis>and</emphasis> the password of
+    an an administrator's regular account would need to be compromised to gain full root
+    access. This section explains how to limit direct root logins on the different
+    levels of the system.
    </para>
    <sect2 xml:id="sec-sec-prot-restrict-root-tty">
     <title>Restricting Local Text Console Logins</title>
 
     <para>
-     TTY devices provide text mode system access via the console. For desktop
+     TTY devices provide text-mode system access via the console. For desktop
      systems these are accessed via the local keyboard or&mdash;in case of server
      systems&mdash;via input devices connected to a KVM switch or a remote
      management card (ILO, DRAC, etc).
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -3570,7 +3570,7 @@ tty1</screen>
      not designed to be run as &rootuser; and are more likely to
      contain security issues than console programs. If you require a graphical
      login, use a non-&rootuser; login. Configure your system to disallow
-     &rootuser; login on graphical session.
+     &rootuser; from logging in to graphical sessions.
     </para>
 
     <para>
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -3574,16 +3574,15 @@ tty1</screen>
     </para>
 
     <para>
-     To deny the &rootuser; user login to graphical sessions you
+     To prevent &rootuser; from logging in to graphical sessions, you
      can apply the same basic steps as outlined in
      <xref linkend="sec-sec-prot-restrict-root-tty"/>.
-     You only need to also add the <literal>pam_securetty</literal> module to
-     the PAM stack file belonging to the display manager, for example
-     <filename>/etc/pam.d/gdm</filename> for GDM. The graphical session also
-     runs on a TTY device, <literal>tty7</literal>, by default. Therefore,
-     when you restrict &rootuser; logins to
-     <literal>tty1</literal>, then a &rootuser; login in the
-     graphical session will be denied.
+     Just add the <literal>pam_securetty</literal> module to the PAM
+     stack file belonging to the display manager&mdash;for example,
+     <filename>/etc/pam.d/gdm</filename> for GDM. The graphical
+     session also runs on a TTY device: by default, <literal>tty7</literal>.
+     Therefore, if you restrict &rootuser; logins to <literal>tty1</literal>,
+     then &rootuser; will be denied login in the graphical session.
     </para>
    </sect2>
 
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -3544,9 +3544,9 @@ tty1</screen>
 
     <important>
      <para>
-      Avoid adding the <literal>pam_securetty</literal> PAM module to the
-      <filename>/etc/pam.d/common-auth</filename> file. This change would
-      break <command>su</command> and <command>sudo</command>, however,
+      Do not add the <literal>pam_securetty</literal> module to the
+      <filename>/etc/pam.d/common-auth</filename> file. This would
+      break the <command>su</command> and <command>sudo</command> commands,
       because these tools would then also reject &rootuser; authentications.
      </para>
     </important>
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -3534,7 +3534,7 @@ tty1</screen>
       <para>
        Check whether logins to other terminals will be rejected for
        &rootuser;. A login on <literal>tty2</literal>, for example, should be
-       rejected right away without even the root password being queried.
+       rejected immediately, without even querying the account password.
        Also make sure that you can still successfully login to
        <literal>tty1</literal> and thus that &rootuser; is not locked out of
        the system completely.
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -3491,7 +3491,7 @@ Account Expires: Never</screen>
     <note>
      <para>
       The steps shown here are tailored towards PC architectures (&x86; and
-      &x86-64;). On architectures like &power; different terminal
+      &x86-64;). On architectures such as &power;, different terminal
       device names than <literal>tty1</literal> may be used. Be careful not to
       lock yourself out completely by specifying wrong terminal device names.
       You can determine the device name of the terminal you are currently
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -3620,7 +3620,7 @@ tty1</screen>
    <title>Setting an Inactivity Timeout for Interactive Shell Sessions</title>
    <para>
     It can be a good idea to terminate an interactive shell session after a
-    certain time of inactivity, for example to:
+    certain period of inactivity; for example, to:
    </para>
     <itemizedlist>
      <listitem>
@@ -3636,7 +3636,7 @@ tty1</screen>
     </itemizedlist>
    <para>
     By default, there is no inactivity timeout for shells. Nothing will
-    happen if a shell stays open unused for days or even years. It is possible
+    happen if a shell stays open and unused for days or even years. It is possible
     to configure most shells in a way that idle sessions terminate
     automatically after a certain amount of time. The following example
     shows how to set an inactivity timeout for a number of
@@ -3646,8 +3646,8 @@ tty1</screen>
     The inactivity timeout can be configured for login shells only or for all
     interactive shells. In the latter case the inactivity timeout is running
     individually for each shell instance. This means that timeouts will
-    accumulate. When a sub or child shell is started then the timeout runs for
-    the sub or child shell and only afterwards will the
+    accumulate. When a sub- or child-shell is started, then a new timeout begins for
+    the sub- or child-shell, and only afterwards will the
     timeout of the parent continue running.
    </para>
    <para>
@@ -3716,29 +3716,28 @@ tty1</screen>
    </informaltable>
    <para>
     Every listed shell supports an internal timeout shell variable that can be
-    set to a specific time value to activate the inactivity timeout. If you want
-    to prevent users from overriding the timeout setting you can mark the
+    set to a specific time value to cause the inactivity timeout. If you want
+    to prevent users from overriding the timeout setting, you can mark the
     corresponding shell timeout variable as readonly. The corresponding variable
     declaration syntax is also found in the table above.
    </para>
 
    <note>
     <para>
-     This feature is only a utility that can help avoiding risks if a user is
-     forgetful or follows unsafe practices. It does not protect against
-     malicious users, however. The timeout discussed here only applies to
-     certain interactive wait states of a shell. A malicious user can always
-     find ways to circumvent the timeout and keep their session open anyway.
+     This feature is only helpful for avoiding risks if a user is forgetful or follows
+     unsafe practices. It does not protect against hostile users. The timeout
+     only applies to interactive wait states of a shell. A malicious user can always
+     find ways to circumvent the timeout and keep their session open regardless.
     </para>
    </note>
 
    <para>
-    To configure the inactivity timeout you need to add the matching timeout
-    variable declaration to a shell startup script. Use either the path for
-    login shells only, or the one for all shells as listed in the table. The
+    To configure the inactivity timeout, you need to add the matching timeout
+    variable declaration to each shell's startup script. Use either the path for
+    login shells only, or the one for all shells, as listed in the table. The
     following example uses paths and settings that are suitable for
     <command>bash</command> and <command>ksh</command>
-    to setup a readonly login shell timeout that cannot be overriden by users.
+    to setup a read-only login shell timeout that cannot be overridden by users.
     Create the file <filename>/etc/profile.d/timeout.sh</filename> with the
     following content:
    </para>
@@ -3751,10 +3750,10 @@ readonly TMOUT=86400</screen>
 
    <tip>
     <para>
-     We recommend to use the <command>screen</command> tool in order to
-     detach sessions before logging out. Screen sessions are not terminated
-     and can be re-attached whenever it is required. Active session can be
-     locked without logging out (read about
+     We recommend using the <command>screen</command> tool in order to
+     detach sessions before logging out. <command>screen</command> sessions
+     are not terminated and can be re-attached whenever required. An active
+     session can be locked without logging out (read about
      <keycombo action="seq">
      <keycombo action="press"><keycap function="control"/><keycap>a</keycap></keycombo>
      <keycap>x</keycap>
--- a/xml/hardening.xml
+++ b/xml/hardening.xml
@@ -65,7 +65,7 @@
   policies consist of the following concepts (fairly generic and incomplete
   list):
  </para>
- <itemizedlist mark="bullet" spacing="normal">
+ <itemizedlist>
   <listitem>
    <para>
     DAC (Discretionary Access Control): File and directory permissions, as set
@@ -198,7 +198,7 @@
    information is available here:
   </para>
 
-  <itemizedlist mark="bullet" spacing="normal">
+  <itemizedlist>
    <listitem>
     <para>
      <link xlink:href="https://www.netiq.com/products/privileged-user-manager/"/>
@@ -386,7 +386,7 @@
     through 7:
    </para>
 
-   <itemizedlist mark="bullet" spacing="normal">
+   <itemizedlist>
     <listitem>
      <para>
       EAL1: Functionally tested
@@ -689,7 +689,7 @@
    find more details and explanations in this chapter. Selected general topics
    are:
   </para>
-  <itemizedlist mark="bullet" spacing="normal">
+  <itemizedlist>
    <listitem>
     <para>
      Physical Security – Protection of the server from environmental threats
@@ -783,7 +783,7 @@
     Additionally, consider questions like:
    </para>
 
-   <itemizedlist mark="bullet" spacing="normal">
+   <itemizedlist>
     <listitem>
      <para>
       Who has direct physical access to the host?
@@ -1565,7 +1565,7 @@
     file system hierarchy. A number of interesting mount options are:
    </para>
 
-   <itemizedlist mark="bullet" spacing="normal">
+   <itemizedlist>
     <listitem>
      <para>
       <literal>noexec</literal>: prevents execution of files.
@@ -1826,7 +1826,7 @@
      <emphasis>No-eXecute</emphasis> (NX bit) for mitigation of buffer
      overflow attacks. For more information, see:
     </para>
-    <itemizedlist mark="bullet" spacing="normal">
+    <itemizedlist>
      <listitem>
       <para>
        <link xlink:href="http://searchenterpriselinux.techtarget.com/tip/Linux-virtual-address-randomization-and-impacting-buffer-overflows"/>
@@ -1965,7 +1965,7 @@
     &aa; consists of:
    </para>
 
-   <itemizedlist mark="bullet" spacing="normal">
+   <itemizedlist>
     <listitem>
      <para>
       A kernel extension which enforces the security profiles.
@@ -1975,7 +1975,7 @@
      <para>
       A collection of RPMs, also shipped with &productname; that provides:
      </para>
-     <itemizedlist mark="bullet" spacing="normal">
+     <itemizedlist>
       <listitem>
        <para>
         A set of &aa; profiles for numerous programs that ship with
@@ -2326,7 +2326,7 @@
     <para>
      In a nutshell, the &rmtool; for &sle; provides customers with:
     </para>
-    <itemizedlist mark="bullet" spacing="normal">
+    <itemizedlist>
      <listitem>
       <para>
        Assurance of firewall and regulatory compliance
@@ -2606,7 +2606,7 @@ telnet &lt;remote_host&gt; 25</screen>
     Here are some basic rules:
    </para>
 
-   <itemizedlist mark="bullet" spacing="normal">
+   <itemizedlist>
     <listitem>
      <para>
       NFS should not be enabled if not needed.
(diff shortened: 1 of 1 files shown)


xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+# in case of inactivity.
+# 24h = 86400 sec
+readonly TMOUT=86400</screen>
+
+   <tip>
+    <para>
+     We recommend to use the <literal>screen</literal> tool in order to
+     detach sessions before logging out. Screen sessions are not terminated
+     and can be re-attached whenever it is required. Active session can be
+     locked without logging out (read about <literal>C-a x</literal> /
```
tomschr, line 3737:
```suggestion
     locked without logging out (read about <keycombo action="seq">
      <keycombo action="press">
       <keycap function="control"/>
      <keycap>a</keycap>
      </keycombo>
      <keycap>x</keycap>
     </keycombo> /
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
    </para>
   </sect1>
   <xi:include href="hardening_pam_stack.xml"/>
+  <sect1 xml:id="sec-sec-prot-restrict-root">
+   <title>Restricting <literal>root</literal> logins</title>
+
+   <para>
+    By default the <literal>root</literal> user is assigned a password
+    and can login on any local terminal, in a graphical session or via SSH.
+    For security sensitive systems <literal>root</literal> logins should be
```
tomschr, line 3454:
```suggestion
    For security sensitive systems &rootuser; logins should be
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+    restricted to as few methods as possible. Shared usage of the root account
+    should be avoided, instead individual administrators should use tools like
+    <literal>su(1)</literal> or <literal>sudo(8)</literal> to elevate privileges.
+    This allows to associate <literal>root</literal> logins with particular users.
```
tomschr, line 3458:
```suggestion
    This allows to associate &rootuser; logins with particular users.
```

tomschr:
Thank you @mgerstner. You're feedback is very welcome.  :+1: You will get feedback soon, I'm almost finished.

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+    <title>Restricting Local Text Console Logins</title>
+
+    <para>
+     TTY devices provide text mode system access via the console. For desktop
+     systems these are accessed via the local keyboard or in case of server
+     systems via input devices connected to a KVM switch or a remote
+     management card (ILO, DRAC, etc).
+     By default, Linux offers 6 different consoles, that can be switched to
+     via the key-combinations Alt+F1 - Alt+F6, when running in text mode, or
+     Ctrl+Alt+F1 - Ctrl+Alt+F6 when running in a graphical session. The
```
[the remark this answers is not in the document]
mgerstner: and there I thought Docbook couldn't become any more bloated ;-)

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+     <literal>tty6</literal> accordingly.
+    </para>
+
+    <para>
+     The following steps restrict root access to the first TTY. Even this
+     access method is only meant for emergency access to the system and should
+     never be used for everyday system administration tasks.
+    </para>
+
+    <itemizedlist mark="bullet" spacing="normal">
```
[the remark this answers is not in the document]
mgerstner: These defaults are already used throughout the document. Should I remove them all then?

mgerstner:
I have rebased the two existing commits to adjust the style according to your comments. Regarding the `itemizedlist` defaults I've removed them from the whole document, to make it consistent, in a separate commit.

xml/hardening.xml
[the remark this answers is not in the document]
tomschr: Haha, I know, I know. :smile: We need it for consistency reasons. The above markup allows us to render them appropriately (gray background, borders etc). Without it, that rendering won't be applied to pure text.

xml/hardening.xml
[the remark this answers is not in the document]
tomschr: Yes please! :+1:

tomschr:
@mgerstner Thank you Matthias! Perfect! :+1:

If you don't object or need more time, I will do the second round and look into text issues.

mgerstner:
> If you don't object or need more time, I will do the second round and look into text issues.

Of course, just go ahead.

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+     </para>
+     <screen>auth     requisite      pam_nologin.so
+auth     [user_unknown=ignore success=ok ignore=ignore auth_err=die default=bad] pam_securetty.so noconsole
+auth     include        common-auth</screen>
+     </listitem>
+     <listitem>
+      <para>
+       This will include the <literal>pam_securetty</literal> PAM module during the
+       authentication process on local consoles to allow root only to login on
+       tty devices that are listed in the file <filename>/etc/securetty</filename>.
```
tomschr, line 3503: This is an explanation, not an action. As such, I think this belongs to the first step. I'd suggest to move that as a para into the first step (after the `<screen>` as an additional `<para>`).

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+       Thus to actually limit the tty devices for root you need to remove all
+       entries from this file except one:
```
tomschr, line 3505: Try to start with a verb:

```suggestion
       Remove all entries from this file except one. This limits the access to TTY devices for root.
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+     <screen>#
+# This file contains the device names of tty lines (one per line,
+# without leading /dev/) on which root is allowed to login.
+#
+tty1</screen>
+     </listitem>
+     <listitem>
+      <para>
+       After these changes root logins from non-allowed terminals will be
+       rejected right away without even querying the root password.
```
tomschr, line 3516: I'm not sure about this. What should the user do here? It sounds like it's pure information? If yes, then it should be added after the action. Maybe add it to the last step after the `<screen>`?. 

I would create a different action here. Something like "Check bla..." to see if everything is okay. Does that make sense?

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+      </para>
+     </listitem>
+    </itemizedlist>
+
+    <important>
+     <para>
+      One might be tempted to add the <literal>pam_securetty</literal> PAM
+      module to the <filename>/etc/pam.d/common-auth</filename> file to make it
+      effective for all PAM services. This would break <literal>su</literal>
+      and <literal>sudo</literal>, however, because these tools would then
```
tomschr, line 3526: Maybe simplify it?

```suggestion
      Avoid adding the  <literal>pam_securetty</literal> PAM  module to the
      <filename>/etc/pam.d/common-auth</filename> file. This change would
      break the <command>su</command> and <command>sudo</command>
      commands as the reject &rootuser; authentication.
```

Just an idea. :wink:

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+      also reject root authentications.
+     </para>
+    </important>
+
+    <important>
```
tomschr, line 3531: If possible, try to avoid three consecutive admonitions (2x important and note). This kind of "dilutes" the warning. Could you combine the two important into one?

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+     <para>
+      These configuration changes will also cause root logins on serial
+      consoles like <filename>/dev/ttyS0</filename> to be denied. In case you
+      require such use cases you need to list the respective tty devices
+      additionally in the <filename>/etc/securetty</filename> file.
+     </para>
+    </important>
+
+    <note>
```
tomschr, line 3540: I think, this note is essential for the procedure above. Maybe move that *before* the procedure?

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+   <sect2 xml:id="sec-sec-prot-retrict-root-graphical">
+    <title>Restricting Graphical Session Logins</title>
+
+    <para>
+     On security sensitive systems like servers it is best not to run a
+     graphical user environment at all. If you require a graphical login
+     anyways then it is even more so important that the &rootuser;
+     user cannot login in a graphical session. Graphical programs are often
+     not designed to be run as &rootuser; and are more likely to
+     contain security issues than console programs.
```
tomschr, line 3565: "On security sensitive systems..." same here. I'd took the liberty here to rephrase it a bit more.

```suggestion
    To improve security on your server, avoid to use
    graphical user environments at all.
    Graphical programs are often not designed to be run as
    &rootuser; and are more likely to contain security issues than
    console programs.
    If you require a graphical login, use a non-&rootuser; login.
    Configure your system to disallow &rootuser; login on graphical
    sessions.
```

I've structured it as follows: what is the goal ("to improve security"), then an explanation (the reason), and after that how to avoid it.

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+
+   <sect2 xml:id="sec-sec-prot-restrict-root-ssh">
+    <title>Restricting SSH Logins</title>
+    <para>
+     By default the &rootuser; user is also allowed to login
+     into a machine remotely via the SSH network protocol (if the SSH port is
+     not blocked by the firewall). To restrict this possibility the following
+     change of the OpenSSH configuration needs to be performed:
+    </para>
+    <itemizedlist>
```
tomschr, line 3590: Same here, also use a procedure/step structure.

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+      <screen>systemctl restart sshd.service</screen>
+     </listitem>
+    </itemizedlist>
+    <note>
+     <para>
+      Using the PAM module <literal>pam_securetty</literal> is not suitable in
+      case of OpenSSH, because SSH logins not necessarily go through the PAM
+      stack during the auth phase (e.g. when using SSH public key authentication).
+      Also an attacker could be able to differentiate between a wrong password
+      and a successful login that was only rejected later on by policy.
```
tomschr, line 3606: This is not an action, but information. Is there anything the user has to do?

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+  </sect1>
+  <sect1 xml:id="sec-sec-prot-inactivity-logout">
+   <title>Setting an Inactivity Timeout for Interactive Shell Sessions</title>
+   <para>
+    It can be a good idea to terminate an interactive shell session after a
+    certain time of inactivity. For example an administrator might have
+    forgotten to terminate a local login session on a sensitive system. This
+    could provide an attack vector if a local attacker finds such a session
+    unguarded. Another effect of idle shell sessions is waste of system
+    resources.
```
tomschr, line 3620: Maybe use a list to present the reasons why it is a good idea to terminate an interactive shell session. It could look like this:

```
It can be a good idea to terminate an interactive shell session after a certain time
of inactivity, for example:

* Prevents open, unguarded sessions on sensitive systems.
* Avoids waste of system resources.
```

Maybe it has to be phrased a bit better, but I hope you get the idea. :wink:

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+   </para>
+   <para>
+    By default there is no inactivity timeout for shells, so nothing will
```
tomschr, line 3623:
```suggestion
    By default, there is no inactivity timeout for shells. Nothing will
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+    happen if a shell stays open unused for days or even years. It is possible
+    to configure most shells in a way that idle sessions terminate
+    automatically after a certain amount of time. In the following example we
+    will show how to set an inactivity timeout for a number of
```
tomschr, line 3627: To avoid "we" and to make it consistent with the other text:

```suggestion
    automatically after a certain amount of time. The following example
    shows how to set an inactivity timeout for a number of
```

tomschr (changes requested):
@mgerstner Thanks Matthias for all your work. Much appreciated! :+1: 

I've went through your text and added some comments. Hope I didn't misread your text. In case you have any questions, just let me know.

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
   </sect1>
   <xi:include href="hardening_pam_stack.xml"/>
+  <sect1 xml:id="sec-sec-prot-restrict-root">
+   <title>Restricting &rootuser; logins</title>
+
+   <para>
+    By default the &rootuser; user is assigned a password
+    and can login on any local terminal, in a graphical session or via SSH.
+    For security sensitive systems &rootuser; logins should be
+    restricted to as few methods as possible. Shared usage of the root account
```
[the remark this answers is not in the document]
mgerstner: > "For security sensitive systems"... that sounds like there are systems which don't need security. I think, this gives the wrong impression. We should tell our customers that security is _always_ important.

I see what you're getting at. But there is a difference between, say, a workstation in a restricted office, a server in a data center and a laptop used in the field. It could also raise the question why these hardenings need to be applied manually then and aren't already setup by default ;-). In this light I think it is legit to notice that these hardenings are not for everybody but for a specific target audience that worry especially about security or have especially sensitive setups. But I'm still okay to drop this part.

> Another issue is, I'm not sure what you mean by "methods". Accounts? As such, I'd suggest to rephrase it like:

Actually, no. The methods refer to the "local terminal", "graphical session" and "SSH", different ways to login as root on a system.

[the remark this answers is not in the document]
tomschr: Ahh, I see what you mean. 🙂 Is that a common term? I heard it before. It could make sense to little bit elaborate on that. It doesn't need much, but maybe an additional sentence would be fine. 😉

Oh, then I misread your sentence, sorry. Well, in that case would it make sense to add an additional hint? Something like this:

    logins should be restricted to as few methods (like local terminal, SSH, graphical sessions, etc.) 
    as possible.

xml/hardening.xml
mgerstner: Makes sense. I'll change it into a verification step.

xml/hardening.xml
mgerstner: fine with me, I'll move if up

xml/hardening.xml
```diff
@@ -3445,6 +3445,311 @@ Account Expires: Never</screen>
...
+      associated terminal devices are named <literal>tty1</literal> .. <literal>tty6</literal> accordingly.
+    </para>
+
+    <para>
+     The following steps restrict root access to the first TTY. Even this
+     access method is only meant for emergency access to the system and should
+     never be used for everyday system administration tasks.
+    </para>
+
+    <itemizedlist>
```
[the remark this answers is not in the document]
mgerstner: A procedure it will be. I think it works out well.

xml/hardening.xml
mgerstner: A lot of admonitions may seem strange but these are facts. Combining the two important blocks is technically possible, of course, but they are still two distinct items. So somebody might read the first few sentences and think "doesn't affect me" and skip over the rest of the block.

xml/hardening.xml
mgerstner: No, that's why this is a `<note>` ;)

xml/hardening.xml
tomschr: I was more speaking from a documentation and readability perspective. If you think this is too hard then leave it as it is for the time being. :-)

xml/hardening.xml
mgerstner: Okay. But ... aren't all systems sensitive? :-P

tomschr: Hehe, that's an almost philosophical question. ;-) It probably depends on how you define "sensitive". ;-)

tomschr:
@mgerstner Just one question Matthias: I'd like to finish this PR by end of this week. Do you think, this is feasible?

xml/hardening.xml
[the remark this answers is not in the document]
mgerstner: I think that would be strangely redundant. I try a bit of a different approach to make it more clear.

You mean if "login methods" is a common term? Well Google knows a couple of million of occurrences ;-). Apart from that I don't think the term "method" is uncommon in technical contexts.

mgerstner:
> @mgerstner Just one question Matthias: I'd like to finish this PR by end of this week. Do you think, this is feasible?

I will push changes in a moment. The rest depends upon how quickly we can agree on the changes.

xml/hardening.xml
tomschr: Hmn... if *all* systems are sensitive, then what's the meaning of insensitive systems? :wink:  To me it raises more questions than that the term solves. Either define what you mean by "sensitive systems" or remove it.

mgerstner:
I've added two new commits that should address most of the changes you requested, except the few that I disputed. Please have a look and mark the things as resolved that you are happy with.

Before you merge anything please let me rebase these commits to achieve a cleaner git history.

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+      </para>
+     </listitem>
+    </itemizedlist>
+
+    <important>
+     <para>
+      One might be tempted to add the <literal>pam_securetty</literal> PAM
+      module to the <literal>/etc/pam.d/common-auth</literal> file to make it
+      effective for all PAM services. This would break <literal>su</literal>
+      and <literal>sudo</literal>, however, because these tools would then
```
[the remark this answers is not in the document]
tomschr: Did you miss this? Or do you disagree with my change? :wink: Can't find it in your file.

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+    <note>
+     <para>
+      The steps shown here are tailored towards PC architectures (&x86; and
+      &x86-64;). On architectures like &power; different terminal
+      device names than <literal>tty1</literal> may be used. Be careful not to
+      lock yourself out completely by specifying wrong terminal device names.
+      You can determine the device name of the terminal you are currently
+      logged into by running the <literal>tty</literal> command. Be careful
+      not to do this in a virtual terminal like via SSH or in a graphical
+      session (device names <literal>/dev/pts/&lt;N&gt;</literal>) but only
```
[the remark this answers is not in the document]
tomschr: Same here? :wink:

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+      from an actual login terminal reachable via Alt+F&lt;N&gt;.
```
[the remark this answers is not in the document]
tomschr: This too. :wink:

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+     not blocked by the firewall). To restrict this possibility the following
+     change of the OpenSSH configuration needs to be performed:
+    </para>
+    <itemizedlist mark="bullet" spacing="normal">
+     <listitem>
+      <para>Edit <literal>/etc/ssh/sshd_config</literal> and adjust the following parameter:</para>
+      <screen>PermitRootLogin no</screen>
+     </listitem>
+     <listitem>
+      <para>Restart the <literal>sshd</literal> service to make the changes effective:</para>
```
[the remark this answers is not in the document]
tomschr: This is also missing in the file. Or do you disagree? :wink:

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+     <colspec colname="col_7" colwidth="14.2858*"/>
+     <thead>
+      <row>
+       <entry align="left" valign="top">package</entry>
+       <entry align="left" valign="top">shell personalities</entry>
+       <entry align="center" valign="top">shell variable</entry>
+       <entry align="center" valign="top">time unit</entry>
+       <entry align="left" valign="top">readonly setting</entry>
+       <entry align="center" valign="top">config path (only login shell)</entry>
+       <entry align="center" valign="top">config path (all shells)</entry>
```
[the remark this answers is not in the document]
tomschr: I can't see this change as well.

xml/hardening.xml
```diff
@@ -3445,6 +3445,301 @@ Account Expires: Never</screen>
...
+       <entry align="left" valign="top"><para><literal>/etc/csh.cshrc.local</literal></para></entry>
+      </row>
+      <row>
+       <entry align="left" valign="top"><para><literal>zsh</literal></para></entry>
+       <entry align="left" valign="top"><para><literal>zsh</literal></para></entry>
+       <entry align="center" valign="top"><para><literal>TMOUT</literal></para></entry>
+       <entry align="center" valign="top"><para>seconds</para></entry>
+       <entry align="left" valign="top"><para><literal>readonly TMOUT=</literal></para></entry>
+       <entry align="left" valign="top"><para><literal>/etc/profile.local, /etc/profile.d/</literal></para></entry>
+       <entry align="left" valign="top"><para><literal>/etc/zsh.zshrc.local</literal></para></entry>
```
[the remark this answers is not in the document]
tomschr: This too.

tomschr:
@mgerstner Thanks Matthias for all your efforts! Much appreciated! :+1: 

I've went through all the changes and marked them as resolved. The suggestions that are left are mostly formal issues (attributes etc). To my understanding, this shouldn't be really controversial. :wink:  Could you have a look?

Not sure if you know that or not. Sorry, I didn't want to educate you. Probably you all know that already.
Just would like to make you aware of one thing in GitHub. You can [incorporate suggestions quite easily](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request) without leaving GitHub. One benefit is they automatically mark the change as "resolved". See the link for more details. :wink:

xml/hardening.xml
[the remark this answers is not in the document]
mgerstner: Seems I missed this. GitHub is not very helpful recently e.g. by "hiding comments" by default.

xml/hardening.xml
tomschr: Ok, fine with me. :-)

xml/hardening.xml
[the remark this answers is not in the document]
mgerstner: I actually adressed this one, but missed the `<replaceable>` part.

xml/hardening.xml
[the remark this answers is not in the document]
tomschr: Hmn, yes. Ok, I'll leave that to Liam, maybe he has some idea.

xml/hardening.xml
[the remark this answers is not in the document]
mgerstner: Maybe GitHub didn't send me all your comments as email, because I worked on each mail I got in my inbox. Sorry for the confusion.

xml/hardening.xml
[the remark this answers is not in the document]
tomschr: No problem, don't worry. :-) Yes, it happens sometimes to me too.

xml/hardening.xml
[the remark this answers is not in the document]
tomschr: All good. Don't worry. :-)

mgerstner:
> Not sure if you know that or not. Sorry, I didn't want to educate you. Probably you all know that already.
> Just would like to make you aware of one thing in GitHub.

Yeah I figured that. Maybe I try that next time. I'd need to add the changes to batch commit. But I also want to keep the changes for the sections separate so I can rebase easily in the end. I can do that more easily on the command line

mgerstner:
So the missing stuff should be covered now. When you're fine with the current state then please tell me so I can rebase and squash the two additional commits before merging.

tomschr:
> Yeah I figured that. Maybe I try that next time. I'd need to add the changes to batch commit.

Sure, whatever works best for you. It wasn't my intention to force you to some workflow. :-)

> So the missing stuff should be covered now. When you're fine with the current state then please tell me so I can rebase and squash the two additional commits before merging.

Will do! Thanks! :+1:
I think, it would be a good idea to squash the commits before we hand-over to Liam for proofreading.

xml/hardening.xml
```diff
@@ -3445,6 +3445,325 @@ Account Expires: Never</screen>
    </para>
   </sect1>
   <xi:include href="hardening_pam_stack.xml"/>
+  <sect1 xml:id="sec-sec-prot-restrict-root">
+   <title>Restricting &rootuser; logins</title>
```
tomschr, line 3449: Just one small correction:

```suggestion
   <title>Restricting &rootuser; Logins</title>
```

tomschr:
@mgerstner: There is a small change of "login" -> "Login" above. Should be fairly easy. :-) I think, when this is integrated, I'm fine with the current state. :+1:

Just squash all the commits, add a reference to the BSC in the commit message, and we can give it to Liam.

Thanks Matthias for your wonderful work!

xml/hardening.xml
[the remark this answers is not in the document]
lproven: I think that "login method" is a valid phrase, but ambiguous. I would just suggest slightly clarifying the list of examples:

> logins should be restricted to as few methods (such as on the local console, via SSH, graphical sessions, etc.) 

I think with the examples, it's clear.

mgerstner:
I've squashed the commits. So from my side this is ready to be merged. Expect I still overlooked something.

tomschr:
@mgerstner Thanks! I see 3 commits. May I ask to squash them into one? :wink:

mgerstner:
> @mgerstner Thanks! I see 3 commits. May I ask to squash them into one?

IMO the exercise is not to end up with as few commits as possible but to end up with well structured commits. Now we have one that harmonizes the existing hardening guide regarding itemizedlist defaults. Then one that adds the section about restricting root logins. And one that adds the section about shell inactivity timeouts.

What would we gain from squashing them into one?

tomschr:
> [...] What would we gain from squashing them into one?

Well, we (or better I) need to cherry-pick that in different maintenance branches too. This is easier with _one_ commit of course. :wink: 

Also, this PR is about a specific Bugzilla entry. From my point of view, I would see that in one single commit. But I know there are probably different perspectives on this. The more important thing is that I can handle it easier when cherry-picking.

mgerstner:
> Well, we (or better I) need to cherry-pick that in different maintenance branches too. This is easier with _one_ commit of course. 
> 
> Also, this PR is about a specific Bugzilla entry. From my point of view, I would see that in one single commit. But I know there are probably different perspectives on this. The more important thing for me is that it's easier when I do cherry-picking.

I think you are following a wrong pattern here. You shouldn't structure commits to simplify your merge workflow. This is something you do only once or twice, but the commit history will stay forever. Anybody trying to find out what was going on will have a hard time dissecting the individual parts. Also there can come up situations like "I only want that section" or "I only want the default attributes removed, but not the additional sections". In this situation the squashing will backfire.

But it is your repo, of course, so if you want it squashed, then so be it.

tomschr:
> I think you are following a wrong pattern here

Yeah, could be. 😁

Well, we have maintenance branches which are normally "stable". These get updates only when we cherry-pick something from master or explicitly add it into the branch. Apart from these, nothing is added.

> Also there can come up situations like "I only want that section" or "I only want the default attributes removed, but not the additional sections". In this situation the squashing will backfire.

Theoretically that's true. However, tThis happened to me only a few times over the years. Maybe for docs it's a big different than for code. :man_shrugging:  Normally, our bugs are a "all or nothing approach". :wink: 

I think, @lproven can start proofreading the work. We can later decide what we can do with our commits.

mgerstner:
It's fine with me, I already squashed all into one commit. Please go ahead.

xml/hardening.xml
```diff
@@ -3445,6 +3445,325 @@ Account Expires: Never</screen>
    </para>
   </sect1>
   <xi:include href="hardening_pam_stack.xml"/>
+  <sect1 xml:id="sec-sec-prot-restrict-root">
+   <title>Restricting &rootuser; Logins</title>
+
+   <para>
+    By default, the &rootuser; user is assigned a password
+    and can login via various methods like on any local terminal, in a
+    graphical session, or remotely via SSH. These methods should be restricted
```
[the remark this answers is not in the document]
mgerstner: We have both `local console` and `local terminal` used throughout the document currently, but not `system console` or `system terminal`. So I'd like to avoid introducing another term for the same thing.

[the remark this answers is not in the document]
mgerstner: I will push a change without the `system console` bit.

xml/hardening.xml
```diff
@@ -3445,6 +3445,325 @@ Account Expires: Never</screen>
...
+    as far as possible. Shared usage of the root account should be avoided,
```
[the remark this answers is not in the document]
mgerstner: It looks like this is missing a bit. When there is a fullstop the next word should start with a capital letter. I will adjust accordingly.

xml/hardening.xml
```diff
@@ -3445,6 +3445,325 @@ Account Expires: Never</screen>
...
+    instead individual administrators should use tools like
```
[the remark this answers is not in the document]
mgerstner: Ah there is the bit that I was missing before. Sorry.

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+    <para>
+     The following steps restrict root access to the first TTY. Even this
+     access method is only meant for emergency access to the system and should
+     never be used for everyday system administration tasks.
+    </para>
+
+    <note>
+     <para>
+      The steps shown here are tailored towards PC architectures (&x86; and
+      &x86-64;). On architectures like &power; different terminal
```
lproven, line 3494:
```suggestion
      &x86-64;). On architectures such as &power;, different terminal
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+      device names than <literal>tty1</literal> may be used. Be careful not to
+      lock yourself out completely by specifying wrong terminal device names.
+      You can determine the device name of the terminal you are currently
+      logged into by running the <literal>tty</literal> command. Be careful
+      not to do this in a virtual terminal like via SSH or in a graphical
```
lproven, line 3499:
```suggestion
      not to do this in a virtual terminal, such as via SSH or in a graphical
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+      session (device names <filename>/dev/pts/<replaceable>N</replaceable></filename>)
```
lproven, line 3500:
```suggestion
      session (device names <filename>/dev/pts/<replaceable>N</replaceable></filename>),
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+     <step>
+      <para>
+       Ensure that the PAM stack configuration file <filename>/etc/pam.d/login</filename>
+       contains the <literal>pam_securetty</literal> module in the <literal>auth</literal> block:
+      </para>
+      <screen>auth     requisite      pam_nologin.so
+ auth     [user_unknown=ignore success=ok ignore=ignore auth_err=die default=bad] pam_securetty.so noconsole
+ auth     include        common-auth</screen>
+      <para>
+        This will include the <literal>pam_securetty</literal> PAM module during the
```
lproven, line 3517:
```suggestion
        This will include the <literal>pam_securetty</literal> module during the
```

lproven: Since the "M" in "PAM" already stands for "module", this is redundant.

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+        authentication process on local consoles to allow root only to login on
+        tty devices that are listed in the file <filename>/etc/securetty</filename>.
```
lproven, line 3519:
```suggestion
        authentication process on local consoles, which restricts &rootuser; to logging-in
        only on TTY devices that are listed in the file <filename>/etc/securetty</filename>.
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+# This file contains the device names of tty lines (one per line,
+# without leading /dev/) on which root is allowed to login.
+#
+tty1</screen>
+     </step>
+     <step>
+      <para>
+       Check whether logins to other terminals will be rejected for
+       &rootuser;. A login on <literal>tty2</literal>, for example, should be
+       rejected right away without even the root password being queried.
```
lproven, line 3537:
```suggestion
       rejected immediately, without even querying the account password.
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+      </para>
+     </step>
+    </procedure>
+
+    <important>
+     <para>
+      Avoid adding the <literal>pam_securetty</literal> PAM module to the
+      <filename>/etc/pam.d/common-auth</filename> file. This change would
+      break <command>su</command> and <command>sudo</command>, however,
+      because these tools would then also reject &rootuser; authentications.
```
lproven, line 3550:
```suggestion
      Do not add the <literal>pam_securetty</literal> module to the
      <filename>/etc/pam.d/common-auth</filename> file. This would
      break the <command>su</command> and <command>sudo</command> commands,
      because these tools would then also reject &rootuser; authentications.
```

lproven: This might be better as a [warning] -- do you agree, @tomschr?

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+     </para>
+    </important>
+
+    <important>
+     <para>
+      These configuration changes will also cause root logins on serial
+      consoles like <filename>/dev/ttyS0</filename> to be denied. In case you
```
lproven, line 3557:
```suggestion
      consoles such as <filename>/dev/ttyS0</filename> to be denied. In case you
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+      require such use cases you need to list the respective tty devices
+      additionally in the <filename>/etc/securetty</filename> file.
```
lproven, line 3559:
```suggestion
      consoles like <filename>/dev/ttyS0</filename> to be denied. If you
      require such use cases, then also list the respective TTY devices
      in the <filename>/etc/securetty</filename> file.
```

lproven: Github timeouts again. >_<  I cannot revise my previous suggestion.

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+     </para>
+    </important>
+   </sect2>
+
+   <sect2 xml:id="sec-sec-prot-retrict-root-graphical">
+    <title>Restricting Graphical Session Logins</title>
+
+    <para>
+     To improve security on your server, avoid to use graphical user
```
lproven, line 3568:
```suggestion
     To improve security on your server, avoid using graphical
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+     environments at all. Graphical programs are often
+     not designed to be run as &rootuser; and are more likely to
+     contain security issues than console programs. If you require a graphical
+     login, use a non-&rootuser; login. Configure your system to disallow
+     &rootuser; login on graphical session.
```
lproven, line 3573:
```suggestion
     &rootuser; from logging in to graphical sessions.
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+     To deny the &rootuser; user login to graphical sessions you
+     can apply the same basic steps as outlined in
+     <xref linkend="sec-sec-prot-restrict-root-tty"/>.
+     You only need to also add the <literal>pam_securetty</literal> module to
+     the PAM stack file belonging to the display manager, for example
+     <filename>/etc/pam.d/gdm</filename> for GDM. The graphical session also
+     runs on a TTY device, <literal>tty7</literal>, by default. Therefore,
+     when you restrict &rootuser; logins to
+     <literal>tty1</literal>, then a &rootuser; login in the
+     graphical session will be denied.
```
lproven, line 3586:
```suggestion
     To prevent &rootuser; from logging in to graphical sessions, you
     can apply the same basic steps as outlined in
     <xref linkend="sec-sec-prot-restrict-root-tty"/>.
     Just add the <literal>pam_securetty</literal> module to the PAM
     stack file belonging to the display manager&mdash;for example,
     <filename>/etc/pam.d/gdm</filename> for GDM. The graphical
     session also runs on a TTY device: by default, <literal>tty7</literal>.
     Therefore, if you restrict &rootuser; logins to <literal>tty1</literal>,
     then &rootuser; will be denied login in the graphical session.
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+    </para>
+   </sect2>
+
+   <sect2 xml:id="sec-sec-prot-restrict-root-ssh">
+    <title>Restricting SSH Logins</title>
+    <para>
+     By default the &rootuser; user is also allowed to login
+     into a machine remotely via the SSH network protocol (if the SSH port is
+     not blocked by the firewall). To restrict this possibility the following
+     change of the OpenSSH configuration needs to be performed:
```
lproven, line 3596:
```suggestion
     By default, the &rootuser; user is also allowed to log
     into a machine remotely via the SSH network protocol (if the SSH port is
     not blocked by the firewall). To restrict this, make the following
     change to the OpenSSH configuration:
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+      <screen>systemctl restart sshd.service</screen>
+     </step>
+    </procedure>
+    <note>
+     <para>
+      Using the PAM module <literal>pam_securetty</literal> is not suitable in
+      case of OpenSSH, because SSH logins not necessarily go through the PAM
+      stack during the auth phase (e.g. when using SSH public key authentication).
+      Also an attacker could be able to differentiate between a wrong password
+      and a successful login that was only rejected later on by policy.
```
lproven, line 3614:
```suggestion
      Using the PAM <literal>pam_securetty</literal> module is not suitable in
      case of OpenSSH, because not all SSH logins go through the PAM
      stack during authorization (for example, when using SSH public-key authentication).
      In addition, an attacker could differentiate between a wrong password
      and a successful login that was only rejected later on by policy.
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+     </para>
+    </note>
+   </sect2>
+
+  </sect1>
+  <sect1 xml:id="sec-sec-prot-inactivity-logout">
+   <title>Setting an Inactivity Timeout for Interactive Shell Sessions</title>
+   <para>
+    It can be a good idea to terminate an interactive shell session after a
+    certain time of inactivity, for example to:
```
lproven, line 3624:
```suggestion
    certain period of inactivity; for example, to:
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+      <para>
+      prevent open, unguarded sessions.
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       avoid waste of system resources.
+      </para>
+     </listitem>
+    </itemizedlist>
```
lproven, line 3637: I think an itemizedlist is not needed for just 2 alternatives, but it is not a serious problem.

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+   <para>
+    By default, there is no inactivity timeout for shells. Nothing will
+    happen if a shell stays open unused for days or even years. It is possible
```
lproven, line 3640:
```suggestion
    happen if a shell stays open and unused for days or even years. It is possible
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+    shows how to set an inactivity timeout for a number of
+    common types of shells.
+   </para>
+   <para>
+    The inactivity timeout can be configured for login shells only or for all
+    interactive shells. In the latter case the inactivity timeout is running
+    individually for each shell instance. This means that timeouts will
+    accumulate. When a sub or child shell is started then the timeout runs for
+    the sub or child shell and only afterwards will the
+    timeout of the parent continue running.
```
lproven, line 3652:
```suggestion
    The inactivity timeout can be configured for login shells only or for all
    interactive shells. In the latter case the inactivity timeout is running
    individually for each shell instance. This means that timeouts will
    accumulate. When a sub- or child-shell is started, then a new timeout begins for
    the sub- or child-shell, and only afterwards will the
    timeout of the parent continue running.
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+   </para>
+   <para>
+    The following table contains configuration details for a selection of
+    common shells shipped with SLE:
```
lproven, line 3656: I suspect that there is an entity for "SLE" here.

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+       <entry><para><literal>readonly TMOUT=</literal></para></entry>
+       <entry><para><filename>/etc/profile.local</filename>, <filename>/etc/profile.d/</filename></para></entry>
+       <entry><para><filename>/etc/zsh.zshrc.local</filename></para></entry>
+      </row>
+     </tbody>
+    </tgroup>
+   </informaltable>
+   <para>
+    Every listed shell supports an internal timeout shell variable that can be
+    set to a specific time value to activate the inactivity timeout. If you want
```
lproven, line 3720:
```suggestion
    set to a specific time value to cause the inactivity timeout. If you want
```

lproven: "Activate the inactivity" sounded odd to me.

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+    to prevent users from overriding the timeout setting you can mark the
```
lproven, line 3721:
```suggestion
    to prevent users from overriding the timeout setting, you can mark the
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+    declaration syntax is also found in the table above.
+   </para>
+
+   <note>
+    <para>
+     This feature is only a utility that can help avoiding risks if a user is
+     forgetful or follows unsafe practices. It does not protect against
+     malicious users, however. The timeout discussed here only applies to
+     certain interactive wait states of a shell. A malicious user can always
+     find ways to circumvent the timeout and keep their session open anyway.
```
lproven, line 3732:
```suggestion
     This feature is only helpful for avoiding risks if a user is forgetful or follows
     unsafe practices. It does not protect against hostile users. The timeout
     only applies to interactive wait states of a shell. A malicious user can always
     find ways to circumvent the timeout and keep their session open regardless.
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+
+   <para>
+    To configure the inactivity timeout you need to add the matching timeout
+    variable declaration to a shell startup script. Use either the path for
+    login shells only, or the one for all shells as listed in the table. The
+    following example uses paths and settings that are suitable for
+    <command>bash</command> and <command>ksh</command>
+    to setup a readonly login shell timeout that cannot be overriden by users.
+    Create the file <filename>/etc/profile.d/timeout.sh</filename> with the
+    following content:
```
lproven, line 3744:
```suggestion
    To configure the inactivity timeout, you need to add the matching timeout
    variable declaration to each shell's startup script. Use either the path for
    login shells only, or the one for all shells, as listed in the table. The
    following example uses paths and settings that are suitable for
    <command>bash</command> and <command>ksh</command>
    to setup a read-only login shell timeout that cannot be overridden by users.
    Create the file <filename>/etc/profile.d/timeout.sh</filename> with the
    following content:
```

xml/hardening.xml
```diff
@@ -3445,6 +3445,327 @@ Account Expires: Never</screen>
...
+# in case of inactivity.
+# 24h = 86400 sec
+readonly TMOUT=86400</screen>
+
+   <tip>
+    <para>
+     We recommend to use the <command>screen</command> tool in order to
+     detach sessions before logging out. Screen sessions are not terminated
+     and can be re-attached whenever it is required. Active session can be
+     locked without logging out (read about
```
lproven, line 3758:
```suggestion
     We recommend using the <command>screen</command> tool in order to
     detach sessions before logging out. <command>screen</command> sessions
     are not terminated and can be re-attached whenever required. An active
     session can be locked without logging out (read about
```

lproven (changes requested):

xml/hardening.xml
tomschr: Yes, we can do that in the fine-tuning part. :wink:

tomschr:
For some reasons I don't know, GitHub doesn't allow me to do batch committing of Liam's changes. So I need to apply them individually. :-(

xml/hardening.xml
tomschr: I used the `&productname;` in commit 18deb0107

xml/hardening.xml
tomschr: I've used inline text instead of the 2 items, comitted in 18deb0107.

PR #658, opened 2020-09-21
AutoYaST Support for Dynamic Profiles

imobachgs:
### Description

This PR adds some sections about AutoYaST capabilities to generate dynamic profiles. In SP3 we have added support for "ERB Templates" and a new command to check whether a profile is correct or not (`check-profile`).

As usually, I am open to any change, as I am not even sure about the organization of the contents.

* [bsc#1175735](https://bugzilla.suse.com/show_bug.cgi?id=1175735)

We are currently **only merging documentation relevant for SLE 15 SP2**
(and Leap 15.2) or lower.

- [x] This PR only applies to SLE 15 SP3 or higher
- [ ] This PR applies to older releases as well.

### Are backports required?

- [ ] To maintenance/SLE15SP2
- [ ] To maintenance/SLE15SP1
- [ ] To maintenance/SLE15SP0
- [ ] To maintenance/SLE12SP5
- [ ] To maintenance/SLE12SP4
- [ ] To maintenance/SLE12SP3
- [ ] To maintenance/SLE12SP2

### DIFF
--- a/xml/ay_create_control_file.xml
+++ b/xml/ay_create_control_file.xml
@@ -340,4 +340,104 @@
     <link xlink:href="http://www.w3.org/TR/xslt">www.w3.org/TR/xslt</link>
    </para>
   </sect1>
+
+  <sect1 xml:id="check-profile">
+   <title>Checking a Control File</title>
+
+   <para>
+    Depending on the use case, creating an &ay; profile can be difficult,
+    especially if you build a dynamic one using <link
+    xlink:href="part-dynamic-profiles">rules/classes, ERB templates or
+    pre-scripts</link>. To simplify the testing and debugging process, &ay;
+    offers the <literal>check-profile</literal> command, which takes care of
+    fetching, building and, optionally, importing the profile to detect any
+    potential problem.
+   </para>
+
+   <note>
+    <para>
+     Although this command uses the same approach than the installation, the
+     results may vary depending on the differences between the current system
+     and installation media: &yast; package versions, architecture, etc.
+    </para>
+   </note>
+
+   <warning>
+    <para>
+     You must be careful when running this command because pre-installation
+     scripts and ERB code would run as the <literal>root</literal> user.
+     So, please, only use profiles that you trust.
+    </para>
+   </warning>
+
+   <sect2>
+    <title>Basic Checks</title>
+
+    <para>
+     The simplest way to use this command is just to read and validate the profile:
+    </para>
+
+<screen>&prompt.sudo; yast2 autoyast check-profile filename=autoinst.xml output=result.xml</screen>
+
+    <para>
+     The <filename>result.xml</filename> file contains the result of evaluating
+     the profile. Bear in mind that, even if you do not use any advanced
+     feature, the content of <filename>autoinst.xml</filename> and
+     <filename>result.xml</filename> may differ. The reason is that &ay; does
+     some clean-up when it processes the profile.
+    </para>
+
+    <para>
+     <literal>check-profile</literal> can deal with remote files too:
+    </para>
+
+<screen>&prompt.sudo; yast2 autoyast check-profile filename=http://192.168.1.100/autoinst.xml output=result.xml</screen>
+
+   </sect2>
+
+   <sect2>
+    <title>Running pre-scripts</title>
+
+    <para>
+     Optionally, &ay; is able to run the scripts that are included in the
+     profile, reporting any error found during the execution. This is especially
+     relevant if you are using a pre-installation script to modify the profile.
+     To enable this feature, you need to set the <literal>run-scripts</literal>
+     to <literal>true</literal>.
+    </para>
+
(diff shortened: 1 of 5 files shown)


xml/ay_create_control_file.xml
```diff
@@ -297,4 +297,83 @@
...
+<screen>yast2 autoyast check-profile filename=http://192.168.1.100/autoinst.xml output=result.xml</screen>
+
+   </sect2>
+
+   <sect2>
+    <title>Running pre-scripts</title>
+
+    <para>
+     Optionally, &ay; is able to run pre-scripts too. In order to enable this feature,
+     you need to set the <literal>run-scripts</literal> to <literal>true</literal>.
```
jreidinger, line 344: This is a bit confusing as it looks like it runs only pre-scripts, but run-scripts run all defined scripts and report any error it find during execution of such scripts.

xml/ay_create_control_file.xml
```diff
@@ -297,4 +297,83 @@
...
+    </para>
+
+<screen>yast2 autoyast check-profile filename=http://192.168.1.100/autoinst.xml output=result.xml run-scripts=true</screen>
+
+    <warning>
+     <para>
+      Needless to say that you must be careful when running pre-installation
+      scripts as root.
```
jreidinger, line 352: Maybe I would rephrase it as all scripts runs as root on current system, so any operations are performed including destructive ones.

xml/ay_create_control_file.xml
```diff
@@ -297,4 +297,83 @@
...
+    </warning>
+   </sect2>
+
+   <sect2>
+    <title>Importing the Profile</title>
+
+    <para>
+     It is possible to face some problems when importing a valid profile, even
+     if it is correct. The reason is that AutoYaST does not perform any logic
+     check when fetching, building and validating the profile.
```
jreidinger, line 363: It is by default enabled, so it is more likely just import-all=false which is needed when you just want resulting profile and not do import checks ( including autoyast issues ). I would also note that it do check on current system with current context, so e.g. different archs or code streams can have different issues or it can contain for them false positives. and same for xml schema check.

xml/ay_dynamic_profiles.xml
```diff
@@ -0,0 +1,60 @@
...
+       profile's content during installation. The <xref
+       linkend="erb-templates"/> section describes how to use them.
+      </para>
+     </listitem>
+   </varlistentry>
+   <varlistentry>
+     <term>Pre-installation scripts</term>
+     <listitem>
+      <para>
+       Perhaps the most flexible approach is to use a pre-installation script to
```
jreidinger, line 53: I think that ERB and pre-script is on same level of flexibility.

xml/ay_erb_templates.xml
```diff
@@ -0,0 +1,467 @@
...
+  <para>
+   ERB templates allow to embed some Ruby code within the profile to modify it
+   during the installation. With this approach, you can inspect the system and
+   adjust the profile by setting values, adding or skipping sections, etc.
+  </para>
+
+  <para>
+   In order to activate the ERB processing, the profile must have the extension
+   <filename>.erb</filename> (e.g., <filename>autoyast.xml.erb</filename>). Hence,
+   it is not possible to combine rules and ERB templates.
```
jreidinger, line 30: not just rules, also classes are not supported.

cwickert:
The "Fix bootloader example" commit would go into other branches as well, right?

imobachgs:
> The "Fix bootloader example" commit would go into other branches as well, right?

Oh, sorry, you are right.. Do you want me to create a separate PR for it?

xml/ay_create_control_file.xml
```diff
@@ -297,4 +297,96 @@
...
+     relevant if you are using a pre-installation script to modify the profile.
+     To enable this feature, you need to set the <literal>run-scripts</literal>
+     to <literal>true</literal>.
+    </para>
+
+<screen>&prompt.sudo; yast2 autoyast check-profile filename=http://192.168.1.100/autoinst.xml output=result.xml run-scripts=true</screen>
+
+    <warning>
+     <para>
+      Needless to say that you must be careful when enabling the
```
lslezak, line 362: Not only the users scripts can be dangerous. With ERB you can simply put a backtick anywhere.

Possibly with a nasty content like this:
```
<% `rm -rf /` %>
```

Users should not blindly try any random XML profile they find on the internet...

So we should add some additional warning about this here.

imobachgs:
I have rebased these changes on `master` including the feedback from your comments. I would say that now it is ready for review. I will open a separate PR to fix the bootloader example.

cjschroder:
Hi @imobachgs, happy new year! Thanks for another great AY PR! I made a few small grammatical corrections, would you look at  ay_erb_templates.xml and make sure it is correct? I think the word "helper" was left out in a few places. If it looks good I will merge.

imobachgs:
@cjschroder Thanks a lot! Yes, you are right, the `helper` word was missing there. I have fixed a conflict and I would say that it is finally ready.

Thanks!

cjschroder:
If it isn't ready it's too late now. Merging!

PR #665, opened 2020-09-29
Container guide podman

dmpop:
### Description

Added basic information on Podman, Buildah, and working with pods.

* bsc#...
* jsc#SLE-...

| Code 15     | Backport Done
| ----------  | ---------- 
| [ ] 15 SP3  | (`master` only)   
| [ ] 15 SP2  | [ ] 
| [ ] 15 SP1  | [ ] 
| [ ] 15 SP0  | [ ] 

| Code 12     | Backport Done
| ----------  | ---------- 
| [ ] 12 SP5  | [ ] 
| [ ] 12 SP4  | [ ] 
| [ ] 12 SP3  | [ ] 
| [ ] 12 SP2  | [ ]


xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,209 @@
...
+   </para>
+  </listitem>
+  <listitem>
+   <para>
+    Create an image from a working container or via Dockerfile.
+   </para>
+  </listitem>
+  <listitem>
+   <para>
+    Build images in the OCI image format or the traditional upstream &deng;
```
Vogtinator, line 29: I don't think `traditional` or `upstream` are really meaningful here:

```suggestion
    Build images in the OCI or &deng; image formats
```

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,209 @@
...
+   </para>
+  </listitem>
+  <listitem>
+   <para>
+    Delete a working container or an image and rename a local container.
+   </para>
+  </listitem>
+ </itemizedlist>
+ <para>
+  Compared to &deng;, &buildah; has several important advantages.
```
Vogtinator, line 51: Whether those are important or not is subjective.

```suggestion
  Compared to &deng;, &buildah; has several advantages.
```

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,209 @@
...
+ </para>
+ <itemizedlist>
+  <listitem>
+   <para>
+    &buildah; adheres the UNIX philosophy by focusing only on one task:
```
Vogtinator, line 56: Some would say that's a disadvantage... IMO it doesn't belong in the list here, more into the description of buildah

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,209 @@
...
+    building container images.
+   </para>
+  </listitem>
+  <listitem>
+   <para>
+    The tool makes it possible to mount a working container's filesystem, so it
+    becomes accessible by host.
```
Vogtinator, line 63:
```suggestion
    becomes accessible by the host.
```

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,209 @@
...
+ <sect1 xml:id="cha-podman-buildah">
+  <title>&podman; and &buildah;</title>
+
+  <para>
+   &podman; and &buildah; are complimentary tools for working with OCI images
+   and containers. While &buildah; focuses on building images, &podman; offers
+   commands and functions for maintaining and modifying images. It also makes
+   it possible to create, run, and maintain containers created from images.
+   Although &podman; allows you to build images via Dockerfile, it uses
+   &buildah;'s golang API to accomplish that.
```
Vogtinator, line 93: This reads like buildah needs to be installed to use podman, but AFAIK that's not true.

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,209 @@
...
+    &docker; socket, which improves security.
+   </para>
+  </listitem>
+ </itemizedlist>
+ <sect1 xml:id="cha-podman-buildah">
+  <title>&podman; and &buildah;</title>
+
+  <para>
+   &podman; and &buildah; are complimentary tools for working with OCI images
+   and containers. While &buildah; focuses on building images, &podman; offers
```
Vogtinator, line 89: Important to mention: `podman`'s CLI is supposed to be compatible with `docker`

Vogtinator (commented):
Looking good, but I don't know much about podman, so I'd leave that to someone else.

@sysrich maybe?

xml/containers-buildah-overview.xml
flavio: I agree, I think the `Although... accomplish that` line can be confusing for a new user. I would just point out that podman build images using `Dockerfile` like the docker container engine does.
The end user doesn't care about the implementation details.

xml/containers-orchestration.xml
```diff
@@ -0,0 +1,59 @@
...
+   work with pods. A pod is a group of one or more containers, with shared
+   storage and network resources. A pod usually encapsulates an application
+   composed of multiple containers into a single unit.
+  </para>
+
+  <para>
+   The <command>podman pod</command> can be used to create, delete, query, and
+   inspect pods. To create a new pod, run the <command>podman pod create --name
+   <replaceable>POD_NAME</replaceable></command> command. This creates a pod
+   with a random name. To list the existing pods, use the <command>podman pod
```
flavio, line 28: If you run `podman pod create --name foo` it won't create a pod with a random name :)

xml/containers-orchestration.xml
```diff
@@ -0,0 +1,59 @@
...
+  </para>
+
+<screen>podman run -dt --pod example registry.suse.com/suse/sle15 bash</screen>
+
+  <para>
+   Containers in a pod can be restarted, stopped, and started without affecting
+   the overall status of the pod. For example, you can stop the container
+   <literal>example</literal> using the <command>sudo podman stop
+   example</command> command.
+  </para>
```
flavio, line 51: I think this is confusing. In the previous example you started a sles container inside of a pod named `example`.

Now you say: you can stop a container named `example` using `<cmd>`; using `example` with two different meanings (first to address a pod, now a single container within a pod) can be confusing IMHO.

I would make something more lengthy such as:

  * When talking about pod creation:
    * We create a pod named `example`
    * We show the actual command being run
    * We show the actual output of `podman ps -a --pod`. We tell the reader the output will look slightly different to him (container IDs, ...)
  * When talking about how to add a container to an existing pod:
    * We show the actual command, we add the container to the `example` pod created before
    * Maybe we don't start an interactive session, we could do something like `sudo podman run -d --pod example registry.suse.com/suse/sle15 sleep 1h`. This start a container that will sleep for 1 hour and then exit
    * We show again the output of `podman ps -a --pod example`
* When talking about how to stop a container:
    * We show the actual command, using the actual data of the "demo" system (aka the IDs shown with the previous `ps -a`)

xml/containers-orchestration.xml
```diff
@@ -0,0 +1,59 @@
...
+
+  <para>
+   To stop the pod, use the <command>podman pod stop</command> command:
+  </para>
+
+<screen>podman pod stop <replaceable>POD_NAME</replaceable></screen>
```
flavio, line 57: I would keep the same format I described before

flavio (changes requested):
Great work, I left some minor comments.

xml/containers-basics.xml
```diff
@@ -135,21 +134,18 @@
...
 
   <para>
-   After you pull a base image, you can use it as a starting point for building
-   a custom image. For example, you can use the SLE base image to generate an
-   image that is configured to run a specific application. To build a custom
-   image, you need to create a special file called Dockerfile, containing
-   building instructions. For example, a Dockerfile can contain instructions to
-   update the system software, install the desired application, open specific
-   network ports, run commands, etc.
+   To build a custom image, you need to create a special file called Dockerfile,
```
sysrich, line 141: Do we not also document how to build containers with our KIWI stack?

xml/containers-buildah-overview.xml
sysrich: Indeed. `podman build` users an embedded/vendored copy of buildah to give `podman`  the equivalent functionality of `docker build`, but you do not need buildah to be installed.

If I were to describe it simply I would say that `podman` providers a `docker build`-like experience for building containers, whereas `buildah` provides an expanded range of options and capabilities for building containers. Whether I'd document the fact `podman` uses the `buildah` codebase to accomplish this seems like a cute optional implimentation detail to me.

xml/containers-podman-overview.xml
```diff
@@ -0,0 +1,152 @@
...
+ <itemizedlist>
+  <listitem>
+   <para>
+    &podman; does not uses a daemon, so the container engine interacts directly
+    with an image registry, containers, and image storage. As &podman; does not
+    have a daemon, it provides integration with systemd. This makes it possible
+    to control containers via systemd units. You can create these units for
+    existing containers as well as generate units that can start containers if
+    they do not exist in the system. Moreover, &podman; can run systemd inside
+    containers.
```
sysrich, line 44: There is an additional benefit for the daemon-less architecture.

In the unfortunate case of the Docker daemon crashing/dying/being terminated then all of the containers become unmanageable.

podman avoids this potential denial-of-service issue, while also enabling the systemd management route described above.

xml/containers-podman-overview.xml
```diff
@@ -0,0 +1,152 @@
...
+
+<screen>&prompt.user; sudo usermod --add-subuids 200000-201000 --add-subgids
+   200000-201000 $USER</screen>
+
+  <para>
+   Log out and log in to apply changes.
+  </para>
+
+  <para>
+   In certain situations, running a container with &podman; in the rootless
```
sysrich, line 78: "In certain situations" - isn't the only situation "when running a SLES container with &podman; in the rootless...."

xml/containers-podman-overview.xml
```diff
@@ -0,0 +1,152 @@
...
+ </sect1>
+ <sect1 xml:id="cha-podman-use">
+  <title>&podman; Basic Usage</title>
+
+  <para>
+   Since &podman; is compatible with &deng;, it features the same commands and
+   options. For example, the <command>podman pull</command> command fetches a
+   container image from a registry, while the <command>podman build</command>
+   command is used to build images.
+  </para>
```
sysrich, line 127: Might this be a good place to suggest `alias docker=podman` ?

xml/containers-basics.xml
```diff
@@ -135,21 +134,18 @@
   <para>
    Both base and custom images are usually available through a repository of
    images called registry. Unless a registry is explicitly specified, &podman;
-   pulls images from the Docker Hub registry.
+   pulls images from the &dhub; registry. While you can fetch a base image manually, &podman; can do that automatically when building a custom image.
```
sysrich, line 137: One of the other advantages of `podman` over `docker` is that it can be configured to search multiple registries.

For example, on openSUSE the default `/etc/containers/registries.conf` contains the following:

```
[registries.search]
registries = ["registry.opensuse.org", "docker.io"]
```

This means that `podman` searches the openSUSE Registry for any requested tags before using Docker Hub as a fallback.

Even if we dont' do something by default in SLE yet (why not?) I think it's probably a good thing to document this functionality and get users thinking about different registries and how to use them, either by configuring `/etc/containers/registries.conf` or by including the full registry domain in all of their `docker|podman pull` calls, etc.

xml/containers-orchestration.xml
```diff
@@ -0,0 +1,59 @@
...
+ <info>
+  <title>Container Orchestration</title>
+ </info>
+ <sect1 xml:id="sec-pod-deployment-with-podman">
+  <title>Pod Deployment with Podman</title>
+
+  <para>
+   In addition to building and managing images, &podman; makes it possible to
+   work with pods. A pod is a group of one or more containers, with shared
+   storage and network resources. A pod usually encapsulates an application
```
saschagrunert, line 20: Storage is not shared across Pods, but can be done via the concept of volumes. Per default the `cgroup`, `ipc`, `net` ,`uts` namespaces are shared. This can be also modified by the user.


```suggestion
   work with pods. A pod is a group of one or more containers, with shared
   resources like the network interface. A pod usually encapsulates an application
```

xml/containers-podman-overview.xml
```diff
@@ -0,0 +1,152 @@
...
+    have a daemon, it provides integration with systemd. This makes it possible
+    to control containers via systemd units. You can create these units for
+    existing containers as well as generate units that can start containers if
+    they do not exist in the system. Moreover, &podman; can run systemd inside
+    containers.
+   </para>
+  </listitem>
+  <listitem>
+   <para>
+    Because &podman; relies on user namespaces, which provide an isolation
```
saschagrunert, line 49: There are multiple namespaces involved to allow rootless execution as well as some glue-tools like slirp4netns and fuse-overlayfs. I'm not sure if we have to deep-dive into the details here. We probably can just mention that it works :shrug:

saschagrunert (commented):
Nice work, I've just got a few comments.

xml/containers-podman-overview.xml
```diff
@@ -0,0 +1,152 @@
...
+    mechanism for Linux processes, it does not require root privileges to
+    create and run containers. This means that &podman; can run in the root
+    mode as well as in an unpriviledged environment.
```
rhafer, line 52: we might want to add a few details about what this actually means.

In theory docker can also be used by "unprivileged" users. You just need to add them to the "docker" group. However, that is dangerous as the docker daemon and the containers are still launched by he root user. (So the user could basically get full root access to the local system e.g. by just starting a privileged container)

With podman however, containers created by unprivileged users won't get any higher privileges on the host  than what the user who started the container has.

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,209 @@
...
+  </listitem>
+  <listitem>
+   <para>
+    The process of building container images using &buildah; can be automated
+    using scripts.
+   </para>
+  </listitem>
+  <listitem>
+   <para>
+    Similar to &podman;, &buildah; supports the rootless mode.
```
rhafer, line 74: "Similar to podman, buildas doesn't require a daemon to run and can be used by unprivileged users". (and then maybe refer to the section about rootless podman.

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,209 @@
...
+
+  <para>
+   If you already have &podman; installed and set up for use in the rootless
+   mode, &buildah; can be used in an unprivileged environment without any
+   further configuration. If you have to enable the rootless mode for
+   &buildah;, run the following command:
+  </para>
+
+<screen>&prompt.user;sudo usermod --add-subuids 200000-201000 --add-subgids
+   200000-201000 $USER</screen>
```
rhafer, line 113: The same comments about subuid/subgid that I already added to the podman overview apply here.

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,209 @@
...
+
+  <note>
+   <title>&buildah; in the rootless mode</title>
+   <para>
+    In the rootless mode, &buildah; commands must be executed in the user
+    namespace of the user. To enter the user namespace, use the
+    <command>buildah unshare</command> command. Otherwise, the <command>buildah
+    mount</command> command will fail.
+   </para>
+  </note>
```
rhafer, line 127: I think we also need to have a Note here about giving access to the sle credentials when building as non-root.

xml/containers-podman-overview.xml
rhafer: I think we should not do that. There are enough smaller difference that just cause surprises/frustation if the `docker` command suddenly behaves weird.

rhafer (changes requested):
Nice work! I added a few comments where I think things should be tweaked.

xml/containers-basics.xml
dmpop: Yes, eventually. Podman and Buildah are the first priority.

xml/containers-basics.xml
```diff
@@ -135,21 +134,18 @@
...
-   network ports, run commands, etc.
+   To build a custom image, you need to create a special file called Dockerfile,
+   containing building instructions. For example, a Dockerfile can contain
+   instructions to update the system software, install the desired application,
+   open specific network ports, run commands, etc.
   </para>
 
   <para>
-   Note that you can build images not only from base images, but also on top of
+   You can build images not only from base images, but also on top of
```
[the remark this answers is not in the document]
dmpop: It's a matter of taste and style, I guess. Avoiding _you_ can make the text sound stilted and contrived. In any case, we use _you_ throughout our entire documentation.

xml/containers-podman-overview.xml
dmpop: Agree. However, I'd like to save it for a more in-depth look at Podman.

xml/containers-buildah-overview.xml
dmpop: Could you elaborate on that? Thanks!

xml/containers-basics.xml
```diff
@@ -68,8 +68,7 @@
   <para>
    Although &deng; is a popular choice for working with images and containers,
    &podman; provides a drop-in replacement for &docker; that offers several
-   advantages. While <xref linkend="sec-build-tools-podman" /> provides an
-   in-depth information on &podman;, this chapter offers a quick introduction
+   advantages. While <xref linkend="cha-podman-overview" /> provides more information on &podman;, this chapter offers a quick introduction
```
saschagrunert, line 71: nit: keep line width of the paragraph

xml/containers-build-tools.xml
```diff
@@ -44,39 +44,26 @@
...
-    in a Docker environment.
+    in a &deng; environment.
    </para>
    <para>
-    For instructions on how to install and use Buildah, see the
-    <link xlink:href="https://buildah.io/">Buildah Documentation</link> and
-    <link xlink:href="https://github.com/containers/buildah/blob/master/docs/tutorials">tutorials</link>
-    .
+    For information on installing and using &buildah;, refer to
+    <xref linkend="cha-buildah-overview" />. .
```
saschagrunert, line 66:
```suggestion
    <xref linkend="cha-buildah-overview" />.
```

xml/containers-podman-overview.xml
```diff
@@ -0,0 +1,172 @@
...
+    have a daemon, it provides integration with systemd. This makes it possible
+    to control containers via systemd units. You can create these units for
+    existing containers as well as generate units that can start containers if
+    they do not exist in the system. Moreover, &podman; can run systemd inside
+    containers.
+   </para>
+  </listitem>
+  <listitem>
+   <para>
+    Because &podman; relies on user namespaces, which provide an isolation
```
saschagrunert, line 47:
```suggestion
    Because &podman; relies on several namespaces, which provide an isolation
```

saschagrunert (approved):
Very well written, I'd like to see it in rather sooner than later. I just found some nits, otherwise 
/lgtm

flavio (approved):
LGTM :clap:

saschagrunert (approved):

sysrich (approved):
LGTM too :)

lproven (changes requested):
WIP review -- more to come

Github is not cooperating so I will come back to this tomorrow.

Additional changes so far:

Line 39:

> containers through <xref linkend="gloss-vt-lxc-cgroups"/> .

No link visible in HTML preview

Line 60

>     profile.
profiles.

xml/containers-build-tools.xml
```diff
@@ -32,7 +32,7 @@
     Docker Open Source Engine&mdash;a lightweight virtualization solution to
     run containers simultaneously on a single host. &docker; containers can be
     built using <xref linkend="Dockerfile"/> Dockerfiles. For a general
```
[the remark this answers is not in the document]
lproven: Or, alternatively:

«
built using &lt;xref linkend="Dockerfile"&gt;Dockerfiles&lt;/xref&gt;. For a general
»

... but AIUI house style means that the text of the URL should be visible in print – no?

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,211 @@
...
+   &podman; makes it possible to build images using Dockerfiles, &buildah;
+   offers an expanded range of image building options and capabilities.
+  </para>
+ </sect1>
+ <sect1 xml:id="cha-buildah-install">
+  <title>&buildah; Installation</title>
+
+  <para>
+   To install &buildah;, run the <command>sudo zypper in buildah</command>. Run
+   then <command>buildah --version</command> to check whether &buildah; has
```
lproven, line 93:
```suggestion
   To install &buildah;, run the <command>sudo zypper in buildah</command>. Then 
   run the command <command>buildah --version</command> to check whether &buildah; has
```

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,211 @@
...
+   been installed successfully.
+  </para>
+
+  <para>
+   If you already have &podman; installed and set up for use in the rootless
+   mode, &buildah; can be used in an unprivileged environment without any
+   further configuration. If you have to enable the rootless mode for
+   &buildah;, run the following command:
```
lproven, line 101:
```suggestion
   further configuration. If you need to enable &buildah;'s rootless mode,
   run the following command:
```

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,211 @@
...
+   the command, log out and log in again to enable the changes.
+  </para>
+
+  <para>
+   The command above defines a range of local uids to which the uids allocated
+   to users inside the container are mapped on the host. Note that the ranges
+   defined for different users must not overlap. It is also important that the
+   ranges do not reuse the uid of an existing local user or group. By default,
+   adding a user with the <command>useradd</command> on &slsa; &product-ga;,
+   automatically allocates subuid and subgid ranges.
```
lproven, line 118:
```suggestion
   The command above defines a range of local UIDs on the host, on to which
   the UIDs allocated to users inside the container are mapped. Note that the ranges
   defined for different users must not overlap. It is also important that the
   ranges do not reuse the UIDs of any existing local users or groups. By default,
   adding a user with the <command>useradd</command> on &slsa; &product-ga;,
   automatically allocates sub-UID and sub-GID ranges.
```

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,211 @@
...
+  </para>
+
+  <note>
+   <title>&buildah; in the rootless mode</title>
```
lproven, line 122:
```suggestion
   <title>&buildah; in rootless mode</title>
```

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,211 @@
...
+   <para>
+    In the rootless mode, &buildah; commands must be executed in a modified
+    user namespace of the user. To enter the user namespace, use the
+    <command>buildah unshare</command> command. Otherwise, the <command>buildah
+    mount</command> command will fail.
```
lproven, line 127:
```suggestion
    In its rootless mode, &buildah; commands must be executed in a modified
    user namespace of the user. To enter this user namespace, run the command
    <command>buildah unshare</command>. Otherwise, the <command>buildah
    mount</command> command will fail.
```

lproven: I have tried to simplify this a little, including removing one of many occurrences of forms of the word "use". However, I think the phrase:

>  a modified user namespace of the user

... is confusing. "User namespace of the user"? Of **what** user?

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,211 @@
...
+   </para>
+  </note>
+ </sect1>
+ <sect1 xml:id="cha-buildah-build-images">
+  <title>Building Images with &buildah;</title>
+
+  <para>
+   Instead of a special file with instructions, &buildah; uses individual
+   commands to build an image. Building an images with &buildah; involves
```
lproven, line 136:
```suggestion
   commands to build an image. Building an image with &buildah; involves
```

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,211 @@
...
+    <callout arearefs="ctr">
+     <para>
+      Run a container (also called a working container) based on the specified
+      image (in this case, <literal>sle15</literal>).
+     </para>
+    </callout>
+    <callout arearefs="run">
+     <para>
+      Run a command in the created working container. In this example,
+      &buildah; run the <command>zypper up</command> command.
```
lproven, line 172:
```suggestion
      Run a command in the working container you just created. In this
      example, &buildah; runs the <command>zypper up</command>
      command.
```

xml/containers-buildah-overview.xml
```diff
@@ -0,0 +1,211 @@
...
+     </para>
+    </callout>
+    <callout arearefs="copy">
+     <para>
+      Copy files and directories to the specified location in the container. In
+      this example, the &buildah; copies the entire contents of the current
```
lproven, line 178:
```suggestion
      this example, &buildah; copies the entire contents of the current
```

xml/containers-orchestration.xml
```diff
@@ -0,0 +1,91 @@
...
+    xmlns:xlink="http://www.w3.org/1999/xlink">
+ <info>
+  <title>Container Orchestration</title>
+ </info>
+ <sect1 xml:id="sec-pod-deployment-with-podman">
+  <title>Pod Deployment with Podman</title>
+
+  <para>
+   In addition to building and managing images, &podman; makes it possible to
+   work with pods. A pod is a group of one or more containers, with shared
```
lproven, line 19:
```suggestion
   work with pods. A pod is a group of one or more containers with shared
```

xml/containers-orchestration.xml
```diff
@@ -0,0 +1,91 @@
...
+   resources like the network interface. A pod usually encapsulates an
```
lproven, line 20:
```suggestion
   resources, such as the network interface. A pod usually encapsulates an
```

lproven (changes requested):
WIP review continued (as Github can choke on double-digit numbers of suggestions)

xml/containers-podman-overview.xml
```diff
@@ -0,0 +1,172 @@
...
+   SUSE registry first and use Docker Hub as a fallback, add the following
+   configuration to the <filename>/etc/containers/registries.conf</filename>
+   file:
+  </para>
+
+<screen>[registries.search]
+registries = ["registry.suse.com", "docker.io"]</screen>
+
+  <para>
+   Similar to &deng;, &podman; can run containers in the interactive mode,
```
lproven, line 150:
```suggestion
   Similar to &deng;, &podman; can run containers in an interactive mode,
```

lproven (commented):
WIP review -- last phase is next, when I will read through the rendered output again in case anything else strikes me.

xml/containers-buildah-overview.xml
lproven: Ah, now I understand. So, no, please ignore this suggestion.

xml/containers-orchestration.xml
```diff
@@ -0,0 +1,91 @@
...
+   of the command looks as follows (the <literal>STATUS</literal> and
+   <literal>CREATED</literal> columns are omitted for brevity):
+  </para>
+
+<screen>POD ID        NAME                # OF CONTAINERS   INFRA ID
+399a120a09ff  suspicious_curie    1                 e57820093817
+</screen>
+
+  <para>
+   Notice that the the command assigned a random name to the pod
```
lproven, line 39:
```suggestion
   Notice that the command assigned a random name to the pod
```

lproven (changes requested):

lproven (commented):
Other edits in a PR.

xml/containers-buildah-overview.xml
rhafer: When building a SLES based images with `builah` (or `podman build`) as a non root user. The credentials to access SCC need to be readable by that user. So basically what is mentioned here: https://github.com/SUSE/doc-sle/pull/665/files#diff-7a248a02092ab868664673831c4f2dd70f5a1e2a8e209178ec60066a8f73988dR100

I think this should be mentioned when talking about "rootless" container builds. (Though I am not sure where exactly ;-)  )

PR #675, opened 2020-10-02
AY: document the new behaviour during validation

imobachgs:
### Description

During validation, now AutoYaST ignores unknown top-level sections. See [bsc#1177173](https://bugzilla.suse.com/show_bug.cgi?id=1177173) for the rationale.

* bsc#1177173

| Code 15     | Backport Done
| ----------  | ---------- 
| [ ] 15 SP2  | [ ] 
| [ ] 15 SP1  | [ ] 
| [ ] 15 SP0  | [ ] 

| Code 12     | Backport Done
| ----------  | ---------- 
| [ ] 12 SP5  | [ ] 
| [ ] 12 SP4  | [ ] 
| [ ] 12 SP3  | [ ] 
| [ ] 12 SP2  | [ ]

### DIFF
--- a/xml/ay_create_control_file.xml
+++ b/xml/ay_create_control_file.xml
@@ -187,6 +187,34 @@
     describes the syntax and classes of an &ay; profile.
    </para>
 
+   <note>
+    <title>Schema Extensions</title>
+
+    <para>
+     &ay; can be extended by other products and modules, but the schema does not
+     contain the specification for those extensions. As a consequence, &ay;
+     a profile that use one of those extensions might be reported as invalid.
+    </para>
+
+    <para>
+     Thus, starting in &productname; <phrase os="osuse">15.3</phrase><phrase
+     os="sles;sled">SP3</phrase>, &ay; does not validate top-level unknown
+     sections. For example, in the example below, the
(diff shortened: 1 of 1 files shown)


xml/ay_create_control_file.xml
```diff
@@ -187,6 +187,34 @@
     describes the syntax and classes of an &ay; profile.
    </para>
 
+   <note>
+    <title>Schema Extensions</title>
+
+    <para>
+     &ay; can be extended by other products and modules, but the schema does not
+     contain the specification for those extensions. As a consequence, &ay;
+     a profile that use one of those extensions might be reported as invalid.
```
cjschroder, line 196: "&ay; a profile" is missing something, does this mean "when &ay; is given a profile that uses one of those extentions, it might report the profile as invalid"?

cjschroder (changes requested):
thanks! Very much appreciate your attention to the AY Guide, and keeping it updated.

xml/ay_create_control_file.xml
imobachgs: Yes, you are right. Thanks for fixing it.

PR #677, opened 2020-10-06
Backporting Merge of Security/Hardening Guide to SLE15SP2 

taroth21:
### Description

Due to the different setup of the translations in the SLE 15 SP2 maintenance branch it is a bit tricky to backport larger changes that introduce new files. New files need be added as symbolic link for each language in the `l10n` directory (for both `sles` and `sled`).

So far, I added symbolic links for `l10n/sles`. @cjschroder : Could you proceed from here (you can commit to this feature branch) and do the same for `l10n/sled`? Then Travis should no longer complain.


cjschroder:
This wasn't too bad, though I only copied Tanja's symlinks and did not have to create any. First I made a list from the commit:

`git diff-tree --no-commit-id --name-only -r 8ff5145`

It is the same six symlinks copied into 13 directories. Stared at them for some minutes trying to think of a quick script to copy the six symlinks all at once. Decided meh, just copy them and get it done. Created a copy command:

`cp -Pv common_criteria.xml file_management.xml physical_security.xml seccheck.xml software_management.xml user_management.xml  ~/Documents/suse/1github/doc-sle/l10n/sled/cs-cz/xml/`

Repeated it for the next 11 dirs, changing the country code for each one. Took about ten minutes.

cjschroder:
Rats, TravisCI failed because there are still some missing symlinks. I will come back to this tomorrow.

cjschroder:
Holy cow, it worked!

PR #684, opened 2020-10-16
Cgroup accounting

Werkov:
### Description

Clarification of usage and effects of resource accounting via cgroups.
Note I didn't compile the XML and the text may miss some common docs conventions.

* jsc#PM-2229

| Code 15     | Backport Done
| ----------  | ---------- 
| [*] 15 SP3  | (`master` only)   
| [*] 15 SP2  | [ ] 
| [ ] 15 SP1  | [ ] 
| [ ] 15 SP0  | [ ] 

| Code 12     | Backport Done
| ----------  | ---------- 
| [ ] 12 SP5  | [ ] 
| [ ] 12 SP4  | [ ] 
| [ ] 12 SP3  | [ ] 
| [ ] 12 SP2  | [ ]

### DIFF
--- a/xml/tuning_cgroups.xml
+++ b/xml/tuning_cgroups.xml
@@ -57,6 +57,22 @@
   </para>
  </sect1>
 
+ <sect1 xml:id="sec-tuning-cgroups-accounting">
+  <title>Resource accounting</title>
+  <para>
+  The placement of processes into different cgroups can be used to obtain
+  per-cgroup information of certain resource consumptions.
+  </para>
+  <para>
+  The accounting has relatively small but non-zero overhead whose impact
+  depends on the workload.
+  Be aware that turning on accounting for one unit will also implicitly turn it
+  on for all units directly contained in the same slice and for all its parent
+  slices and the units directly contained therein.
+  Therefore the accounting cost is not exclusive to the single unit.
+  </para>
+ </sect1>
+
  <sect1 xml:id="sec-tuning-cgroups-usage">
   <title>Setting Resource Limits</title>
   <note>
--- a/xml/tuning_cgroups.xml
+++ b/xml/tuning_cgroups.xml
@@ -71,6 +71,14 @@
   slices and the units directly contained therein.
   Therefore the accounting cost is not exclusive to the single unit.
   </para>
+  <para>
+  The accounting can be set on per-unit basis with directives such as
+  <literal>MemoryAccounting=</literal> or globally for all units in
+  <filename>/etc/systemd/system.conf</filename> with respective directive
+  <literal>DefaultMemoryAccounting=</literal>.
+  Refer to <literal>systemd.resource-control (5)</literal> for the exhaustive
+  list of possible directives.
+  </para>
  </sect1>
 
  <sect1 xml:id="sec-tuning-cgroups-usage">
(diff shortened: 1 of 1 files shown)


cjschroder:
Thank you @Werkov! I will review this, and don't worry about markup. That is our job :+1:

xml/tuning_cgroups.xml
```diff
@@ -57,6 +57,29 @@
...
  </sect1>
 
+ <sect1 xml:id="sec-tuning-cgroups-accounting">
+  <title>Resource accounting</title>
+  <para>
+  The placement of processes into different cgroups can be used to obtain
+  per-cgroup information of certain resource consumptions.
+  </para>
+  <para>
+  The accounting has relatively small but non-zero overhead whose impact
```
cjschroder, line 67: This is a bit confusing, is this correct? "Be aware that turning on  accounting for one unit will also implicitly turn it on for all units contained in the same
slice and for all its parent slices and the units contained therein." I copied it from man systemd.resource-control.

cjschroder (changes requested):
thanks much, I have just one questions.

xml/tuning_cgroups.xml
Werkov: The quote from the man page is correct (it's not all units in the same slice but just direct descendants of the given slice). That is what I intended to explain in the second sentence of this paragraph. The overhead in the first sentence refers to overhead that exists just because of putting a process in a cgroup (i.e. turning accounting on). Does it clarify what I meant? Should the docs wording be different to express this?

cjschroder: Yes please re-word. The accounting overhead sentence is clear, it's the bit about slices and units that confused me. It is fine to copy the man page, or you might want to say it differently. Thanks!

Werkov: I see. I've force-pushed an updated version.

cjschroder (approved):

xml/tuning_cgroups.xml
```diff
@@ -57,6 +57,30 @@
...
+  depends on the workload.
+  Be aware that turning on accounting for one unit will also implicitly turn it
+  on for all units directly contained in the same slice and for all its parent
+  slices and the units directly contained therein.
+  Therefore the accounting cost is not exclusive to the single unit.
+  </para>
+  <para>
+  The accounting can be set on per-unit basis with directives such as
+  <literal>MemoryAccounting=</literal> or globally for all units in
+  <filename>/etc/systemd/system.conf</filename> with respective directive
```
fbuihuu, line 77: I would refrain from mentioning /etc/systemd/system.conf or any other main config files. Those files can be overridden by drop-ins (wherever they're located) and thus be overridden by the drop-ins we (downstream) ship to customize upstream defaults.

This behavior is documented in systemd documentation but it's quite unexpected by users. I still need to clarify whether this is really intended (see https://github.com/systemd/systemd/pull/17161) and if upstream decides to keep this behavior then we will probably extend our documentation by explaining that customers settings should be done via drop-ins (with higher priority than the one we use).

cjschroder: Thank you @fbuihuu, that is an important clarification. I recall looking at this before, and we need to get it right in the docs. I will review the docs to see if anything needs updating.

PR #686, opened 2020-10-21
Described accessing / partition from emergency shell (bsc#1171581)

tbazant:
### Description

To not lose changes made from dracut emergency shell, you need to mount and chroot the system root file system.

Rendered version attached (section 17.15.2)
[sec-multipath-trouble_color_en.pdf](https://github.com/SUSE/doc-sle/files/5465239/sec-multipath-trouble_color_en.pdf)

* bsc#1171581

| Code 15     | Backport Done
| ----------  | ---------- 
| [x] 15 SP2  | [ ] 
| [x] 15 SP1  | [ ] 
| [x] 15 SP0  | [ ] 

| Code 12     | Backport Done
| ----------  | ---------- 
| [x] 12 SP5  | [ ] 
| [x] 12 SP4  | [ ] 
| [x] 12 SP3  | [ ] 
| [ ] 12 SP2  | [ ]

### DIFF
--- a/xml/storage_multipath.xml
+++ b/xml/storage_multipath.xml
@@ -4643,6 +4643,32 @@ Give root password for maintenance
      </para>
     </listitem>
    </itemizedlist>
+   <important>
+    <para>
+     At this stage, you are working in a <command>dracut</command> emergency
+     shell inside initrd. All modification resulting from the following
+     procedures will be lost at the next reboot. To make them persistent, you
+     need to either:
+    </para>
+    <itemizedlist>
+     <listitem>
+      <para>
+       <command>chroot</command> to the system root <filename>/</filename> file
+       system from the dracut emergency shell. You may need to mount the
+       <filename>/proc</filename>, <filename>/sys</filename>, and
+       <filename>/dev</filename> partitions as well if dracut needs access to
+       them.
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       Use the rescue mode and <command>chroot</command> to the system root
+       <filename>/</filename> file system from there. Find more information
+       about rescue mode in <xref linkend="sec-trouble-data-recover-rescue"/>.
+      </para>
+     </listitem>
+    </itemizedlist>
+   </important>
    <procedure xml:id="pro-multipath-trouble-root-blacklist">
     <title>Emergency Shell: Blacklist File Systems</title>
     <para>
--- a/xml/storage_multipath.xml
+++ b/xml/storage_multipath.xml
@@ -4617,58 +4617,91 @@ Timed out waiting for device dev-disk-by\x2duuid-c4a...cfef77d.device.
 Welcome to emergency shell
 Give root password for maintenance
 (or press Control-D to continue):</screen>
-   <para>
-    This issue occurs in the following situations:
-   </para>
-   <itemizedlist>
-    <listitem>
-     <para>
-      A non-multipath root file system is not blacklisted when multipath is
-      enabled. See
-      <xref linkend="pro-multipath-trouble-root-blacklist"
-       xrefstyle="HeadingOnPage"/>.
-     </para>
-    </listitem>
-    <listitem>
-     <para>
-      On a system with a multipath root file system, multipath has been
-      enabled/disabled without rebuilding the <filename>initrd</filename>. See
-      <xref linkend="pro-multipath-trouble-root-initrd"/>.
-     </para>
-    </listitem>
-    <listitem>
-     <para>
-      Network-storage drivers are missing in the <filename>initrd</filename> on
-      a system with network-attached storage and multipath enabled.
-     </para>
-    </listitem>
-   </itemizedlist>
-   <important>
     <para>
-     At this stage, you are working in a <command>dracut</command> emergency
-     shell from the initrd environment. All modification resulting from the following
-     procedures will be lost at the next reboot. To make them persistent, you
-     need to either:
+     At this stage, you are working in a temporary <command>dracut</command>
+     emergency shell from the initrd environment. To make the configuration
+     changes described below persistent, you need to perform them in the
+     in the environment of the installed system.
     </para>
-    <itemizedlist>
-     <listitem>
+    <important>
+     <para>
+      As a general rule, rebuild <filename>initrd</filename> whenever
+      you make changes to <filename>multipath.conf</filename>.
+     </para>
+    </important>
+    <procedure>
+     <step>
       <para>
-       Mount the system root (<filename>/</filename>) file system and
-       <command>chroot</command> to it from the <command>dracut</command> emergency shell. You may
-       need to mount the <filename>/proc</filename>, <filename>/sys</filename>,
-       and <filename>/dev</filename> partitions as well if dracut needs access
-       to them.
+       Identify what the system root (<filename>/</filename>) file system is.
+       For help, see devices under <filename>/dev/disk/by-uuid</filename>,
+       <filename>/dev/disk/by-id</filename>, or
+       <filename>/dev/disk/by-path</filename>.
       </para>
-     </listitem>
-     <listitem>
+     </step>
+     <step>
       <para>
-       Use the rescue mode and <command>chroot</command> to the system root
-       (<filename>/</filename>) file system from there. Find more information
-       about rescue mode in <xref linkend="sec-trouble-data-recover-rescue"/>.
+       Verify whether the root file system is mounted by using the
+       <command>mount</command> command's output and mount it if needed.
       </para>
-     </listitem>
-    </itemizedlist>
-   </important>
+      <tip>
+       <para>
+        <command>dracut</command> mounts the root file system under
+        <filename>/sysroot</filename> by default.
+       </para>
+      </tip>
+      <para>
+       From now on, let us assume that the root file system is mounted under
+       <filename>/sysroot</filename>.
+      </para>
+     </step>
+     <step>
+      <para>
+       <command>chroot</command> into the <filename>/sysroot</filename>
+       mount point to access the installed system properly. Refer to
+       <xref linkend="sec-trouble-data-recover-rescue-access"/> for more
+       details.
+      </para>
+     </step>
+     <step>
+      <para>
+       Select the right scenario why the system escaped to the emergency shell
+       and follow the related procedure:
+      </para>
+      <itemizedlist>
+       <listitem>
+        <para>
+         A non-multipath root file system is not blacklisted when multipath is
+         enabled. See
+         <xref linkend="pro-multipath-trouble-root-blacklist"
+          xrefstyle="HeadingOnPage"/>.
+        </para>
+       </listitem>
+       <listitem>
+        <para>
+         On a system with a multipath root file system, multipath has been
+         enabled/disabled without rebuilding the <filename>initrd</filename>. See
+         <xref linkend="pro-multipath-trouble-root-initrd"/>.
+        </para>
+       </listitem>
+       <listitem>
+        <para>
+         Network-storage drivers are missing in the <filename>initrd</filename> on
+         a system with network-attached storage and multipath enabled. See
+         <xref linkend="pro-multipath-trouble-root-drivers"/>.
+        </para>
+       </listitem>
+      </itemizedlist>
+     </step>
+     <step>
+      <para>
+       Exit the <command>chroot</command> environment by entering the
+       <command>exit</command> command, then exit the emergency shell and
+       reboot the server by pressing <keycombo><keycap function="control"/>
+       <keycap>D</keycap></keycombo>.
+      </para>
+     </step>
+    </procedure>
+
    <procedure xml:id="pro-multipath-trouble-root-blacklist">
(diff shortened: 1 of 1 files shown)


xml/storage_multipath.xml
```diff
@@ -4643,6 +4643,32 @@ Give root password for maintenance
      </para>
     </listitem>
    </itemizedlist>
+   <important>
+    <para>
+     At this stage, you are working in a <command>dracut</command> emergency
+     shell inside initrd. All modification resulting from the following
```
lproven, line 4649:
```suggestion
     shell from the <filename>initrd</filename>. All modification resulting from the following
```

lproven: Just a proposal. "Inside" the initrd seems an odd way to describe it to me.

xml/storage_multipath.xml
```diff
@@ -4643,6 +4643,32 @@ Give root password for maintenance
...
+     procedures will be lost at the next reboot. To make them persistent, you
+     need to either:
+    </para>
+    <itemizedlist>
+     <listitem>
+      <para>
+       Mount the system root <filename>/</filename> file system and
```
lproven, line 4656:
```suggestion
       Mount the system root (<filename>/</filename>) file system and
```

xml/storage_multipath.xml
```diff
@@ -4643,6 +4643,32 @@ Give root password for maintenance
...
+       <command>chroot</command> to it from the dracut emergency shell. You may
```
lproven, line 4657:
```suggestion
       <command>chroot</command> to it from the <command>dracut</command> emergency shell. You may
```

xml/storage_multipath.xml
```diff
@@ -4643,6 +4643,32 @@ Give root password for maintenance
...
+       need to mount the <filename>/proc</filename>, <filename>/sys</filename>,
+       and <filename>/dev</filename> partitions as well if dracut needs access
+       to them.
+      </para>
+     </listitem>
+     <listitem>
+      <para>
+       Use the rescue mode and <command>chroot</command> to the system root
+       <filename>/</filename> file system from there. Find more information
```
lproven, line 4666:
```suggestion
       (<filename>/</filename>) file system from there. Find more information
```

lproven (changes requested):
I am making suggestions only here -- I do not feel strongly about them and if you wish to reject them, that is fine!

xml/storage_multipath.xml
tbazant, line 4649:
```suggestion
     shell from the initrd environment. All modification resulting from the following
```

tbazant (commented):
accepted & 1 siggestion

xml/storage_multipath.xml
```diff
@@ -4643,6 +4643,32 @@ Give root password for maintenance
...
+    <para>
+     At this stage, you are working in a <command>dracut</command> emergency
+     shell inside initrd. All modification resulting from the following
+     procedures will be lost at the next reboot. To make them persistent, you
+     need to either:
+    </para>
+    <itemizedlist>
+     <listitem>
+      <para>
+       <command>chroot</command> to the system root <filename>/</filename> file
```
hramrach: This does not say how to mount the / filesystem or where it is typically mounted by dracut.
In the emergency environment / is the ramdisk so it needs clarification that the system / is elsewhere.

xml/storage_multipath.xml
```diff
@@ -4643,6 +4643,32 @@ Give root password for maintenance
...
+       system from the dracut emergency shell. You may need to mount the
+       <filename>/proc</filename>, <filename>/sys</filename>, and
+       <filename>/dev</filename> partitions as well if dracut needs access to
```
hramrach: it is not dracut that needs access to these filesystems but the commands that you execute in the chroot. dracut already has them mounted in the ramdisk

xml/storage_multipath.xml
tbazant: > This does not say how to mount the / filesystem or where it is typically mounted by dracut.
> In the emergency environment / is the ramdisk so it needs clarification that the system / is elsewhere.

Jiri Slaby suggested to not `chroot` at all as accessing config files is possible right after mounting. can you agree with that?

hramrach: that depends on what you want to do. for editing a file it might suffice to access is from the dracut shell, for running a command that makes use of the changed file or installing a package you need to chroot. also ramdisk typically does not have an editor

xml/storage_multipath.xml
```diff
@@ -4653,11 +4653,11 @@ Give root password for maintenance
     <itemizedlist>
      <listitem>
       <para>
-       <command>chroot</command> to the system root <filename>/</filename> file
-       system from the dracut emergency shell. You may need to mount the
-       <filename>/proc</filename>, <filename>/sys</filename>, and
-       <filename>/dev</filename> partitions as well if dracut needs access to
-       them.
+       Mount the system root <filename>/</filename> file system and
+       <command>chroot</command> to it from the dracut emergency shell. You may
```
hramrach: are people supposed to just know how to do this?
i have done this a few times but following this guide i would have no idea what to do

xml/storage_multipath.xml
mwilck: I disagree with Jiri's suggestion. In my experience chrooting is necessary in most  cases. It's definitely necessary if the initrd needs to be rebuilt, and I can't think of many scenarios where it would make fixing the problem harder. 
True, if only `/etc/fstab` needs to be edited, it's not necessary. But how do you explain customers to differentiate between cases where **chroot** is necessary, and where it's not?

xml/storage_multipath.xml
mwilck: Why don't you just provide a command line for copy and paste? I always use

     for x in proc sys dev run; do mount --bind /$x /mnt/$x; done

xml/storage_multipath.xml
mwilck: @hramrach, I agree. Imagine a happy user who just accidentally ran into emergency mode for the first time in his life. She needs clear step-by-step instructions.

xml/storage_multipath.xml
```diff
@@ -4643,6 +4643,32 @@ Give root password for maintenance
...
+     need to either:
+    </para>
+    <itemizedlist>
+     <listitem>
+      <para>
+       Mount the system root (<filename>/</filename>) file system and
+       <command>chroot</command> to it from the <command>dracut</command> emergency shell. You may
+       need to mount the <filename>/proc</filename>, <filename>/sys</filename>,
+       and <filename>/dev</filename> partitions as well if dracut needs access
+       to them.
```
mwilck, line 4660: What about `/run` ? Also, more often than not, *after* chroot, it may be necessary to mount additional file systems (subvolumes), like this:

    mount -a -t btrfs

mwilck (changes requested):
Why is this section in `storage_multipath.xml`? I won't deny that multipath issues can cause the need of repair, perhaps more often than other issues. Still, it's certainly not the only possible cause of emergency mode. I believe this section belongs in a general chapter, and the multipath chapter should only link to it.

tbazant:
@mwilck , @hramrach can you please have a look once more and tell me what's missing now ? :-)

xml/storage_multipath.xml
```diff
@@ -4617,32 +4617,91 @@ Timed out waiting for device dev-disk-by\x2duuid-c4a...cfef77d.device.
...
+     At this stage, you are working in a temporary <command>dracut</command>
+     emergency shell from the initrd environment. To make the configuration
+     changes described below persistent, you need to perform them in the
+     in the environment of the installed system.
+    </para>
+    <important>
      <para>
-      Network-storage drivers are missing in the <filename>initrd</filename> on
-      a system with network-attached storage and multipath enabled.
+      As a general rule, rebuild <filename>initrd</filename> whenever
```
mwilck, line 4628: Why mention this here, in the middle of explaining how to chroot?

mwilck: It's absolutely true, but should rather be mentioned where we explain how to make `multipath.conf` changes. (I suppose there's already something like that in the docs).

xml/storage_multipath.xml
```diff
@@ -4617,32 +4617,91 @@ Timed out waiting for device dev-disk-by\x2duuid-c4a...cfef77d.device.
...
-    </listitem>
-   </itemizedlist>
+    </important>
+    <procedure>
+     <step>
+      <para>
+       Identify what the system root (<filename>/</filename>) file system is.
+       For help, see devices under <filename>/dev/disk/by-uuid</filename>,
+       <filename>/dev/disk/by-id</filename>, or
+       <filename>/dev/disk/by-path</filename>.
```
mwilck, line 4638: The root device is set on the kernel command line (`cat /proc/cmdline`). Using that is probably more helpful than searching through `/dev/disk`. In the dracut environment, additional information might be available under `/etc/cmdline.d`.

xml/storage_multipath.xml
```diff
@@ -4617,32 +4617,91 @@ Timed out waiting for device dev-disk-by\x2duuid-c4a...cfef77d.device.
...
+      </para>
+     </step>
+     <step>
+      <para>
+       Verify whether the root file system is mounted by using the
+       <command>mount</command> command's output and mount it if needed.
+      </para>
```
mwilck, line 4645: `systemctl status sysroot.mount` should do the trick. `mount` output is pretty noisy these days because of the many mounted pseudo file systems.

xml/storage_multipath.xml
```diff
@@ -4617,32 +4617,91 @@ Timed out waiting for device dev-disk-by\x2duuid-c4a...cfef77d.device.
...
+     </step>
+     <step>
+      <para>
+       Select the right scenario why the system escaped to the emergency shell
+       and follow the related procedure:
+      </para>
+      <itemizedlist>
+       <listitem>
+        <para>
+         A non-multipath root file system is not blacklisted when multipath is
```
mwilck, line 4673: This is not an error. Blacklisting the root FS is not required. The only requirement is that the blacklisting must be the same in initrd and outside initrd.

xml/storage_multipath.xml
```diff
@@ -4617,32 +4617,91 @@ Timed out waiting for device dev-disk-by\x2duuid-c4a...cfef77d.device.
...
+       </para>
+      </tip>
+      <para>
+       From now on, let us assume that the root file system is mounted under
+       <filename>/sysroot</filename>.
+      </para>
+     </step>
+     <step>
+      <para>
+       <command>chroot</command> into the <filename>/sysroot</filename>
```
mwilck, line 4659: Where is the documentation about the all-important bind mounts? Is it in that general chapter you linked to? I'd give a short command example here. If the user forgets the bind mounts, she won't be able to do much in the chroot.

xml/storage_multipath.xml
```diff
@@ -4617,32 +4617,91 @@ Timed out waiting for device dev-disk-by\x2duuid-c4a...cfef77d.device.
...
+       mount point to access the installed system properly. Refer to
+       <xref linkend="sec-trouble-data-recover-rescue-access"/> for more
+       details.
+      </para>
+     </step>
+     <step>
+      <para>
+       Select the right scenario why the system escaped to the emergency shell
```
mwilck, line 4667: I don't want to over-complicate matters, but this is exactly the problem. **How** would a non-experienced user "select" the right thing here?

Another point, probably to address in a separate / later patch: When the system fails to boot because of multipath issues, it doesn't necessarily happen in dracut's emergency shell. It's actually more likely to happen *after* switching root, when additional file systems or LVM volumes can't be found/activated because of broken multipath configuration.

In either case, the user ends up in systemd's emergency mode. She may have to figure out first if booting failed before or after the root switch. In the latter case, no chrooting is necessary, obviously. Typically you'd check if `/etc/initrd-release` exists, it does so only in the initrd environment (see [systemd documentation](https://systemd.io/INITRD_INTERFACE/)).

xml/storage_multipath.xml
mwilck: I understand you patch didn't add this paragraph. Still, I think it rather causes confusion, because it suggests that using multipath on a single-path root file system is a mistake. It is not. Actually, it's the default.

xml/storage_multipath.xml
```diff
@@ -4617,32 +4617,91 @@ Timed out waiting for device dev-disk-by\x2duuid-c4a...cfef77d.device.
...
+       and follow the related procedure:
+      </para>
+      <itemizedlist>
```
mwilck, line 4670: I'm unsure what this list buys you here. You could simply refer to the following chapters, explaining possible issues and the workarounds for them.

mwilck (changes requested):

lproven (commented):
Since @mwilck has left many comments, I have not further reviewed this until we address those. But from a language/punctuation/etc standpoint, it looks fine.

xml/storage_multipath.xml
tbazant: i removed this guidepost completely, in favour of referencing the procedures directly

tbazant (commented):
few comments and closures

PR #695, opened 2020-11-05
First test of removing unused images files

cjschroder:
Re Trello https://trello.com/c/BWIzK3u0/743-clean-up-unused-files-and-images-in-doc-sle-maintenance-branches

### Description

dap -d DC-FILE list-srcfiles-unused does not provide very useful results, and then you have to do a 3-way diff (DC-SLES-all, DC-SLED-all, DC-opensuse-all) to try to find unused files. I tried a simpler approach: first list all image files, and store this list in a file. Then grep xml/ for all references to images files, then diff the two files.

It should then be possible to delete the image files in the diff, but it seems that .svg and .dia files are interpreted as .png files, so this method still reports some false positives. 

* bsc#...
* jsc#SLE-...

| Code 15     | Backport Done
| ----------  | ---------- 
| [ ] 15 SP3  | (`master` only)   
| [ ] 15 SP2  | [ ] 
| [ ] 15 SP1  | [ ] 
| [ x] 15 SP0  | [ ] 

| Code 12     | Backport Done
| ----------  | ---------- 
| [ ] 12 SP5  | [ ] 
| [ ] 12 SP4  | [ ] 
| [ ] 12 SP3  | [ ] 
| [ ] 12 SP2  | [ ]


cwickert (changes requested):
Beware with the files from the odg_masters directory. 

acls_map_ext.odg is used to create acls_map_ext.png which is referenced in security_acls.xml. 

For the record: There is no acls_map_ext.png but acls_map_ext.svg, not sure what's going on, but please check this do make sure nothing gets deleted accidentally.

cjschroder:
this turned into a mess, closing

PR #713, opened 2020-11-24
Fix table issue in storage_filesystems.xml

tomschr:
### Description

Columns were 4 instead of 2 (missing rows)


dmpop (approved):
Perfect! Thank you very much for your help, Tom!

PR #730, opened 2020-12-17
WIP: Ye olde sentence-case switch

sknorr:
### Description

Henlo i did a [terrible fing nobody asked for](https://github.com/sknorr/daps-migrate-tools/blob/master/sentence-case)

| Code 15     | Backport Done
| ----------  | ---------- 
| [ ] 15 SP2  | [ ] 
| [ ] 15 SP1  | [ ] 
| [ ] 15 SP0  | [ ] 

| Code 12     | Backport Done
| ----------  | ---------- 
| [ ] 12 SP5  | [ ] 
| [ ] 12 SP4  | [ ] 
| [ ] 12 SP3  | [ ] 
| [ ] 12 SP2  | [ ]

### DIFF
--- a/xml/adm_shell.xml
+++ b/xml/adm_shell.xml
@@ -10,7 +10,7 @@
   xmlns:xi="http://www.w3.org/2001/XInclude"
   xmlns:xlink="http://www.w3.org/1999/xlink">
 <!-- Better Title? -->
- <title>Bash and Bash Scripts</title>
+ <title>Bash and Bash scripts</title>
  <info>
   <abstract>
    <para>
@@ -28,7 +28,7 @@
  </info>
 <!-- Better Title? -->
  <sect1 xml:id="sec-adm-whatistheshell">
-  <title>What is <quote>The Shell</quote>?</title>
+  <title>What is <quote>The shell</quote>?</title>
 
 <!--<para>A shell is fundamental to Linux system operation: Its
       interface allows users to enter commands that execute different
@@ -45,7 +45,7 @@
   </para>
 
   <sect2 xml:id="sec-adm-configfiles">
-   <title>Bash Configuration Files</title>
+   <title>Bash configuration files</title>
    <para>
     A shell can be invoked as an:
    </para>
@@ -84,7 +84,7 @@
     configuration files.
    </para>
    <table xml:id="tab-adm-shell-config-loginshells">
-    <title>Bash Configuration Files for Login Shells</title>
+    <title>Bash configuration files for login shells</title>
     <tgroup cols="2">
      <thead>
       <row>
@@ -158,7 +158,7 @@
     <xref linkend="tab-adm-shell-configs-nonloginshells"/>.
    </para>
    <table xml:id="tab-adm-shell-configs-nonloginshells">
-    <title>Bash Configuration Files for Non-Login Shells</title>
+    <title>Bash configuration files for non-login shells</title>
     <tgroup cols="2">
      <tbody>
       <row>
@@ -205,7 +205,7 @@
     Additionally, Bash uses some more files:
    </para>
    <table>
-    <title>Special Files for Bash</title>
+    <title>Special files for Bash</title>
     <tgroup cols="2">
      <thead>
       <row>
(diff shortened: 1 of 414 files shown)


xml/tuning_memory.xml
```diff
@@ -477,7 +477,7 @@
   </sect2>
 
   <sect2 xml:id="cha-tuning-memory-vm-writetiming" os="sles;sled">
-   <title>Timing Differences of I/O Writes between &sle; 12 and &sle; 11</title>
+   <title>Timing differences of i/o writes between &sle; 12 and &sle; 11</title>
```
taroth21, line 480: I/O should not be changed IMHO as it is kind of an abbreviation.

xml/adm_wbem.xml
```diff
@@ -1898,7 +1898,7 @@ localhost:5988/root/cimv2:Linux_Ext4FileSystem FSReservedCapacity=, \
     </listitem>
    </varlistentry>
    <varlistentry>
-    <term><link xlink:href="http://sblim.sourceforge.net/wiki/index.php/Main_Page"/>
+    <term><link xlink:href="http://sblim.sourceforge.net/wiki/index.php/main_Page"/>
```
taroth21, line 1901: Changing case within links may be critical, too. I checked: in this case, the link still works but it took long to load.

xml/adm_shell.xml
```diff
@@ -473,7 +473,7 @@ echo "Hello World" <co xml:id="co-adm-shell-echo"/></screen>
 
   <variablelist>
    <varlistentry>
-    <term><literal>Command &gt; File</literal>
```
taroth21, line 476: Changing from uppercase to lowercase within `<literal/>` tag might be critical (although in this case , it would be better to write the commands in lowercase anyway). Or use `<replaceable/>` (did not check the entire file but I think those are example commands).

xml/deployment_yast_installer.xml
```diff
@@ -2699,7 +2699,7 @@ sle-live-patching 8c541494</screen>
     </para>
    </sect3>
    <sect3 xml:id="sec-s390-inst-reipl-reconnect-x">
-    <title>Using X to Connect</title>
+    <title>Using x to connect</title>
```
taroth21, line 2702: Tricky one to catch, but 'X' should stay in capital letters.

xml/grub2_yast_i.xml
```diff
@@ -171,7 +171,7 @@
      </listitem>
     </varlistentry>
     <varlistentry>
-     <term><guimenu>Enable Trusted Boot Support</guimenu>
```
taroth21, line 174: Even if the actual UI string should be written in title style (which I have not checked), nobody will kill us for changing this, I assume. OTOH, I often also used title style for a UI string in the docs although the software string was written in sentence style. As long as the docs cover what the string actually says, the spelling should not much make a difference to readers.

taroth21 (changes requested):
Only found some minor corner cases, otherwise looks really good to me! Thanks a lot, Stefan!

xml/adm_shell.xml
sknorr: That is fixed.

xml/adm_wbem.xml
sknorr: Good catch, I think that is fixable though.

xml/tuning_memory.xml
sknorr: This one is going to be hard to fix because I defined `/` as word-split character, meaning I/O is treated as `I/` + `O`. I could special-case this particular one if this is a bigger issue.

xml/deployment_yast_installer.xml
sknorr: Wondering whether to handle that via a rule for single-character words or via the word list I already use. Thoughts?

xml/grub2_yast_i.xml
sknorr: The new default is ignoring all tagged content (i.e. no case changes within those). That may not be the right option, e.g. for emphasis. We could define a list of tags that should be converted though. What do you think?

sknorr:
Closing this for the time being, we will revisit the topic in January, with separate PRs, split up by guide.

Feel free to continue to add reviews, however.

xml/deployment_yast_installer.xml
taroth21: Your choice, but as I assume we would not have more single character terms than 'X' adding it to the word list is probably best.

xml/grub2_yast_i.xml
taroth21: Off the top of my head I would vote for changing from title style to sentence style at least within the followings tags:
- citetitle
- guimenu
- emphasis

PR #767, opened 2021-02-16
Deployment Guide -> Searching software online: Rework for usability changes (BSC#1165913)

cwickert:
### Description

yast2 software's 'Online Search' feature has been reworked to address some usability issues. I updated the screenshot and rewrote the text accordingly. 

* https://bugzilla.suse.com/show_bug.cgi?id=1165913

| Code 15     | Backport Done
| ----------  | ---------- 
| [x] 15 SP2  | [ ] 
| [ ] 15 SP1  | [ ] 
| [ ] 15 SP0  | [ ] 

| Code 12     | Backport Done
| ----------  | ---------- 
| [ ] 12 SP5  | [ ] 
| [ ] 12 SP4  | [ ] 
| [ ] 12 SP3  | [ ] 
| [ ] 12 SP2  | [ ]


cwickert:
I'm assigning this to @jfaltenbacher for training purposes.

xml/yast2_sw.xml
```diff
@@ -336,21 +336,20 @@
      </mediaobject>
     </informalfigure>
     <para>
-     To perform a search, just write a term in the <guimenu>Package
-     Name</guimenu> input field and press the <guimenu>Search</guimenu> button.
+     To perform a search, just enter a term in the <guimenu>Package
+      Name</guimenu> input field and press <keycap function="enter"/> or the
+     <guimenu>Search</guimenu> button.
      &yast; then contacts the &scc; and, after a few seconds, shows the results
```
jfaltenbacher, line 342: suggestion to remove "after a few seconds"

xml/yast2_sw.xml
```diff
@@ -336,21 +336,20 @@
...
-     corresponding table row and pressing the <guimenu>Toggle
-     Selection</guimenu> button. Alternatively, you can just double-click on a
-     row. If the package belongs to a not activated module or extension, &yast;
-     asks for confirmation to activate it. Click <guimenu>Next</guimenu>,
+     To install a package, select it and press the <guimenu>Toggle
+      Selection</guimenu> button. Alternatively, you can just double-click on
+     the package. If it belongs to an unregistered module or extension, &yast;
+     asks for confirmation to register it. Packages are marked for installation
+     with <guimenu>&plus;</guimenu> in the <guimenu>Status</guimenu> column.
+     Selected all packages you want to install, click <guimenu>Next</guimenu>,
```
jfaltenbacher, line 352:
```suggestion
     Select all packages you want to install, click <guimenu>Next</guimenu>,
```

xml/yast2_sw.xml
cwickert: "A few seconds" may not be correct, but there is indeed a notable delay because the search is performed online by the SUSE Customer Center. Better ideas? "a few moments"?" "after a short moment"?

xml/yast2_sw.xml
cwickert: Thanks for catching this. I'll fix it when we have the "few seconds" sorted out.

jfaltenbacher (changes requested):
please see my comments

cwickert:
@jfaltenbacher what do you think now?

I made two separate proposals: First commit is with your feedback, some style checks and some simplification. I then turned this into a tip with a procedure in the second commit to get rid of the lone `sect3`. How you you like it?

xml/yast2_sw.xml
```diff
@@ -320,47 +320,49 @@
      repository remain installed.
     </para>
    </tip>
-   <sect3 os="sles;sled">
+   <tip os="sles;sled" xml:id="tip-yast-software-search-online">
```
jfaltenbacher, line 323: Suggestion to have this as plain procedure instead of tip

xml/yast2_sw.xml
```diff
@@ -320,47 +320,49 @@
...
+       Alternatively, you can just double-click on a row. If the package belongs
+       to an unregistered module or extension, &yast; asks for confirmation to
+       register it.
+      </para>
+     </step>
+     <step>
+      <para>
+       Click <guimenu>Next</guimenu>, review the changes, and install the
+       packages.
+      </para>
```
jfaltenbacher, line 362: Suggestion to add result para

xml/yast2_sw.xml
```diff
@@ -320,47 +320,49 @@
...
+       &yast; contacts the &scc; and shows the results in a table, including the
+       module or extension of each package. Select a package to see additional
+       details.
+      </para>
+     </step>
+     <step>
+      <para>
+       Select one or more packages for installation by clicking the
+       corresponding table row and <guimenu>Toggle Selection</guimenu>.
+       Alternatively, you can just double-click on a row. If the package belongs
```
jfaltenbacher, line 353:
```suggestion
       Alternatively, you can double-click on a row. If the package belongs
```

jfaltenbacher (approved):
I like it :) just a few suggestions

xml/yast2_sw.xml
cwickert: The online search needs to be separated from the normal/offline search.

* We need at least a title and a sentence that explains the benefit of the online search (search all modules, even unregistered ones).
* It's SLE-only and needs to be profiled.

Introducing a `<sec3>` would work, but results in a lone section. We could then move the content of the normal/offline search to yet another `<sec3>`, but IHNMO that is too much overhead and too much nesting I think. The online search is an extra by definition, thus I think a `<tip>` is just right.

xml/yast2_sw.xml
cwickert: I see your point, but I left it out on purpose. In step 4 of the procedure I'm already installing a package, but installing and removing packages is covered by the following section. Thus I made the last step as brief. Adding a result ("You have successfully installed a package") would anticipate the next section.

I agree that the structure of the chapter is not optimal. We currently have a section for searching software and anther for installing or removing packages. I think that installing and removing are two separate things, and searching probably belongs to installing. Nevertheless I don't want to rewrite/restructure everything now.

PR #779, opened 2021-03-18
intro about available docs: reorg as discussed in SLE writers sync

taroth21:
### Description

Reorganize and shorten info about available documentation. 

| Code 15     | Backport Done
| ----------  | ----------
| [x] 15 SP3  | *(no backport necessary)*
| [ ] 15 SP2  | [ ]
| [ ] 15 SP1  | [ ]
| [ ] 15 SP0  | [ ]

| Code 12     | Backport Done
| ----------  | ----------
| [ ] 12 SP5  | [ ]
| [ ] 12 SP4  | [ ]
| [ ] 12 SP3  | [ ]
| [ ] 12 SP2  | [ ]

### DIFF
--- a/xml/common_intro_available_doc_i.xml
+++ b/xml/common_intro_available_doc_i.xml
@@ -38,7 +38,8 @@
    <note>
     <title>Latest updates</title>
     <para>
-     The latest documentation updates are usually available in the English version of the documentation.
+     The latest documentation updates are usually available in the English
+     version of the documentation.
     </para>
    </note>
   </listitem>
--- a/xml/common_intro_available_doc_i.xml
+++ b/xml/common_intro_available_doc_i.xml
@@ -20,134 +20,49 @@
   </dm:docmanager>
  </info>
 
- <note>
-  <title>Online documentation and latest updates</title>
-  <para>
-   Documentation for our products is available at
-   <link os="sles;sled" xlink:href="https://documentation.suse.com/"/><link os="osuse" xlink:href="http://doc.opensuse.org/"/>,
-   where you can also find the latest updates, and browse or download the documentation in various formats.
-   The latest documentation updates are usually available in the English version of the documentation.
-  </para>
- </note>
- <note>
-  <title>Manual Pages</title>
-  <para>
-   Many commands are described in detail in their <emphasis>manual
-   pages</emphasis>. You can view manual pages by running the
-   <command>man</command> command followed by a specific command name. If the
-   <command>man</command> command is not installed on your system, install it
-   by running <command>zypper install man</command>.
-  </para>
- </note>
- <para>
-  The following documentation is available for this product:
- </para>
-
- <variablelist>
-  <varlistentry os="sles;sled">
-   <term><xref linkend="art-sle-installquick"/>
-   </term>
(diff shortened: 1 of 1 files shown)


taroth21:
Section would look like this: [avail_docs.pdf](https://github.com/SUSE/doc-sle/files/6165185/avail_docs.pdf)

lproven (approved):

xml/common_intro_available_doc_i.xml
```diff
@@ -20,134 +20,49 @@
...
+<variablelist>
+ <varlistentry>
+  <term>Online documentation</term>
+  <listitem>
+   <para>
+    The online documentation for this product is available at
+<!--taroth 2021-03-18: keep the profiled links in one line,
+    moving them to separate lines would create a whitespace before
+    the full stop-->
+    <link os="sles" xlink:href="https://documentation.suse.com/#sles"/><link os="sled" xlink:href="https://documentation.suse.com/#sled"/><link os="osuse" xlink:href="http://doc.opensuse.org/"/>.
```
sknorr, line 32: Might make sense to use an entity for the URL here -- which could be defined along with &productname; and &productnumber; -- this would allow adding the product version number without too much danger of it later being forgotten. In the release notes, we already use `:doc-url: https://documentation.suse.com/sles/15-SP3` -- so, for consistency, &doc-url; would make most sense to me.

xml/common_intro_available_doc_i.xml
```diff
@@ -20,134 +20,49 @@
...
+    Browse or download the documentation in various formats.
+   </para>
+   <para os="sles;sled">
+    Find the online documentation for other products
+    at <link xlink:href="https://documentation.suse.com/"/>.</para>
+   <note>
+    <title>Latest updates</title>
     <para>
-     &abstract_security;
+     The latest documentation updates are usually available in the English version of the documentation.
```
sknorr, line 41: 1. This line seems long.
2. Maybe add "only" -- or is that too negative-feeling?

sknorr (commented):
Makes sense to me in general, two small suggestion though ...

xml/common_intro_available_doc_i.xml
taroth21: Thanks, I see your point, but as none of the other URLs contains entities for several reasons (outlined here https://github.com/SUSE/doc-sle/pull/579#issue-402627684) I would not want to introduce them now.

xml/common_intro_available_doc_i.xml
taroth21: Good catch!
ad 1. I shortened the line.
ad 2. I stayed away from 'only', because otherwise we need to adjust this again in case we should move to more continuous localizations in the future .

xml/common_intro_available_doc_i.xml
sknorr: Ugh. I had just about forgotten about this kerfuffle. :/ In any case, this would be the entire URL in an entity, not just part of the URL.

The way it is right now, when you click this link, you still need to select a product version, which I think is a click we should ideally save those users, because we know what product version they are looking at currently.

PR #781, opened 2021-03-18
Streamlining the book introductions 

taroth21:
### Description

Streamlining the book introductions. Removed the info about the book structure (as this can be seen from the TOCs), added missing intros to two books and made sure each book has an abstract. 

| Code 15     | Backport Done
| ----------  | ----------
| [x] 15 SP3  | *(no backport necessary)*
| [ ] 15 SP2  | [ ]
| [ ] 15 SP1  | [ ]
| [ ] 15 SP0  | [ ]

| Code 12     | Backport Done
| ----------  | ----------
| [ ] 12 SP5  | [ ]
| [ ] 12 SP4  | [ ]
| [ ] 12 SP3  | [ ]
| [ ] 12 SP2  | [ ]

### DIFF
--- a/xml/book_sles_containers.xml
+++ b/xml/book_sles_containers.xml
@@ -22,6 +22,7 @@
   <xi:include href="common_copyright_gfdl.xml"/>
   <abstract>
    <para>
+    This guide provides an introduction to the &suse; container ecosystem.
     This document is a work in progress. The content in this document is
     subject to change without notice.
    </para>
--- a/xml/ay_intro.xml
+++ b/xml/ay_intro.xml
@@ -1,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE preface
+[
+  <!ENTITY % entities SYSTEM "entity-decl.ent">
+    %entities;
+]>
+
+<preface version="5.0" xml:id="ay-preface"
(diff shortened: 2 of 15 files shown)


xml/book_sles_containers.xml
```diff
@@ -22,6 +22,7 @@
   <xi:include href="common_copyright_gfdl.xml"/>
   <abstract>
    <para>
+    This guide provides an introduction to the &suse; container ecosystem.
```
lproven, line 25: I do not like this word -- it sounds too much like marketing to me. However, any replacement I can think of would be a whole sentence.

lproven (commented):
I think I am missing something here -- I see a lot of deleted text, but nothing to replace it. Is this a WIP or does it pair up with one of the other PRs?

taroth21:
> I think I am missing something here -- I see a lot of deleted text, but nothing to replace it. Is this a WIP or does it pair up with one of the other PRs?

No, I really removed a lot of text about the book structure, as the book structure (which parts etc.) is visible from the TOC.

xml/book_sles_containers.xml
taroth21: I just moved over to the abstract what was already written elsewhere (before the sentence was in container_intro.xml). I think we have another chance to fix it when the work on the guide continues.

PR #795, opened 2021-03-24
SEO Improvements (book names, DC files, rootids) 

taroth21:
### Description
@sknorr suggested some changes to improve the SEO:

- remove `-sle*` from rootids
- make DC files more self-explanatory (use full English words, connected by hyphens) and consistent

I also removed `sle` from the XML filenames for the books (this may not be necessary and may be an additional pain for backporting, let me know if to revert this). I did not rename the articles files yet for this reason.

With the changes of the rootids some xrefs and the parameters in the DC files needed to be adjusted.

The DC filename changes also need to be reflected on out daily snapshot servers and the configuration for d.s.c. 

| Code 15     | Backport Done
| ----------  | ----------
| [ x] 15 SP3  | *(no backport necessary)*
| [ ] 15 SP2  | [ ]
| [ ] 15 SP1  | [ ]
| [ ] 15 SP0  | [ ]

| Code 12     | Backport Done
| ----------  | ----------
| [ ] 12 SP5  | [ ]
| [ ] 12 SP4  | [ ]
| [ ] 12 SP3  | [ ]
| [ ] 12 SP2  | [ ]


DC-SLED-install-quick
```diff
@@ -5,7 +5,7 @@
 ##
 ## Basics
 MAIN="MAIN.SLEDS.xml"
-ROOTID=art-sle-installquick
+ROOTID=art-installquick
```
sknorr, line 8:
```suggestion
ROOTID=art-install-quick
```

DC-SLES-virtualization-best-practices
```diff
@@ -5,7 +5,7 @@
 ##
 ## Basics
 MAIN="MAIN.SLEDS.xml"
-ROOTID=article-vt-best-practices
+ROOTID=article-virtualization
```
sknorr, line 8: This one is inconsistent with the others, which are named just `art-...` not `article-...`. However, from the standpoint of using complete English words, I like this version better.

xml/art_sles_rpi_quick.xml
```diff
@@ -13,7 +13,7 @@
 <article role="sbp" xmlns="http://docbook.org/ns/docbook"
   xmlns:xi="http://www.w3.org/2001/XInclude"
   xmlns:xlink="http://www.w3.org/1999/xlink" version="5.0"
-  xml:id="art-rpiquick" xml:lang="en">
+  xml:id="art-raspi" xml:lang="en">
```
sknorr, line 16: The late "Raspbian OS" (now Raspberry Pi OS) had an additional `b` between p and i.
I think the "was there a b in SUSE's particular abbreviation or not?" issue could nicely be sidestepped by using the full product name:
```suggestion
  xml:id="art-raspberry-pi" xml:lang="en">
```

xml/book_deployment.xml
```diff
@@ -12,7 +12,7 @@
 <!-- SLE Deployment Guide                                                  -->
 <!-- ##################################################################### -->
 <?provo dirname="deployment/"?>
-<book version="5.0" xml:lang="en" xml:id="book-sle-deployment"
+<book version="5.0" xml:lang="en" xml:id="book-deployment"
```
sknorr, line 15: Maybe make this consistent with book-install and book-upgrade:

```suggestion
<book version="5.0" xml:lang="en" xml:id="book-deploy"
```

xml/book_admin.xml
```diff
@@ -12,7 +12,7 @@
 <!-- Administration Guide                                                  -->
 <!-- ##################################################################### -->
 <?provo dirname="admin/"?>
-<book version="5.0" xml:lang="en" xml:id="book-sle-admin"
+<book version="5.0" xml:lang="en" xml:id="book-admin"
```
sknorr, line 15: A complete word:

```suggestion
<book version="5.0" xml:lang="en" xml:id="book-administration"
```

(The suggestion is not consistent with install/upgrade now, because we use the noun and not the verb here. But `book-administrate` sounds off.)

sknorr (commented):
Thanks! A few comments enclosed.

I wonder whether we should abolish capital letters in DC file names as well (i.e., rename to `DC-sles-upgrade`). That would probably be irrelevant for SEO, but might make our URLs less painful to type.

DC-SLES-virtualization-best-practices
taroth21: I was wondering whether to stick with 'art' or move to 'article', too. Probably will do the latter.

sknorr:
Fwiw, the CI fails right now, because `.travis-check-docs` contains mentions of some DC files that have been renamed. I almost wanted to update this but decided against it, since I am not sure you are done here.

PR #864, opened 2021-05-24
Update AY Guide for SLE Micro with additional profiling

cjschroder:
re BSC# 1185835, profiled second stage options not supported
in SLE Micro. No AY second-stage options are supported.

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP4/openSUSE Leap 15.4 *(if applicable to 15 SP4 _only_ -- do not merge yet!)*
  - [x] SLE 15 SP3/openSUSE Leap 15.3 *(current `main`, no backport necessary)*
  - [ ] SLE 15 SP1
  - [ ] SLE 15 GA
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3

### DIFF
--- a/xml/ay_auto_installation.xml
+++ b/xml/ay_auto_installation.xml
@@ -166,7 +166,7 @@ xml:id="Invoking">
      If you install via PXE, the installation will run in an endless
      loop. This happens because after the first reboot, the machine performs
      the PXE boot again and restarts the installation instead of booting from
-     the hard disk for the second stage of the installation.
+     the hard disk<phrase os="sles;sled;osuse"> for the second stage of the installation</phrase>.
     </para>
     <para>
      There are several ways to solve this problem. You can use an HTTP
@@ -811,7 +811,7 @@ textmode: 1
     </para>
    </sect2>
   </sect1>
-  <sect1 xml:id="System-Configuration">
+  <sect1 os="sles;sled;osuse" xml:id="System-Configuration">
    <title>System configuration</title>
 
    <para>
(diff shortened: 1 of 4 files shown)


imobachgs:
Thanks Carla! In general, it looks good. However, I have spotted other places to change.

* **9.2.3 Booting via PXE over the network** mentions the second stage. I think that just removing the last part of the paragraph should be enough:

*If you install via PXE, the installation will run in an endless loop. This happens because after the first reboot, the machine performs the PXE boot again and restarts the installation instead of booting from the hard disk ~~for the second stage of the installation~~.*

* The `halt` element described in the **4.1.1 The mode section** is basically the equivalent to `final_halt`, as there is no 2nd stage.

Additionally, I think we should remove the following sections:

* **9.4.2 System customization**. Perhaps 9.4.1 content could be included directly in 9.4.
* **4.1.6 The wait section**

imobachgs:
>     * The `halt` element described in the **4.1.1 The mode section** is basically the equivalent to `final_halt`, as there is no 2nd stage.

I have checked the AutoYaST code and I would say that we should drop the following elements: `final_halt`, `final_reboot`, `final_restart_services`, `activate_systemd_default_target` and `max_systemd_wait`. All those elements are processed during the 2nd stage, so they are basically ignored in MicroOS.

However, we should keep `halt` but removing the reference to the *second stage*. Something like: *Instead of rebooting ~~into stage two~~, the machine is turned off.*

imobachgs (changes requested):
Sorry, I forgot to post my comments as a review. I hope it helps.

cjschroder:
Thanks @imobachgs! I believe @lvicoun is going to finish this PR, as she maintains the SLE Micro guide.

lvicoun:
Reviewers comments applied.

PR #893, opened 2021-06-15
Kernel live patching: Replace KLP image with DITAA version

sknorr:
### PR creator: Description

This replaces the PNG version and the incompatible SVG version of the Kernel Live Patching image with a DITAA version.
What it looks like when rendered:

DITAA

This is partly an attempt to see how editing works. (I guess Emacs org mode would have helped. I also missed block selection in my editor.)

I do agree that it does not quite have the graphical fidelity of the original version:
original PNG

* jsc#DOCTEAM-191

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP4/openSUSE Leap 15.4 *(if applicable to 15 SP4 _only_ -- do not merge yet!)*
  - [ ] SLE 15 SP3/openSUSE Leap 15.3 *(current `main`, no backport necessary)*
  - [ ] SLE 15 SP1
  - [ ] SLE 15 GA
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3


lproven (approved):

dmpop (changes requested):
Two questions:

1. Are you going to redo _all_ diagrams made with draw.io? Because if you are not, then what's the point?
2. Were there any discussions about using DITAA for diagramming?

sknorr:
> 1. Are you going to redo all diagrams made with draw.io? Because if you are not, then what's the point?

The point was for me to actually create a DITAA image. The other point was to get you (and others) to look at DITAA and see whether it makes sense to you.
I was not going to re-create all diagrams. At least not right at this point.


> 2. Were there any discussions about using DITAA for diagramming?

Essentially, this is the discussion. Or the start of a discussion.

dmpop:
I mean, if you want us to take a look at DITAA and start a discussion, filing a PR request might not be the best course of action. Bring it up an weekly call?

sknorr:
Ok, whatever. I don't need this, I guess.

janajaeger:
ditaa seems perfect for flowcharts and the likes, but there are a few drawbacks which we must consider:
- no option to change the default font - there seems to be a patch somewhere to support just that
- arrow heads default to a design not compatible with (current) SUSE style (proposed fix: ignore and live with it?)
- line color defaults to black - we need an option to colorize lines (just using dashes instead doesn't cut it)

But a PR is not the right place to discuss this. Will bring it up in one of the next meetings.

PR #922, opened 2021-07-13
Maintenance/sle15 sp3

PaulTheScott:
I would like contribute towards pdf and HTML Documentation.


sknorr:
Hi! As a PR, this seems rather unhelpful, sorry to say. However, if you have a question on contributing: https://github.com/SUSE/doc-sle/discussions/categories/q-a

PR #938, opened 2021-08-16
Improve style of LDAP section

sknorr:
### PR creator: Description

Hi @cjschroder -- this PR tries to correct a few things within the LDAP chapter. I've ultimately refrained from making this exhaustive though (because time). I would like to make sure this PR doesn't break any eggs to make the omelette though.

A few things in the chapter are imo inconsistent with SUSE style:

* Collecting multiple commands within a screen is something that I don't think is normally done in SUSE docs. I've tried to reduce those occurrences but I didn't want to rewrite the entire chapter. (Afaik this is not an explicit style rule [sadly].)
* Using things that look like actual values rather than a generic placeholder within replaceables (such as `ldap1` rather than `INSTANCE_NAME` in this chapter). (While the capitalization is a style rule, the rule to use a generic placeholder is [sadly] implicit only.)
* Using "we" to mean roughly "the writer + the reader". Either don't address the reader or use "you".
  "We" means "SUSE", as in "we recommend" = "official SUSE recommendation". (This one is an explicit style rule.)
* The source tends to be formatted badly and include many trailing spaces. `daps-xmlformat` / `daps xmlformat` can resolve both issues.

- SLE 15/openSUSE Leap 15.x
  - [x] ? SLE 15 SP2/openSUSE Leap 15.2
  - [ ] SLE 15 SP1
  - [ ] SLE 15 GA
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3

### DIFF
--- a/xml/entity-decl.ent
+++ b/xml/entity-decl.ent
@@ -322,6 +322,7 @@ they also have a trademark policy, let's be careful here. -->
 <!ENTITY gpg           "GPG"> <!-- or GnuGP -->
 <!ENTITY clamav        "ClamAV">
 <!ENTITY ds389         "389 Directory Server">
+<!ENTITY ds389a        "389 DS">
 <!-- oddities -->
 <!ENTITY thrdmrk       "*">
 <!ENTITY kexec          "Kexec">
--- a/xml/security_ldap.xml
+++ b/xml/security_ldap.xml
@@ -11,9 +11,9 @@
         <para>
     The Lightweight Directory Access Protocol (LDAP) is a protocol
     designed to access and maintain information directories. LDAP can be
-    used for tasks such as user and group management, system configuration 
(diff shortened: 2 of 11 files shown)


cjschroder:
Thank you @sknorr! I found a few more small errors, and will fix them in a separate commit. Merging!

sknorr:
Thanks for merging! (Though you could have used this branch, I guess.)

cjschroder:
Not sure if this should be backported to SP2, I will will find out.

sknorr:
Probably not. I checked the branch at some point and I think your original LDAP changes haven't gone into SP2 either. Sorry for the doing the weird `[x]? ` next 15 SP2.

PR #1012, opened 2021-11-12
SEO: remove editurls_trans_SLE15SP1

jfaltenbacher:
### PR creator: Description

dm:editurls tag removed for all languages of SLE15 SP1.

Will be done for each branch version separately therefore no backports needed


sknorr:
Thank you!

PR #1014, opened 2021-11-15
SEO: remove editurls from l10n SLE15SP2

jfaltenbacher:
### PR creator: Description

dm:editurls tag removed for all languages of SLE 15 SP2.

Will be done for each branch version separately therefore no backports needed


sknorr:
Thank you! Squashed into one commit, 03a398e, merged.

PR #1015, opened 2021-11-15
aligned prompts to sudo an added recommendations for ARM64 arcjitecture

tbazant:
### PR creator: Description

Aligned prompts to use `sudo` + make preconditions to run qemu on ARM64 architecture more clear.
Rendered PDF is attached 
[cha-qemu-running-basic_color_en.pdf](https://github.com/SUSE/doc-sle/files/7562656/cha-qemu-running-basic_color_en.pdf)

- SLE 15/openSUSE Leap 15.x
  - [x] SLE 15 SP1
  - [x] SLE 15 GA
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3

### DIFF
--- a/xml/qemu_running_vms_qemukvm.xml
+++ b/xml/qemu_running_vms_qemukvm.xml
@@ -53,10 +53,37 @@
   <tip>
    <title>&aarch64; architecture</title>
    <para>
-    Running &qemu; on the &aarch64; architecture requires you to specify
-    a firmware image file with the <option>-bios</option> option. For example:
+    &kvm; support is available only for 64-bit ARM architecture (&aarch64;).
+    Running &qemu; on the &aarch64; architecture requires you to specify:
    </para>
-<screen>&prompt.user;qemu-system-arm [...] -bios /usr/share/qemu/qemu-uefi-aarch64.bin</screen>
+   <itemizedlist>
+    <listitem>
+     <para>
+      A firmware image file using the <option>-bios</option> option.
+     </para>
+    </listitem>
+    <listitem>
+     <para>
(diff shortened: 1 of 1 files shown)


tbazant:
comment from Slack:
"...the part "A CPU type with the same features as the host machine using the
      <option>-cpu</option> option." should actually be just "-cpu host". It is not about matching the same features using another cpu model, really only the "host" cpu  (passthrough) works. I am not sure that -bios is actually mandatory

liayan:
AArch64 is actually an ISA term here. There are aarch32 and aarch64 isa sets that could be run on armv8  CPU architecture.
In qemu, there are binary files: qemu_system-arm and qemu_system-aarch64. We only support qemu_system-aarch64.
Btw, we could run both aarch32 and aarch64 ISA.
 
For qemu-system-aarch64, "-machine" option is a must, there is no default for it.
default of cpu is actually cortex-15, however, we recommend "-cpu host"
Another option that needs to be pointed out here is "gic-version", the default is "2", and we suggest "host" instead.

liayan:
Another place is the graphic device, qemu_system-aarch64 only supports the virtio-gpu-pci device

liayan:
Nice work! A few small things are below:

1. Could you move "machine type" to the first line, there is no default value for it, this is a mandatory option for user.
2. Could you mention that "host" is recommand option whilt it has cortex-15 as default option.
3. same for GIC, it has 2 as default, but host is recommand.
4. for graphic device, could you metion "When grphic mode is needed, choose device type virtio-gpu-pci", because user could also use serial termial with "-nographic".

PR #1056, opened 2022-01-15
remove misleading or no longer valid example

crrodriguez:
The documentation says make -j NO_CPUS will benefit from this setting  off.. this may have been the case 7 or 8 years ago when it was added, but in 2016 make was changed to use vfork everywhere.. later make was modified to execute jobs using posix_spawn. both ways are always child_first (the calling process is suspended) 
So this example is at least out of date or at most never quite true.


dmpop:
Thank you very much for creating the PR!

PR #1074, opened 2022-02-03
Clarify and expand description of AY GRUB 2 password option 

cjschroder:
Try to be more specific about the use case, and what it does

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP1
  - [ ] SLE 15 GA
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3

### DIFF
--- a/xml/ay_bootloader.xml
+++ b/xml/ay_bootloader.xml
@@ -264,13 +264,24 @@
     <term>password</term>
     <listitem>
      <para>
-      If defined, then it protects boot via password. It has three subelements.
-      <literal>value</literal> holds password value. It can be either plain text
-      or already encrypted one. If <literal>encrypted</literal> is set to <literal>
-      true</literal> it means that <literal>value</literal> is already encrypted
-      password. <literal>unrestricted</literal> is to set unrestricted option for
-      password protection. <phrase os="sles;sled">For more information on boot passwords,
-      see <xref linkend="vle-grub2-yast2-boot-password"/>.</phrase>
+      If this is defined, it protects the boot loader with a password. The
+      system will not boot until the password is entered. You may define
+      specific users who are allowed to boot the system, or allow any user
+      who has the password to boot the system.
+    </para>
+    <para>
+      It has three subelements: <literal>value</literal>, 
+      <literal>encrypted</literal>, and <literal>unrestricted</literal>.
+    </para>
+    <para>
+      <literal>value</literal> holds the password value. It can be either 
+      plain text, or an already encrypted password. Set <literal>encrypted</literal> 
+      to <literal>true</literal> when you use an already encrypted
+      password. 
+    </para>
+    <para><literal>unrestricted</literal> allows any user who has the 
+      password to boot the system. <phrase os="sles;sled">For more 
+      information on managing boot passwords, see <xref linkend="vle-grub2-yast2-boot-password"/>.</phrase>
      </para>
 <screen>&lt;password&gt;&lt;value&gt;my_strong_password&lt;/value&gt;&lt;encrypted&gt;false&lt;/encrypted&gt;&lt;unrestricted&gt;false&lt;/unrestricted&gt;&lt;/password&gt;</screen>
     </listitem>
--- a/xml/ay_bootloader.xml
+++ b/xml/ay_bootloader.xml
@@ -265,9 +265,7 @@
     <listitem>
      <para>
       If this is defined, it protects the boot loader with a password. The
-      system will not boot until the password is entered. You may define
-      specific users who are allowed to boot the system, or allow any user
-      who has the password to boot the system.
+      system will not boot until the password is entered.
     </para>
     <para>
       It has three subelements: <literal>value</literal>, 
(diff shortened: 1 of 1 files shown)


JRivrain:
Thanks, the part about encrypted password is more clear. Regarding the "restricted" option, 
image
Yast bootloader only allows root to unlock grub, but that could be done manually by editing /etc/grub.d/42_password (not tested, and probably not recommended as the edited file could be overwritten by Yast)

cjschroder:
Thank you @JRivrain. I'm not clear on the `<unrestricted/>` option, what happens if the admin omits it?

cjschroder:
> Thank you @JRivrain. I'm not clear on the `<unrestricted/>` option, what happens if the admin omits it?

If I am reading https://www.gnu.org/software/grub/manual/grub/grub.html#menuentry correctly, the `<unrestricted/>` option allows everyone who knows the boot password to boot the system, and edit the boot menu from the GRUB command line. Omitting the `<unrestricted/>` option means only root can boot the system. Is that correct?

cjschroder:
> > Thank you @JRivrain. I'm not clear on the `<unrestricted/>` option, what happens if the admin omits it?
> 
> If I am reading https://www.gnu.org/software/grub/manual/grub/grub.html#menuentry correctly, the `<unrestricted/>` option allows everyone who knows the boot password to boot the system, and edit the boot menu from the GRUB command line. Omitting the `<unrestricted/>` option means only root can boot the system. Is that correct?

No wait, that does not make sense. Omitting the `<unrestricted/>` option means that anyone with the boot password can boot the system, but not edit GRUB menu entires. Is that correct?

JRivrain:
> > > Thank you @JRivrain. I'm not clear on the `<unrestricted/>` option, what happens if the admin omits it?
> > 
> > 
> > If I am reading https://www.gnu.org/software/grub/manual/grub/grub.html#menuentry correctly, the `<unrestricted/>` option allows everyone who knows the boot password to boot the system, and edit the boot menu from the GRUB command line. Omitting the `<unrestricted/>` option means only root can boot the system. Is that correct?
> 
> No wait, that does not make sense. Omitting the `<unrestricted/>` option means that anyone with the boot password can boot the system, but not edit GRUB menu entires. Is that correct?

I have not tested that. I guess if you omit it, it will default to unrestricted=true, so the reply would be yes. But I'll try it and confirm that, also I can try omitting the other fields to see what happens.

JRivrain:
So, if we omit `<unrestricted>`, it defaults to true (users can boot without password but cannot edit grub menu). If we omit `<encrypted>`, it defaults no false (the password in `<value>` must be not encrypted, and will be encrypted by Yast).
Here is the resulting /etc/grub.d/42_password file if we omit both `<unrestricted>` and `<encryypted>`:
```
#! /bin/sh
exec tail -n +3 $0
# File created by YaST and next YaST run probably overwrite it
set superusers="root"
password_pbkdf2 root grub.pbkdf2.sha512.10000.190B0C88A08A83BE826674B4816E206CF112EEC108C998CC39B60C135C861081D6B7F312B50AD29CFA921F08FA2D533AC98A54D68C112851325FBA042AB93C45.2A777ED857955305E38988299E57A3EF3A33BF11309AAA6CA1B50C2D80C0587C8026CA262D5D4C1E486089F720EA9530C26D5D171E1B3310B04C6987412A4A28
export superusers
set unrestricted_menu="y"
export unrestricted_menu
```

cjschroder:
> So, if we omit `<unrestricted>`, it defaults to true (users can boot without password but cannot edit grub menu). If we omit `<encrypted>`, it defaults no false (the password in `<value>` must be not encrypted, and will be encrypted by Yast). Here is the resulting /etc/grub.d/42_password file if we omit both `<unrestricted>` and `<encryypted>`:
> 
> ```
> #! /bin/sh
> exec tail -n +3 $0
> # File created by YaST and next YaST run probably overwrite it
> set superusers="root"
> password_pbkdf2 root grub.pbkdf2.sha512.10000.190B0C88A08A83BE826674B4816E206CF112EEC108C998CC39B60C135C861081D6B7F312B50AD29CFA921F08FA2D533AC98A54D68C112851325FBA042AB93C45.2A777ED857955305E38988299E57A3EF3A33BF11309AAA6CA1B50C2D80C0587C8026CA262D5D4C1E486089F720EA9530C26D5D171E1B3310B04C6987412A4A28
> export superusers
> set unrestricted_menu="y"
> export unrestricted_menu
> ```

Excellent, thank you!

xml/ay_bootloader.xml
```diff
@@ -264,13 +264,27 @@
...
+    <para>
+      It has three subelements: <literal>value</literal>, 
+      <literal>encrypted</literal>, and <literal>unrestricted</literal>.
+    </para>
+    <para>
+      <literal>value</literal> holds the password. It can be either 
+      plain text, which &yast; will encrypt, or an already encrypted 
+      password. Set <literal>encrypted</literal> to <literal>true</literal> when you use an already encrypted password. 
+    </para>
+    <para>When <literal>unrestricted</literal> is set to
```
JRivrain, line 279: Hi, sorry I was absent for a while, I could not review your changes. This is not completely correct. It would be:

"when unrestricted is set to "false", users need the password defined as <value> in order to boot or edit grub menu entries. When it is set to "true", users can boot the system without a password, but need a password to edit grub menu entries.
If the option is omitted, it defaults to "true".

For the encryption thing, I think we could make it more precise, as customers will likely have no idea what we mean when we say they can use an "already encrypted password". It took me a while to figure it out myself. I would suggest this:

 "value" holds the password. It can be either 
 plain text, which &yast; will encrypt, _or a password already encrypted using grub-mkpasswd-pbkdf2_.
 Set <literal>encrypted</literal> to <literal>true</literal> when you use an already encrypted password. 
If the option is omitted, it defaults to "false".

cjschroder:
I opened a new PR, will leave it open a couple days in case anyone wants to comment.
https://github.com/SUSE/doc-sle/pull/1086

xml/ay_bootloader.xml
cjschroder: excellent, thank you!

JRivrain:
> I opened a new PR, will leave it open a couple days in case anyone wants to comment. #1086

I'll have a look today

PR #1093, opened 2022-02-24
DOCTEAM-409 contribution para changed

jfaltenbacher:
### PR creator: Description

Introduction para for contributing to documentation for foreign languages changed. As 'Edit source' is no longer available for other languages then English.
Changes also need to be done in the other product repos.

* jsc#DOCTEAM-409

- SLE 15/openSUSE Leap 15.x
  - [x] SLE 15 SP1
  - [x] SLE 15 GA
- SLE 12
  - [x] SLE 12 SP5
  - [x] SLE 12 SP4
  - [x] SLE 12 SP3
- SLE 11
  - [x] SLE 11 SP4

### DIFF
--- a/xml/common_intro_feedback_i.xml
+++ b/xml/common_intro_feedback_i.xml
@@ -91,26 +91,6 @@
     </para>
    </listitem>
   </varlistentry>
-  <varlistentry condition="beta">
-   <term>Beta program requests</term>
-   <listitem>
-    <para>
-     &suse; Beta Software is bound by the &suse; Beta EULA.
-     This EULA is shipped with the software and can also be read
-     at <link xlink:href="https://documentation.suse.com/beta/eula"/>.
-     &suse; can only provide support via the Beta Program Channel.
-    </para>
-    <para>
-     To open a bug report, follow the instructions on the appropriate
-     product page under <link xlink:href="https://suse.com/betaprogram/beta/"/>.
-     For enhancement requests or any other inquiries, contact us via the
-     product-specific public beta mailing lists. They are listed at
-     <link xlink:href="https://suse.com/betaprogram/beta/"/>.
-     For private requests, contact us directly via
-     <link xlink:href="mailto:beta-programs@lists.suse.com"/>.
-    </para>
-   </listitem>
-  </varlistentry>
   <varlistentry os="osuse">
    <term>Help</term>
    <listitem>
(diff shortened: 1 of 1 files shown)


jfaltenbacher:
Hi @taroth21: if you don't find the time for this, I can ask someone else in the team. I have selected backporting for all versions, unless you disagree with some of the older versions. Just let me know

jfaltenbacher:
needs to be decided in SLE writers call

cwickert:
Was the "Edit source" link actually removed from translated docs or did it never exist there? I guess it was the latter.

Also, this should probably be handled through doc-kit, right?

jfaltenbacher:
Hi @cwickert: the "Edit source" link existed and got removed around Nov/Dec last year by Stefan after we had encountered that it does not really work for the languages.

xml/common_intro_feedback_i.xml
```diff
@@ -75,11 +83,31 @@
...
-     the URL) and provide a concise description of the problem.
+     You can also report errors and send feedback concerning the
+     documentation to <email>doc-team@suse.com</email>. Include the
+     document title, the product version, and the publication date of the
+     document. Additionally, include the relevant section number and title (or
+     provide the URL) and provide a concise description of the problem.
+    </para>
+   </listitem>
+  </varlistentry>
+  <varlistentry condition="beta">
```
taroth21, line 94: I'm not sure this profiling condition would work out of the box, let me check tomorrow.

xml/common_intro_feedback_i.xml
```diff
@@ -15,22 +15,23 @@
  </info>
 
  <para>
-  Your feedback and contributions to this documentation are welcome! Several
```
taroth21, line 18: To be honest, I think the old sentence was easier to read than the new one with the insertions between the commas.

jfaltenbacher: fixed as suggested with minor improvement of 2nd sentence

xml/common_intro_feedback_i.xml
taroth21: The text appears but as we don't use the related profiling parameter in the DC file, it will always appear, irrelevant if the beta phase is running or not. I would therefore remove it from here and keep it in doc-kit.

taroth21 (changes requested):
Thanks, Julia! I would remove the varlistentry about the Beta (see my comment), otherwise it's fine.

taroth21:
Backported the changes to all needed maintenance branches (see commits above).

PR #1150, opened 2022-04-01
Tuning Guide: Rework section about IO control with cgroups

Werkov:
### PR creator: Description

Very rough draft to capture:
- simple(r) IO control setup with systemd,
- prerequisites of such setups and mainly
- expectations when configuring IO control.

### PR creator: Are there any relevant issues/feature requests?

* bsc#1195915
* bsc#1175160
* jsc#SLE-22013 (loosely related)

### PR creator: Which product versions do the changes apply to?

When opening a PR, check all versions of the documentation that your PR applies to.

- SLE 15/openSUSE Leap 15.x
  - [x] SLE 15 SP4/openSUSE Leap 15.4 *(current `main`, no backport necessary)*
  - [ ] SLE 15 SP3/openSUSE Leap 15.3
  - [ ] SLE 15 SP2/openSUSE Leap 15.2
  - [ ] SLE 15 SP1
  - [ ] SLE 15 GA
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3

It also assumes systemd version >v243 (and ideally needs some more
backports both into systemd and kernel to work good enough in described
scenarios).

### PR reviewer only: Have all backports been applied?

The doc team member merging your PR will take care of backporting to older documents.
When opening a PR, do *not* set the following check box.

- [ ] all necessary backports are done

PR #1154, opened 2022-04-07
Tuning Guide: Rework section about IO control with cgroups

Werkov:
### PR creator: Description

Very rough draft to capture (replacement PR of #1150):
- simple(r) IO control setup with systemd,
- prerequisites of such setups and mainly
- expectations when configuring IO control.

* bsc#1195915
* bsc#1175160
* jsc#SLE-22013 (loosely related)

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP1
  - [ ] SLE 15 GA
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3

It also assumes systemd version >v243 (and ideally needs some more
backports both into systemd and kernel to work good enough in described
scenarios).

### DIFF
--- a/xml/tuning_cgroups.xml
+++ b/xml/tuning_cgroups.xml
@@ -3,7 +3,7 @@
 [
   <!ENTITY % entities SYSTEM "generic-entities.ent">
   %entities;
-  <!-- custom prompt entities for the 'Controlling I/O with Proportional Weight Policy' section --> 
+  <!-- custom prompt entities for the 'Controlling I/O with Proportional Weight Policy' section -->
   <!ENTITY reader1 "reader1:/io-cgroup #">
   <!ENTITY prompt.reader1 "<prompt role='root' xmlns='http://docbook.org/ns/docbook'>[reader1] host1:/io-cgroup # </prompt>">
   <!ENTITY prompt.reader2 "<prompt role='root' xmlns='http://docbook.org/ns/docbook'>[reader2] host1:/io-cgroup # </prompt>">
@@ -217,7 +217,7 @@ DefaultTasksMax=256
     following example demonstrates how to raise MariaDB's limit to 8192.
    </para>
 <screen>&prompt.sudo;<command>systemctl set-property mariadb.service TasksMax=8192</command>
-&prompt.user;<command>systemctl status mariadb.service</command> 
+&prompt.user;<command>systemctl status mariadb.service</command>
 ● mariadb.service - MariaDB database server
    Loaded: loaded (/usr/lib/systemd/system/mariadb.service; disabled; vendor preset: disab>
   Drop-In: /etc/systemd/system/mariadb.service.d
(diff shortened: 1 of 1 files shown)


cjschroder:
Here is a PDF build --
[sec-control-io-cgroups_color_draft_en.pdf](https://github.com/SUSE/doc-sle/files/8464811/sec-control-io-cgroups_color_draft_en.pdf)

Werkov:
Thanks. I've pushed few more updates/fixups, I've also added the section intro texts as per added TODOs.

Here are my stylistic/form remarks that I'm seeking some advice on:
- Maybe listing the history of writeback charging support (in 9.5.1.1) is not relevant for the docs and the paragraph could just list the filesystems with cgroup writeback support and link to Mathias's flowchart (or similar).
- 9.5.1.6 it looks a bit text-wall-y, would it make sense to turn each paragraph into a bullet?
- the `echo` command in 9.5.1.3 should run as root -- I've used the `sudo` prefix as with other examples, however, that doesn't work in this case, I'd need something like `echo bfq | sudo tee /sys/class/block/$DEV/queue/scheduler` or mark the whole code block as run as root,
- the "picture" in 9.5.1.4 would (IMO) look better centered, should I modify the ASCII art?

cjschroder:
> Thanks. I've pushed few more updates/fixups, I've also added the section intro texts as per added TODOs.
> 
> Here are my stylistic/form remarks that I'm seeking some advice on:
> 
>     * Maybe listing the history of writeback charging support (in 9.5.1.1) is not relevant for the docs and the paragraph could just list the filesystems with cgroup writeback support and link to Mathias's flowchart (or similar).

I agree. 

> 
>     * 9.5.1.6 it looks a bit text-wall-y, would it make sense to turn each paragraph into a bullet?

Yes, a bulleted list is easier to read.
> 
>     * the `echo` command in 9.5.1.3 should run as root -- I've used the `sudo` prefix as with other examples, however, that doesn't work in this case, I'd need something like `echo bfq | sudo tee /sys/class/block/$DEV/queue/scheduler` or mark the whole code block as run as root,

The correct entity for the root prompt is `&prompt.root;` 
You should also replace $DEV with an example in replaceable tags, like this:
`/sys/class/block/<replaceable>sda1</replaceable>/queue/scheduler`

> 
>     * the "picture" in 9.5.1.4 would (IMO) look better centered, should I modify the ASCII art?

No, it is fine, centering causes problems. What is the command that produces this output, is this something the user can do?

I am happy to make these changes, I can do it tomorrow. Thanks for all of your work!

cjschroder:
Hi @Werkov, I made the discussed changes. I will make one more review, for wording, grammar, and spelling, after you are satisfied with the content and organization. I appreciate all of your work.
[sec-control-io-cgroups_color_draft_en.pdf](https://github.com/SUSE/doc-sle/files/8476770/sec-control-io-cgroups_color_draft_en.pdf)

xml/tuning_cgroups.xml
```diff
@@ -286,246 +286,225 @@ TasksMax=16284
...
-   <title>Using cgroup-v1</title>
-   <para>
-    The following proportional weight policy files can be used to grant a
-    reader process a higher priority for I/O operations than other reader
-    processes accessing the same disk.
-   </para>
+<sect3 xml:id="sec-cgroups-filesystems">
+<title>Filesystem</title>
+<para>
+You must use a cgroup-aware filesystem. The recommended SLES
```
Werkov, line 311: Technically, it's "cgroup-writeback-aware" (sounds terrible).

Here's bacgrkound to what I meant. We can speak of two types of I/O: direct (can be both reader or writer) vs writeback. The I/O control of the former doesn't need special support of the filesystem, however, the latter requires the FS support.

xml/tuning_cgroups.xml
```diff
@@ -286,246 +286,225 @@ TasksMax=16284
...
-...
-&prompt.reader2;echo $$; dd if=file2.txt of=/dev/null bs=4k count=131072
-5251
-131072+0 records in
-131072+0 records out
-536870912 bytes (537 MB, 512 MiB) copied, 1.69026 s, 318 MB/s
-</screen>
-  </sect2>
+<para>
+As of &productname;&nbsp;3, any of the named filesystems can be used.
```
Werkov, line 335: This renders as "SUSE Linux Enterprise Server 3".

xml/tuning_cgroups.xml
```diff
@@ -286,246 +286,225 @@ TasksMax=16284
...
-[...]
+ &prompt.root;<command>echo bfq >/sys/class/block/<replaceable>sda1</replaceable>/queue/scheduler</command>
 </screen>
-   <para>
-    Each process gets half of the throughput for I/O operations. Set up two
-    control groups&mdash;one for each process&mdash;verify that BFQ is the
-    active scheduler, and set a different weight for reader2:
-   </para>
+<para>
+<literal>$DEV</literal> is the disk device, not a partition and the
```
Werkov, line 373: This should be the same replaceable as in the code block above.
But more importantly, it should be only `sda` (without the partition number).

xml/tuning_cgroups.xml
```diff
@@ -286,246 +286,225 @@ TasksMax=16284
...
+The writeback activity depends on the amount of dirty pages, besides the
+global sysctl knobs (<filename>vm.dirty_background_ratio</filename> and
+<filename>vm.dirty_ratio</filename>) memory limits of individual cgroups
+come into play when the dirty limits are distributed among cgroups and
+this in turn may affect I/O intensity of affected cgroups.
+</para>
+</listitem>
+<listitem>
+<para>
+As explained in the introduction, inodes are the unit of writeback charging,
```
Werkov, line 461: The introduction lost the inodes explanation, so this is a broken "reference". I think "As explained in the introduction" can be dropped here too.

Werkov (commented):
Thanks for another round. I've posted my current comments inline where relevant.

> No, it is fine, centering causes problems. What is the command that produces this output, is this something the user can do?

OK. It's not an output of anything, it's just how I drew a tree hierarchy (quick'n'cheap ASCII art).

cjschroder:
Thank you very much for your hard work, @Werkov! It is important to explain these things clearly and concisely, which can be a lot of work. Fresh PDF attached; I made a few minor wording changes as well as the corrections. If it looks all right to you I will merge. You can open a PR or ping me any time for corrections & additional content. 
[sec-control-io-cgroups_color_en.pdf](https://github.com/SUSE/doc-sle/files/8483311/sec-control-io-cgroups_color_en.pdf)

Werkov:
I've pushed some more fixups (described in respective commit messages).

Splendid, I'm satisfied :-) I think you can merge/squash and make it official.
Thanks for your help!

xml/tuning_cgroups.xml
```diff
@@ -308,8 +308,9 @@ during runtime.
 <sect3 xml:id="sec-cgroups-filesystems">
 <title>Filesystem</title>
 <para>
-You must use a cgroup-writeback-aware filesystem. The recommended SLES
-filesystems added support in the following upstream releases:
+You should use a cgroup-writeback-aware filesystem (otherwise writeback charing is not possible).
```
cjschroder, line 311: Is "charing" supposed to be "sharing"?

xml/tuning_cgroups.xml
```diff
@@ -286,246 +286,221 @@ TasksMax=16284
...
-io
-</screen>
-   <para>
-    Now run all the test steps, similarly to the steps for cgroup-v1. Note that
-    some of the directories are different. Run a single reader process (in the
-    examples, reading from an SSD) without controlling its I/O, using
-    file2.txt:
-   </para>
+<sect3 xml:id="sec-cgroups-unified-hierarchy">
+<title>Unified cgroup hiearchy</title>
```
cjschroder, line 341: "hierarchy"

cjschroder (changes requested):
This is really good. I found a couple of typos, not sure what to do with the first one.

xml/tuning_cgroups.xml
Werkov: Hm, the default vim spellchecker is confused with XML and pass this through. Thanks for spotting. Pushed one more fixup.

xml/tuning_cgroups.xml
Werkov: "charging" or "accounting" both make sense. (In more detail, the filesystem must support means that it's known which cgroup should be charged for the writeback I/O or which cgroup the writeback I/O should be accounted to.)

Pushed a commit with my original intention.

cjschroder:
Thank you! And now, the fun part: merging. :partying_face:

PR #1160, opened 2022-04-08
add explanation about hibernation handling in autoyast bootloader

jreidinger:
### PR creator: Description

Adds explanation how hibernation is handled as part of kernel parameters.
Related pull request with explanation: https://github.com/yast/yast-bootloader/pull/666/

Please note that for SP3 and SP4 parts will work only if installer self update is enabled as it is not in default insts-sys.
And the part that won't work is detection if device will exist. So invalid device is just applied without self-update.

* https://bugzilla.suse.com/show_bug.cgi?id=1187690
* https://bugzilla.suse.com/show_bug.cgi?id=1197192

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP1
  - [ ] SLE 15 GA
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3

### DIFF
--- a/xml/ay_bootloader.xml
+++ b/xml/ay_bootloader.xml
@@ -110,10 +110,10 @@
       key with the device path. The recommended way to get stable results is
       configuring your own partitioning, and having a swap device with a
       label:
+    </para>
       <screen>
         &lt;append&gt;quiet resume=/dev/disk/by-label/my_swap&lt;/append&gt;
       </screen>
-     </para>
     <para>
       If you do not use <literal>resume</literal> or
       <literal>noresume</literal>, or if <literal>resume</literal> specifies
--- a/xml/ay_bootloader.xml
+++ b/xml/ay_bootloader.xml
@@ -93,6 +93,35 @@
    <command>yast2-bootloader</command>. Unless you have some special
    requirements, do not specify the boot loader configuration in the XML file.
   </para>
+  <tip>
+    <title>Hibernation</title>
+    <para>
+     If there is a need for specific hibernation settings then
+     <literal>resume</literal> or <literal>noresume</literal> in
+     <literal>append</literal> configuration can be used. It has specific
+     handling.
+    </para>
+    <para>
+      To disable hibernation regardless what installation proposes then specify
+      <literal>noresume</literal> as kernel parameter in
+      <literal>append</literal> section.
+    </para>
+    <para>
+      To specify hibernation device use <literal>resume</literal> key with
+      value which is device path. Recommended way to get stable result is using
+      own partitioning and having swap device with label. Example with labeled
+      swap:
+      <screen>
+        &lt;append&gt;quiet resume=/dev/disk/by-label/my_swap&lt;/append&gt;
+      </screen>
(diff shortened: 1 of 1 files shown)


jreidinger:
It is first time I use that tip element, so feel free to correct me if I use it wrongly and of course feel free to rephrase it as needed. For more context see mentioned pull request.

cjschroder:
Thank you @jreidinger! I made a few changes:
1. <screen/> tags cannot be inside <para/> tags
2. Some wording changes. Please review to make sure I did not introduce any errors.

When you are OK with my changes, I will merge & backport.

xml/ay_bootloader.xml
```diff
@@ -96,30 +96,29 @@
...
+     </para>
     <para>
-      If <literal>resume</literal> neither <literal>noresume</literal> is used
-      or if resume contains device that won't exist on installed system then
-      hibernation is proposed by installer, either enabled or disabled and also
-      hibernation device will be proposed.
+      If you do not use <literal>resume</literal> or
+      <literal>noresume</literal>, or if <literal>resume</literal> specifies
+      a device that will not exist on the installed system, then
+      hibernation will be proposed by installer, either enabled or disabled, and the installer will also propose a hibernation device.
```
jreidinger, line 121: when I read it I probable use not so clear wording here. As with disabled hibernation it can mean that installer propose noresume, but in fact in such case installer does not add any resume or noresume. Not sure how to write it to make it obvious.

jreidinger:
@cjschroder thanks for changes. I just add one small note which I am not sure how to express. Installer proposal basically adds resume=something_valid or nothing depending on several conditions.

xml/ay_bootloader.xml
cjschroder: Thanks, this needs to be clear. Is this correct:

"If you do not use resume or
      noresume, or if resume specifies
      a device that will not exist on the installed system, then
      noresume will be proposed by installer. The installer will not make any changes unless you approve them."

jreidinger: I think it is better. Just last sentence is not correct. It will do change and confirm depends on confirm mode. So to keep it simple, I would remove the last sentence as it is complicated enough :)

jreidinger: maybe I would slightly change wording to:

"If you do not use resume or
noresume, or if resume specifies
a device that will not exist on the installed system, then
the installer may propose a correct value for resume=, or it may remove hibernation parameter completely, depending on installer logic."

What do you think?

cjschroder: I think I am happy that the AY team maintains the manual 😁
That sounds good, short and to the point.

cjschroder:
@jreidinger Thank you for the extra effort to get the wording correct. Explaining things clearly and concisely is difficult, and necessary. Are you satisfied this is ready to merge?

jreidinger:
@cjschroder yes, now it looks good to me. Thanks for your effort

PR #1210, opened 2022-05-10
Replace modules & extensions table

cwickert:
### PR creator: Description

The giant modules and extensions table does not work on small screens, especially not with modules like the NVIDIA Cuda module since it has `<important>`s .

In order to get rid of the table, I first tried a `<variablelist>` but then settled for `<sect2>`s and an `<itemizedlist>`. Looks great and works on small screens.

Additionally, I added a 'Support' section that contains the content of the former footnotes (since we cannot have a `<para>` after a `<sect2>`) and some additional information.

Everything is validated, spell- and linkchecked, so I see no reasons to not merge it, even if it is late in the cycle.

- SLE 15/openSUSE Leap 15.x
  - [x ] SLE 15 SP4/openSUSE Leap 15.4 *(current `main`, no backport necessary)*

Backports are optional.

### DIFF
--- a/xml/art_modules.xml
+++ b/xml/art_modules.xml
@@ -117,873 +117,582 @@
    https://bugzilla.suse.com/1189275#c9 and
    https://jira.suse.com/browse/SLE-17270
   -->
-   <informaltable>
-    <tgroup cols="4">
-     <colspec colnum="1" colname="name" colwidth="30*"/>
-     <colspec colnum="2" colname="dependency" colwidth="25*"/>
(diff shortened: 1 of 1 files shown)


taroth21 (changes requested):
Apart from the screenshots still showing SP3 this looks really good to me - many thanks, @cwickert!

taroth21 (approved):
With the screenshots updates being integrated in main, I'm fine with merging this.  

 > Screenshots are updates in main already with [649b817](https://github.com/SUSE/doc-sle/commit/649b81750a10eaf5409512d369d82b8855bb8a9e)

PR #1213, opened 2022-05-11
Fix links  (single-html -> html)

taroth21:
### PR creator: Description

With single-HTML being deprecated as part of the new HTML layout, we need to fix the links.

* jsc#DOCTEAM-620

No backports as such, will check each maintenance branch individually.


taroth21:
As we don't touch single-html for SBPs/TRDs/Smart Docs (yet), the other two single-html links I found via `grep` can stay and we can merge this PR.

PR #1223, opened 2022-05-12
Improve sev doc

jfehlig:
### PR creator: Description

Update and improve the AMD SEV doc for SLE15 SP4.

* jsc#SLE-18834

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP1
  - [ ] SLE 15 GA
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3

### DIFF
--- a/xml/art_amd-sev.xml
+++ b/xml/art_amd-sev.xml
@@ -101,7 +101,12 @@
   <title>VM requirements</title>
   <para>
    The VM must be the modern <literal>Q35</literal> machine type and must use
-   UEFI firmware.
+   UEFI firmware. &libvirt; can automatically select an appropriate SEV or
+   SEV-ES enabled UEFI firmware, or one can be specified manually. Currently,
+   the only firmwares supported are <filename>/usr/share/qemu/ovmf-x86_64-code.bin</filename>
+   and <filename>/usr/share/qemu/ovmf-x86_64-4m-code.bin</filename>. See
+   <xref linkend="sec-vt-installation-ovmf"/> for more details on using UEFI
+   firmware and the auto-selection feature.
   </para>
   <note>
    <title>No IDE support in Q35</title>
--- a/xml/art_amd-sev.xml
+++ b/xml/art_amd-sev.xml
@@ -118,7 +118,9 @@
    All <literal>virtio-net</literal> devices need to be configured with the iPXE
    option ROM disabled. iPXE is currently not compatible with SEV and SEV-ES.
    In addition, all memory regions used by the VM must be locked for Direct Memory
-   Access (DMA) and to prevent swapping.
+   Access (DMA) and to prevent swapping. This includes memory for the VM and any
+   memory regions allocated by &qemu; to support running the VM, such as UEFI
+   pflash for firmware and variable store, video RAM, etc.
   </para>
  </sect1>
  <sect1 xml:id="sec-amd-sev-config">
@@ -133,7 +135,7 @@
     &lt;memory unit='KiB'&gt;4194304&lt;/memory&gt;
     &lt;currentMemory unit='KiB'&gt;4194304&lt;/currentMemory&gt;
     &lt;memoryBacking&gt;
-    &lt;locked/&gt;
+    &lt;locked/&gt; <co xml:id="sec-amd-sev-ex-memlock"/>
     &lt;/memoryBacking&gt;
     &lt;os&gt;
     &lt;type arch='x86_64' machine='pc-q35-2.11'&gt;hvm&lt;/type&gt;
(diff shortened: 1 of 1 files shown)


jfehlig:
FYI @vasiliy-ul @hw-claudio @joergroedel @darrenrdavis

darrenrdavis:
Will review.

vasiliy-ul:
Looks good from my perspective :+1:

xml/art_amd-sev.xml
```diff
@@ -118,7 +118,9 @@
    All <literal>virtio-net</literal> devices need to be configured with the iPXE
```
hw-claudio, line 118: not sure if mentioned elsewhere, but should we mention that all devices (not only virtio-net) should have option roms disabled?

hw-claudio (commented):
Did we add somewhere the limitation about the Naples generation?

If not, I'd add something like:

"Supported Hardware" - SUSE does not recommend using the SEV and SEV-ES features with 
AMD EPYC™ first generation (codename Naples) 7000 series of processors with SUSE Linux products. It is recommended to use at least second generation (codename Rome) 7002 series processors.
SUSE Linux products on "Naples" generation AMD EPYC systems.

xml/art_amd-sev.xml
```diff
@@ -440,6 +440,14 @@ spec:
...
    </listitem>
+   <listitem>
+    <para>
+     SEV-ES VMs cannot be rebooted from within using <command>reboot</command>,
+     <command>shutdown -r now</command>, etc. A reboot must be done by shutting
+     down the VM and starting it again. This limitation does not apply to SEV
+     VMs, only SEV-ES.
+    </para>
+   </listitem>
   </itemizedlist>
```
hw-claudio, line 460: Do we also have to mention something about Secure Boot for guests? If I understand correctly at least current generation SEV/SEV-ES have some limitation when trying to use them in combination with OVMF firmwares that provide SMM mode support, so are SEV/SEV-ES fundamentally incompatible with Secure Boot? Or what is the right relationship? @joergroedel

xml/art_amd-sev.xml
```diff
@@ -487,6 +491,12 @@ spec:
      &mdash; libvirt SEV configuration settings
     </para>
    </listitem>
+   <listitem>
+    <para>
+     <link xlink:href="https://libvirt.org/kbase/launch_security_sev.html"/>
+     &mdash; libvirt knowledge base article on AMD SEV
+    </para>
+   </listitem>
   </itemizedlist>
```
hw-claudio, line 500: should we mention also virt-manager configuration specifically for SUSE Linux 15SP4 as a TP Feature?
Explaining where to find the checkbox on launch for virt-manager (advanced dialog, memory section as "Launch Security"), and also mentioning the use of virt-install to install the guests, clarifying the additional launch security values/options? @jfehlig (and at charles, but could not find the user)

xml/art_amd-sev.xml
vasiliy-ul: Interesting. If that is the case then probably this will need to be adjusted in KubeVirt as well. So is it confirmed?

xml/art_amd-sev.xml
jfehlig: Secure boot is not enabled by default and it is not a commonly used feature like reboot, save/restore, and migration. I don't see SEV + secure boot not working as a limitation. Also, one of the commits in this PR explicitly lists the two supported firmwares.

hw-claudio: well with the layman in mind, how is one supposed to know that SEV and secure boot cannot be combined if we don't document it somewhere..? There is explicit list of firmwares, but that assumes that a user makes the two connections on his own (ie which firmwares are required for secure boot, which ones for SEV, and put two and two together..) would it hurt to be explicit?

xml/art_amd-sev.xml
jfehlig: I just tried e1000 and it was NOT necessary to disable any option ROM.

xml/art_amd-sev.xml
hw-claudio: well at least we should doc it in 15SP4 Release notes or doc, or even if we decide to mark it as TP, nobody will have a chance to even try the feature...

xml/art_amd-sev.xml
jfehlig: Nod. I'll add another list item stating the two are currently incompatible.

xml/art_amd-sev.xml
jfehlig: IDK. Maybe @joergroedel or someone familiar with virtio devices.

hw-claudio:
@jfehlig great work thanks!!

xml/art_amd-sev.xml
jfehlig: Added it in V2.

xml/art_amd-sev.xml
hw-claudio: @aginies this is possible something for at least 15SP4 Release notes, to sync with Charles (on 15SP4 we have some downstream ease-of-life option in virt-manager to enable launch security backed by AMD SEV. Probably we can mention this virt-manager enablement as TP or whatever, may be replaced by a slightly different GUI solution in the next releases)

aginies: It could be added in the TP section of the RN.

jfehlig: We can add it to the SEV doc too, but can be a separate PR. No need for it to block this PR.

tbazant:
Is this one fully approved and ready for merging?

jfehlig:
> @jfehlig Is this one fully approved and ready for merging?

Yes, I think so. Fine-tuning can be done via followup patches.

PR #1227, opened 2022-05-13
SLES Gnome User Guide proofing integration

dariavladykina:
Integrated proofing corrections for SLES Gnome User Guide from VT.

### DIFF
--- a/xml/gnome_start.xml
+++ b/xml/gnome_start.xml
@@ -239,7 +239,7 @@
    <para>
     The top bar provides access to your windows and applications. Here you can
     set the volume or screen brightness, check the battery status, log out or
-    switch users, and many more.
+    switch users, and much more.
    </para>
    <para>
     The <guimenu>Activities</guimenu> Overview allows you to access your
@@ -349,7 +349,7 @@
     </itemizedlist>
    </sect3>
    <sect3 xml:id="sec-gnome-start-activities-use">
-    <title>Using the <guimenu>Activities</guimenu> overview</title>
+    <title>Using the <guimenu>Activities</guimenu> Overview</title>
     <para>
      In the following, the most important parts of the
      <guimenu>Activities</guimenu> Overview are explained.
@@ -564,7 +564,7 @@
     </step>
     <step>
      <para>
-      Expand the <guimenu>Power Off/Log out</guimenu> menu.
+      Expand the <guimenu>Power Off/Log Out</guimenu> menu.
      </para>
     </step>
     <step>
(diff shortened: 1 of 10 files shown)


xml/gnome_start.xml
```diff
@@ -349,7 +349,7 @@
     </itemizedlist>
    </sect3>
    <sect3 xml:id="sec-gnome-start-activities-use">
-    <title>Using the <guimenu>Activities</guimenu> overview</title>
+    <title>Using the <guimenu>Activities</guimenu> Overview</title>
```
taroth21, line 352: If the name of the UI element was _Activities Overview_, I'd agree, but here (and for the following occurrences) of _Activities Overview_ I'm wondering if this change is correct. 

Why should we spell 'Overview' with a capital 'O' when 

- we use sentence style for headings? 
- the name of this UI element seems to be _Activities_ ? (which is also correctly tagged with `<guimenu/>` in the XML file) - see also the screenshot on page 23 in the diff-PDF with the corrections (Figure 1.4). I rather think it should be `<guimenu>Activities</guimenu> overview` everywhere.

taroth21 (changes requested):
I think one term needs to be changed consistently (in cause of doubt, check with @jfaltenbacher as I only can rely on the screenshot, I don't have a VM with a graphical desktop installed at the moment). Thanks!

xml/gnome_start.xml
jfaltenbacher: Hi all,
here is a sceenshot from the GNOME user help. They write overview with lower case:
grafik
I can also do the corrections, if by tomorrow is sufficient

dariavladykina: Hi Tanja and Julia, 
I agree that we should correct 'overview' everywhere. The reason why I kept the capitalized 'Overview' throughout the entire guide was the simple majority of instances where it was historically spelled with the capital 'O'. Also, lots of external resources have it capitalized for some reason :-)

xml/gnome_use.xml
```diff
@@ -29,7 +29,7 @@
   <title>Managing files and directories</title>
 
   <para>
-   To start &nautilus;, open the <guimenu>Activities</guimenu> Overview  by
+   To start &nautilus;, open the <guimenu>Activities</guimenu> overview  by
```
dariavladykina, line 32:
```suggestion
   To start &nautilus;, open the <guimenu>Activities</guimenu> overview by
```

dariavladykina (commented):
Extra space removed

jfaltenbacher (approved):
many thanks

PR #1259, opened 2022-06-10
Container Guide proofing integration

dariavladykina:
Integrated proofing corrections for the Container Guide from VT.

### DIFF
--- a/xml/containers-support.xml
+++ b/xml/containers-support.xml
@@ -349,7 +349,7 @@
   </itemizedlist>
 
   <para>
-   Container images are labeled as Tech Preview are marked as such at
+   Container images are labeled as Tech Preview and are marked as such at
    <link
 xlink:href="https://registry.suse.com">registry.suse.com</link>.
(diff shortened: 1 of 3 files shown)


xml/containers-support.xml
```diff
@@ -349,7 +349,7 @@
   </itemizedlist>
 
   <para>
-   Container images are labeled as Tech Preview are marked as such at
+   Container images are labeled as Tech Preview and are marked as such at
```
taroth21, line 352:
```suggestion
   Container images are labeled as <literal>Tech Preview</literal> and are marked as such at
```

taroth21: OK, I noticed that the rest of the para uses the correct terminology ('technology preview'). I assume 'Tech Preview' is the label being used in registry.suse.com, therefore I would put a ```<literal/>``` element around it.

taroth21 (commented):
@dariavladykina : Thanks, one small comment from my side only.

PR #1299, opened 2022-07-06
Fix validation errors for SLED/SLES in zh_cn

tomschr:
### PR creator: Description

* Add missing DOCTYPE
* Disabled xi:include to missing file `security_ldap_kerberos_ad_yast.xml`

For `maintenance/SLE15SP2`.


cjschroder:
Thank you! It validates! So amaze :)

PR #1316, opened 2022-07-25
overhauled graphics for 2022

janajaeger:
Overhauled graphics for the Container Guide. 

- container-* seems to be redundant nowadays?


janajaeger:
as there's the migration to asciidoc under way, it makes no sense to merge here ;) Dima has already taken what he needs ;)

PR #1321, opened 2022-07-27
Jhalackova multipath proofread

lvicoun:
### PR creator: Description

Storage multipath proofread.

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 next/openSUSE Leap next *(current `main`, no backport necessary)*
  - [X] SLE 15 SP4/openSUSE Leap 15.4
  - [ ] SLE 15 SP1
  - [ ] SLE 15 SP0
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4
  - [ ] SLE 12 SP3


xml/storage_multipath.xml
```diff
@@ -1747,658 +1686,639 @@ devices {
...
+   </varlistentry>
+   <varlistentry>
+    <term>fast_io_fail_tmo</term>
+    <listitem>
+     <para>
+      If an error on the SCSI transport layer is detected (for example on a
+      Fibre Channel remote port), the kernel transport layer waits for this
+      amount of time (in seconds) for the transport to recover. After that, the
+      path device fails with "transport offline" state. This is very useful for
+      multipath, because it allows quick path failover for a frequently
```
dariavladykina, line 1788:
```suggestion
      multipath, because it allows a quick path failover for a frequently
```

xml/storage_multipath.xml
```diff
@@ -1747,658 +1686,639 @@ devices {
...
+       </listitem>
+      </varlistentry>
+      <varlistentry>
+       <term>queue</term>
+       <listitem>
+        <para>
+         I/O on the multipath map is queued in the device-mapper layer and sent
+         to the device when path devices become available again. This is the
+         safest option to avoid losing data, but it can have negative effects
+         if the path devices don't get reinstated for a long time. Processes
```
dariavladykina, line 1838:
```suggestion
         if the path devices do not get reinstated for a long time. Processes
```

xml/storage_multipath.xml
```diff
@@ -1747,658 +1686,639 @@ devices {
...
+     </para>
+    </listitem>
+   </varlistentry>
+   <varlistentry>
+    <term>deferred_remove</term>
+    <listitem>
+     <para>
+      If set to <literal>yes</literal> and all path devices of a map are
+      deleted, wait for holders to close the file descriptors for the map
+      device before flushing and removing it. If paths reappear before the the
```
dariavladykina, line 1879:
```suggestion
      device before flushing and removing it. If paths reappear before the
```

xml/storage_multipath.xml
```diff
@@ -1747,658 +1686,639 @@ devices {
...
+      <xref linkend="sec-multipath-grouping"/>).
+      After the reevaluation, the highest-priority path group may be one of the
+      currently inactive path groups. This parameter determines what happens in
+      this situation.
+     </para>
+     <important>
+      <title>Observe vendor recommendations</title>
+      <para>
+       The optimal failback policy depends on the property of the storage
+       device. It's therefore strongly encouraged to verify
```
dariavladykina, line 1901:
```suggestion
       device. It is therefore strongly encouraged to verify
```

dariavladykina (commented):
Hi Jana, please see some minor suggestions.

xml/storage_multipath.xml
```diff
@@ -3711,181 +3605,178 @@ after mounting them and attach it to a bug report.
...
+      udev properties, in particular <literal>SYSTEMD_READY</literal> (see
+      above)
+     </para>
+<screen>&prompt.root;udevadm info -e</screen>
+    </step>
+    <step>
+     <para>
+      If the previous step showed unexpected udev properties, something may
+      have gone wrong during udev rule processing. Check other properties, in
+      particular those used for device identification (see
```
dariavladykina, line 3678:
```suggestion
      particular, those used for device identification (see
```

xml/storage_multipath.xml
```diff
@@ -3711,181 +3605,178 @@ after mounting them and attach it to a bug report.
...
+       </callout>
+       <callout arearefs="mp-co-kparm-rd">
+        <para>
+         Trace the scripts in the initramfs.
+        </para>
+       </callout>
+      </calloutlist>
+     </informalexample>
+     <para>
+      In addition, it may make sense to enable logging for certain drivers, and
```
dariavladykina, line 3767:
```suggestion
      In addition, it may make sense to enable logging for certain drivers and
```

xml/storage_multipath.xml
```diff
@@ -3711,181 +3605,178 @@ after mounting them and attach it to a bug report.
...
+     <para>
+      You should be able to correct the situation with a few shell commands,
+      exit the emergency shell, and boot successfully. You will still need to
+      adjust your configuration to make sure the same problem will not occur
+      again in the future.
+     </para>
+     <para>
+      Otherwise, you will need to boot the rescue system, set up the devices
+      manually in order to <command>chroot</command> into the real root file
+      system, and attempt to fix the problem based on the insight you got in
```
dariavladykina, line 3720:
```suggestion
      manually to <command>chroot</command> into the real root file
      system, and attempt to fix the problem based on the insight you got in
```

dariavladykina (commented):
Hi Jana, please see some more minor suggestions.

xml/storage_multipath.xml
```diff
@@ -2489,723 +2406,682 @@ blacklist_exceptions {
...
+       This option controls the behavior of <command>multipath</command> and
+       <command>multipathd</command> when a device that is not excluded is
+       first encountered. The possible values are:
+      </para>
+      <variablelist>
+       <varlistentry>
+        <term>greedy</term>
+        <listitem>
+         <para>
+          Every non-excluded device is included. This is the default on &sle;.
```
mwilck, line 2449: So you removed `&opensuse;`? What does this mean for the openSUSE docs? Will this chapter be updated there, eventually, too?

xml/storage_multipath.xml
```diff
@@ -2489,723 +2406,682 @@ blacklist_exceptions {
...
+          If this setting is active, the only way to prevent devices from being
+          added to multipath maps is setting them as excluded.
```
mwilck, line 2451: We should keep the term "blacklisting" here. I know it's considered politically incorrect these days, but it's the technical term used in the the config file, and we should be precise. "excluded" isn't 100% unambiguous. If you must keep "excluded", please add an xref to make sure readers understand what we mean.

xml/storage_multipath.xml
```diff
@@ -2489,723 +2406,682 @@ blacklist_exceptions {
...
-	</listitem>
-      </varlistentry>
-    </variablelist>
-  <warning>
+   <title>Using autogenerated user-friendly names</title>
+   <para>
+    <systemitem>multipath-tools</systemitem> also supports autogenerated
+    aliases, so-called &ldquo;user-friendly names&rdquo;. The naming scheme of
+    the aliases follows the pattern:
+    <replaceable>A_NAME_PREFIX_LOWER_CASE_LETTERS</replaceable>, for example,
```
mwilck, line 2651: I don't like this much. Many readers will interpret this as a literal, and not realize that it's actually comprised of two parts (prefix and "number"). Is this a convention followed elsewhere in the SLE docs?

mwilck: The examples below may clarify the meaning,  but I still find `<replaceable>A_NAME_PREFIX_LOWER_CASE_LETTERS</replaceable>` confusing.

xml/storage_multipath.xml
```diff
@@ -3225,233 +3101,233 @@ devices {
...
     LVM2 has built-in support for detecting multipath devices. It is activated
     by default in <filename>/etc/lvm/lvm.conf</filename>:
-  </para>
-  <screen>    multipath_component_detection=1</screen>
-  <para>
-    In order to make this work reliably, LVM2 must be configured to obtain
+   </para>
+<screen>    multipath_component_detection=1</screen>
+   <para>
+    To perform this work reliably, LVM2 must be configured to obtain
```
mwilck, line 3314:
```suggestion
    This only works reliably if LVM2 is also configured to obtain
```

xml/storage_multipath.xml
```diff
@@ -3528,180 +3431,172 @@ size=64G features='3 queue_if_no_path pg_init_retries 50'<co xml:id="mp-co-top-f
...
+     Consider a system with multipathing where the root device is not
+     multipathed, and no devices are excluded from multipath (see
+     <xref linkend="vl-multipath-planning-type-noroot-noinitrd"/>
+     in
+     <xref linkend="sec-multipath-planning-type-noroot"/>).
+     The root file system is mounted in the initramfs.
+     <command>systemd</command> switches to the root file system and
+     <command>multipathd</command> starts up. Because the device is already
+     mounted, <command>multipathd</command> fails to set up the multipath map
+     for it. Because the root device is not in the blocklist, it is considered a
```
mwilck, line 3503:
```suggestion
     for it. Because the root device is not in the blacklist, it is considered a
```

xml/storage_multipath.xml
```diff
@@ -3528,180 +3431,172 @@ size=64G features='3 queue_if_no_path pg_init_retries 50'<co xml:id="mp-co-top-f
...
+    </listitem>
+    <listitem>
+     <para>
+      Is the multipath dracut module included in the initramfs?
+     </para>
+    </listitem>
+    <listitem>
+     <para>
+      Is my root device configured as a multipath device? If not, is the root
+      device properly blocklisted, or are you relying on the absence of the
```
mwilck, line 3583:
```suggestion
      device properly blacklisted, or are you relying on the absence of the
```

mwilck (changes requested):
Apart from some very minor corrections, I am fine with this. Just note that it's "blacklist", not "blocklist", even though this isn't politically correct. "blocklist" has an entirely different meaning in the context of Linux block devices.

xml/storage_multipath.xml
lvicoun: Yes, this is stated even in the Preface of our docs that it is a placeholder and you need to replace it with our own value.

xml/storage_multipath.xml
```diff
@@ -444,165 +410,165 @@
...
+   <para>
+    If you select &ldquo;No&rdquo; at this prompt (not recommended), the
+    installation will proceed as in
+    <xref linkend="sec-multipath-installing-nomp"/>. In the partitioning stage,
+    do not use/edit devices that will later be part of a multipath map.
+   </para>
+   <para>
+    If you select &ldquo;Yes&rdquo; at the multipath prompt,
+    <command>multipathd</command> will run during the installation. No
+    blocklist will be configured, thus all SCSI and DASD devices, including
```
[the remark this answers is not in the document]
lvicoun: I'm not keen on using blocklists here, but we decided to use inclusive language and replace the blacklist with something else, even though in the product it is blacklist. I'm fine with using another word than blocklist if blocklist has a different meaning. And I'm all ears for your suggesstions.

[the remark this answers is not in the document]
mwilck: Hm, we've been using "blacklist" all over the place in this document (I count 39 occurences in my version). It's even in section headers. I don't see the point in avoiding it here. IMO clarity and non-ambiguity is more important than being politically correct. (I won't hide that I don't understand why anyone would seriously relate this term to skin tone, and yes I've read a couple of discussions on the matter and I'm still not convinced).

If we are serious about the inclusive language, we should replace "blacklist" everywhere, using a consistent term that doesn't look like a typo (e.g. "exclude list"), and prominently explain (in the nomenclature section?) that we use this term instead of the technical one (which isn't going to change over the lifetime of SLE, unless I'm very mistaken). We should also add xrefs everywhere to avoid misunderstanding.

xml/storage_multipath.xml
mwilck: My main point is that this obfuscates the fact that the placeholder consists of two independent parts. Actually, the underscore suggests (to me at least) that it's a single "variable". What about something like `<replaceable>PREFIX</replaceable>&nbsp;<replaceable>INDEX</replaceable>`, and explaining the two parts below?

Anyway, you're the boss, so if you want it this way, fair enough.

xml/storage_multipath.xml
lvicoun: Opensuse does not have the document.

xml/storage_multipath.xml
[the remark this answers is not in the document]
lvicoun: Yep, the other blacklist replacement is under construction right now. I agree with you, I'd keep blacklist :) though the decision has been made for me. Well, this particular term is not about the skin tone so much, but about giving colors a meaning or a mood, which can be in different cultures interpreted in a different way. 
I'll go through the chapter (and the doc later) to clarify this change as much as possible.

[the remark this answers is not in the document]
mwilck: I would be grateful if we could separate this part of the modifications from my work. It's clearly independent.

xml/storage_multipath.xml
```diff
@@ -1329,249 +1276,247 @@
...
     be created automatically. If this does not happen,
     <xref linkend="sec-multipath-trouble-steps" xrefstyle="SecTitleOnPage"/>
-    lists some shell commands commands that
-    can be used to examine the situation.
-    When the LUNs are not seen by the HBA driver, check the
-    zoning setup in the SAN. In particular, check whether LUN masking is active
-    and whether the LUNs are correctly assigned to the server.
+    lists some shell commands that can be used to examine the
+    situation. When the LUNs are not seen by the HBA driver, check the zoning
+    setup in the SAN. In particular, check whether LUN masking is active and
```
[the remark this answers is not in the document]
lvicoun: Agreed.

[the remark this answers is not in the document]
mwilck: Is this term mentioned anywhere else in the SLE docs, or in the storage admin guide specifically?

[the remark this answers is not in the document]
lvicoun: No, only here.

PR #1350, opened 2022-08-30
NVIDIA CUDA toolbox download change (bsc#1202659)

tbazant:

* bsc#1202659

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP1
  - [ ] SLE 15 SP0
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4

### DIFF
--- a/xml/art-nvidia-vgpu.xml
+++ b/xml/art-nvidia-vgpu.xml
@@ -987,7 +987,13 @@ EndSection
    <step>
     <para>
      Download and install the CUDA toolkit. You can find it at
-     <link xlink:href="https://developer.nvidia.com/cuda-downloads?target_os=Linux&amp;target_arch=x86_64&amp;Distribution=SLES&amp;target_version=15"/>.
+     <link
+     xlink:href="https://developer.nvidia.com/cuda-downloads?target_os=Linux&amp;target_arch=x86_64&amp;Distribution=SLES&amp;target_version=15&amp;target_type=runfile_local"/>.
+    </para>
+   </step>
+   <step>
+    <para>
+     Download CUDA samples from <link xlink:href="https://github.com/nvidia/cuda-samples"/>.
     </para>
    </step>
    <step>
@@ -995,7 +1001,7 @@ EndSection
      Run CUDA sampling example:
     </para>
 <screen>
-&prompt.user;cd ~/NVIDIA_CUDA-12.0_Samples/0_Simple/clock
+&prompt.user;cd <replaceable>YOUR_GIT_CLONE_LOCATION</replaceable>/cuda-samples/Samples/0_Introduction/clock
 &prompt.user; make
 /usr/local/cuda/bin/nvcc -ccbin g++ -I../../common/inc  -m64    --threads 0 -gencode arch=compute_35,code=sm_35 -gencode arch=compute_37,code=sm_37 -gencode
 [...]


xml/art-nvidia-vgpu.xml
```diff
@@ -987,15 +987,21 @@ EndSection
    <step>
     <para>
      Download and install the CUDA toolkit. You can find it at
-     <link xlink:href="https://developer.nvidia.com/cuda-downloads?target_os=Linux&amp;target_arch=x86_64&amp;Distribution=SLES&amp;target_version=15"/>.
+     <link
+     xlink:href="https://developer.nvidia.com/cuda-downloads?target_os=Linux&amp;target_arch=x86_64&amp;Distribution=SLES&amp;target_version=15&amp;target_type=runfile_local"/>.
```
cwickert, line 991: IHMO we should recommend the RPM package, not the installer script. The package can be checksummed and it's RPM signature can be verified.

```suggestion
     xlink:href="https://developer.nvidia.com/cuda-downloads?target_os=Linux&amp;target_arch=x86_64&amp;Distribution=SLES&amp;target_version=15&amp;target_type=rpm_local"/>.
```

cwickert (commented):
Is there any reason to not use SUSEconnect?
`SUSEConnect -p sle-module-NVIDIA-compute/15/x86_64`

If you stick to these instructions, better use the RPM instead of the installer script.

xml/art-nvidia-vgpu.xml
Julie-CAO: I stated the reason why a runfile rather than a rpm should be used in the bug https://bugzilla.suse.com/show_bug.cgi?id=1202659. 

- The first reason is that the vGPU grid driver is installed in runfile which is mentioned in this guide, and NVIDIA sugguests to use the runfile to install cuda as well. 

- The second reason is that I did install cuda with a rpm, but the version of gpu driver in cuda 11.7 mismatch with the one in vGPU grid driver. As a consequence, the gpu driver was uninstalled by cuda automatically. Of course neither the CUDA or the vGPU did not work finally. I struggled with several versions and failed to find the completely same ones at last, so I'm afraid that I'll be not the only person who will face the issue.

I know little about the checksum or suseconnect thing with CUDA toolkit in SUSE. The PM or any management level can decide which way to use based on the situations.

Julie-CAO:
These changes fully conform to the request I submitted in the bug. Thanks for the effort.

dariavladykina (approved):
Not much for me to comment on here :-)

xml/art-nvidia-vgpu.xml
Julie-CAO: Hi @cwickert, No matter whether rpm or runfile is used here, no error should be reported when accessing sle-module-NVIDIA-compute repo. I think this problem needs to be solved. 

I am a QA engineer to test the NVIDIA vGPU feature based on the guide, I just managed to install CUDA toolkit and run basic sample test to verify whether the vGPU functions well. I know little about sle-module-NVIDIA-compute repo. Maybe opening a bug is a good way to solve it? Or do you need any help from my PM who has connection with NVIDIA?

cwickert: Yes, if we recommend the SUSEConnect method, we need to get this bug fixed first. Seems the problem is already known as https://bugzilla.suse.com/show_bug.cgi?id=1174657

Background: The key used to sign the repodata is not yet known to rpm, thus zypper does not trust the repomd.xml file.

Solution: 
1. Open the repo https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/ with your browser. You will find two public keys, [7fa2af80.pub](https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/7fa2af80.pub) and [D42D0685.pub](https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/D42D0685.pub). Download both.
2. To find the correct one, open the repodata folder at https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/repodata/ and download the [repomd.xml.key](https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/repodata/repomd.xml.key) file.
3. Investigate the file with
```
sles15sp3:~ # gpg ~/Downloads/repomd.xml.key 
gpg: WARNING: no command supplied.  Trying to guess what you mean ...
pub   rsa4096 2022-04-14 [SC]
      610C7B14E068A878070DA4E99CD0A493D42D0685
uid           cudatools <cudatools@nvidia.com>
```
4. Do the same for the two public key files. You'll see that 7fa2af80.pub is an old key and D42D0685.pub is the correct one, it's date matches.
5. Import the public key:
```
sles15sp3:~ # rpm --import https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/D42D0685.pub
```
6. Now enable the module:
```
sles15sp3:~ # SUSEConnect -p sle-module-NVIDIA-compute/15/x86_64
Registering system to SUSE Customer Center

Updating system details on https://scc.suse.com ...

Activating sle-module-NVIDIA-compute 15 x86_64 ...
-> Adding service to system ...
-> Installing release package ...

Successfully registered system
```
Viola!

So IHMO we should add the `rpm --import...` step to the instructions. Even if we do not recommend the rpm or SUSEConnect, it will help since it avoids warnings during even with the runfile.

But the decision which method to recommend is up to a PM I guess.

cwickert: Forget SUSEConnect and manually importing the key, there is a very simple and officially supported ways: Just add the repo with `yast registration`. Unlike SUSEConnect, it can handle 3rd party repository keys.

@tbazant Do you think you can add this to the instructions?

tbazant: PM's decision is to go for what @Julie-CAO  suggested.

PR #1391, opened 2022-10-31
SLE-25081 add gTile

jfaltenbacher:
Description of how to install and use the gTile extension for window management added.

* jsc#SLE-25081

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP1
  - [ ] SLE 15 SP0
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4

### DIFF
--- a/xml/gnome_custom.xml
+++ b/xml/gnome_custom.xml
@@ -645,6 +645,164 @@ commented out the screen. Texts taken from GNOME help.
     <xref linkend="sec-gnome-settings-hardware-screen-single"/>.
    </para>
   </sect2>
+ <sect2>
+  <title>Managing tiling windows on <productname>Wayland</productname></title>
+  <para>
+   If you are running &productname; on <productname>Wayland</productname> and
+   you want to have your windows arranged next to each other, it is recommended
+   to use the <productname>gTile</productname> extension.
+   <productname>gTile</productname> lets you re-arrange windows based on a
+   configurable grid scheme.  For example, you can launch six terminals, and 
+   have your terminals arranged in 3 rows and 2 columns, so that they use the
+   whole desktop area.
+  </para>
+  <para>
+   For more information about <productname>gTile</productname> and to download
+   the extension, refer to
+   <link xlink:href="https://extensions.gnome.org/extension/28/gtile/">.
+   </link>
+  </para>
+  <sect3>
+   <title>Installing <productname>gTile</productname> extension</title>
+   <procedure>
+    <title>Installing for a single user</title>
+    <step>
+     <para>
+     Download the extension from
+     <link xlink:href="https://extensions.gnome.org/extension/28/gtile/"></link>.
+     </para>
+     <remark>jufa 2022-10-31: as I don't see the On/Off button, I haven't
+      included this step here. Check, if it it now the Install button instead</remark>
+     <para>
+      The extension is stored to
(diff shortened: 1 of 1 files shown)


jfaltenbacher:
content review will be done offline. Waiting for feedback

xml/gnome_custom.xml
```diff
@@ -645,6 +645,164 @@ commented out the screen. Texts taken from GNOME help.
...
+  <sect3>
+   <title>Installing <productname>gTile</productname> extension</title>
+   <procedure>
+    <title>Installing for a single user</title>
+    <step>
+     <para>
+     Download the extension from
+     <link xlink:href="https://extensions.gnome.org/extension/28/gtile/"></link>.
+     </para>
+     <remark>jufa 2022-10-31: as I don't see the On/Off button, I haven't
```
federicomenaquintero, line 674: To see the button:

* You need to be running a GNOME desktop.
* You need the `gnome-browser-connector` package so that the browser chrome extension is installed. [Description of gnome-shell integration for browsers](https://wiki.gnome.org/Projects/GnomeShellIntegration); [Installation of the browser extension](https://wiki.gnome.org/Projects/GnomeShellIntegration/Installation)

jfaltenbacher: Hi Federico, thanks for the hints. I wonder if there is a more easier way to install this. The GNOME Shell browser integration Installation Guide states that is more easy to install from Package Manager. Unfortunately, there is none for SLE (however, I found https://build.opensuse.org/package/show/GNOME:Factory/gnome-browser-connector). I just believe, that is is getting to complicated for the users. Keep in mind, that the GNOME user guide is more for the end user which is not necessarily an experienced developer. If there is a more easy way to install gtile, I would be happy to describe it. If not, we may have to think of moving these instructions into the Admin guide. That would fit better to the target audience.

jfaltenbacher: Hi @federicomenaquintero , is there another option or do you have any opinion on this? The PR is currently blocked. Regards Julia

federicomenaquintero: Ugggh, something changed relatively recently:

- SLE15 SP4 uses GNOME 41, with the `chrome-gnome-shell` package.  This is in the default installation of SP4.
- Upstream GNOME has moved to a `gnome-browser-connector` package, [as documented in the wiki](https://wiki.gnome.org/Projects/GnomeShellIntegration/Installation).  This is in the default installation of openSUSE Tumbleweed, for instance.
- Both require going into Firefox or Chrome to install the corresponding browser extension - no problem, it works.
- However, after that, `https://extensions.gnome.org` complains on SLE15 SP4 because the API version that `chrome-gnome-shell` supports is too old (it wants a "v6 API", presumably from the newer `gnome-browser-connector`).

I'll do some investigation to see if that new API can be backported for `chrome-gnome-shell` in SP4.

I don't know if it's possible to install the GTile extension for GNOME 41 if one downloads it by hand; it may (should?) still work.

jfaltenbacher: Thank you for coming back to this. Hope, you can find a feasible solution.

jfaltenbacher: adding @tbazant as he may move the instructions to the Admin Guide

jfaltenbacher:
closed without merging as content has moved to the Admin guide

PR #1402, opened 2022-11-23
Installation: add SLE-RT 15 SP4 to screenshots

taroth21:
As requested by Jeffrey via mail after SLE-RT 15 SP4 has been released recently.


taroth21:
SLED Install Quick: Screenshot_20221123_110408

SLES Install Quick: 
Screenshot_20221123_110620

cwickert:
I'd rather not use gimped versions and thought I had these screens already in my library, but the ones I have don't show SLE-RT. Even when doing a fresh install today SCC does not offer SLE-RT. As long as it's not really available, we should not document it. ("[Do not promise future features. Only document features that exist already or that will be finished before the document is first published.](https://documentation.suse.com/style/current/single-html/docu_styleguide/#sec-content)")

taroth21:
> I'd rather not use gimped versions and thought I had these screens already in my library, but the ones I have don't show SLE-RT. Even when doing a fresh install today SCC does not offer SLE-RT. As long as it's not really available, we should not document it. ("[Do not promise future features. Only document features that exist already or that will be finished before the document is first published.](https://documentation.suse.com/style/current/single-html/docu_styleguide/#sec-content)")

SLE-RT is released with the Quarterly Updates media, and QU1 (incl. SLE-RT) has been released on Nov 10th). To get a screenshot with SLE-RT, you have to download the QU1 media.

cwickert:
Updated the screens from QU1, everything is fine now.

cwickert (approved):

PR #1411, opened 2022-12-12
Remove <indexterm> from localized docs

cwickert:
### PR creator: Description

The tag `<indexterm>` causes problems when used inline, see https://bugzilla.suse.com/show_bug.cgi?id=1205252

This change removes all `<indexterm>`s from the localized SLE 12 SP5 docs. The English have already been dealt with in cdd322699.

* https://bugzilla.suse.com/show_bug.cgi?id=1205252

The change was generated with the following script:

```
#! /bin/bash
FILES=$(grep -r indexterm l10n/ | awk -F : '{print $1}' | sort -u)
for file in $FILES
    do 
    xmlstarlet ed --inplace -P -N d=http://docbook.org/ns/docbook -d //d:indexterm "$file"
    sed -i 's|</info> |</info>|' "$file"
    sed -i 's|</title> |</title>|' "$file"
    sed -i 's|</para> |</para>|' "$file"
    sed -i 's|</systemitem> |</systemitem>|' "$file"
done
```

For some reason xmlstarlet sometimes inserts whitespaces when deleting the the tags. I'm sanitizing this with sed. However there are still some whitespace changes, e.g. at the end of the `<para>`s, but I can live with that. This change is only for the localized files, which already have poor formatting and are likely not going to be touched anway.

If you consider this output okay, I'll do the same for SLE 12 SP4.

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 next/openSUSE Leap next *(current `main`, no backport necessary)*
  - [ ] SLE 15 SP1
  - [ ] SLE 15 SP0
- SLE 12
  - [x] SLE 12 SP5
  - [x] SLE 12 SP4

### DIFF
--- a/l10n/sles/de-de/xml/64bit_issues.xml
+++ b/l10n/sles/de-de/xml/64bit_issues.xml
@@ -6,8 +6,7 @@
    <dm:bugtracker/>
    <dm:translation>Ja</dm:translation>
   </dm:docmanager>
- </info><indexterm>
- <primary> 64-Bit-Linux</primary></indexterm> 
+ </info>
  <para>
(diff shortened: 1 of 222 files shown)


tomschr (approved):
Thanks Christoph for taking care! :+1: 
I think, that's fine what you did. As the code isn't touched, we can live with all the smaller issues.

From my perspective, go for it and continue with the other.

tomschr:
By the way, you could combine all the sed expressions into one:

```
sed -i 's|</info> |</info>|;s|</title> |</title>|;s|</para> |</para>|;s|</systemitem> |</systemitem>|' "$file"
```

Perhaps it could be even more "optimized" by using parenthesis so it becomes quite readable:

```
sed -i 's_</\(info|title|para|systemitem\)> _</\1>_' $file
```

PR #1414, opened 2022-12-13
Modernize line graphics in Security and Hardening Guide

janajaeger:
edges.

### PR creator: Description

Apply current brand guide lines to our illustrations

Current main. No backporting necessary. Always include these as PNG only.


taroth21:
@janajaeger : Thanks a lot! I have compared the HTML output of your branch against the latest SLES 15 SP4 version of the Security Guide - all graphics look much better now, thank you for the Zen gardening exercise!

PR #1419, opened 2022-12-23
Bugfix/ldap modules

taroth21:
### PR creator: Description

LDAP chapter: Fix the following issues:

1. First the topic is managing *plugins* not *modules* in 389-ds 
2. Second, this section needs to be before 6.6 (SSSD integration).
3. Third, SSSD section needs a warning in the example before the configuration in point 4 that the ldap access filter relies on memberOf being configured.

* bsc#1204350
* jsc#DOCTEAM-801

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP1
  - [ ] SLE 15 SP0
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4

### DIFF
--- a/xml/security_ldap_modules.xml
+++ b/xml/security_ldap_modules.xml
@@ -21,7 +21,7 @@
   </dm:docmanager>
  </info>
   <para>
-    Use the following command to list all available pug-ins, enabled and
+    Use the following command to list all available plug-ins, enabled and
     disabled. Use your server's hostname rather than the instance name
     of your &ds389;, like the following example hostname of
     <replaceable>LDAPSERVER1</replaceable>:
--- a/xml/security_ldap_modules.xml
+++ b/xml/security_ldap_modules.xml
@@ -37,10 +37,10 @@ ACL Plugin
 ACL preoperation
 [...]</screen>
   <para>
-    The following command enables the MemberOf plug-in referenced in
-    <xref linkend="sec-security-ldap-server-sssd"/>. MemberOf simplifies
+    The following command enables the <literal>MemberOf</literal> plug-in referenced in
+    <xref linkend="sec-security-ldap-server-sssd"/>. <literal>MemberOf</literal> simplifies
     user searches, by returning the user and any groups the user
-    belongs to, with a single command. Without MemberOf, a client must
+    belongs to, with a single command. Without <literal>MemberOf</literal>, a client must
     run multiple lookups to find a user's group memberships.
   </para>
   <screen>
@@ -65,7 +65,7 @@ ACL preoperation
 </screen>
 
 <para>
- Next, configure the plug-in. The following example enables MemberOf to
+ Next, configure the plug-in. The following example enables <literal>MemberOf</literal> to
  search all entries. Use your instance name rather than the server's
  hostname:
 </para>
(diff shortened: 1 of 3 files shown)


taroth21:
Find a PDF of the current version here: 
[cha-security-ldap_draft_20221223.pdf](https://github.com/SUSE/doc-sle/files/10295693/cha-security-ldap_draft_20221223.pdf)

xml/security_ldap_modules.xml
```diff
@@ -15,13 +15,13 @@
...
  <info>
-  <title>Managing modules</title>
+  <title>Managing plug-ins</title>
   <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
    <dm:translation>yes</dm:translation>
   </dm:docmanager>
  </info>
   <para>
-    Use the following command to list all available modules, enabled and
+    Use the following command to list all available pug-ins, enabled and
```
Firstyear, line 24: pug-ins is a typo :) should be "plugins"

xml/security_ldap_modules.xml
```diff
@@ -37,19 +37,19 @@ ACL Plugin
 ACL preoperation
 [...]</screen>
   <para>
-    The following command enables the MemberOf plugin referenced in
-    <xref linkend="sec-security-ldap-server-sssd"/>. MemberOf simplifies
+    The following command enables the <literal>MemberOf</literal> plug-in referenced in
```
Firstyear, line 40: 389-ds calls them 'plugins' so shourld we stay consistent?

Firstyear (changes requested):

xml/security_ldap_modules.xml
taroth21: Oops, thanks, changed to 'plug-ins' now.

xml/security_ldap_modules.xml
taroth21: According to the standards defined in our SUSE style guide for product documentation, the spelling within our product docs is 'plug-ins', see https://documentation.suse.com/style/current/single-html/docu_styleguide/#sec-terminology

Which means as long as I do not cite from the code or UI, I need to stick to our officially agreed spelling (which in this case differs from what 389-ds uses...).

Firstyear: In this case wouldn't it be a "reference" to the memberof plugin in the code though ?

taroth21: Looking at the `screen` elements in l. 47 and l. 73, I'm wondering now, too. Let's ask @dariavladykina, our editor.

dariavladykina (approved):

xml/security_ldap_modules.xml
taroth21: @Firstyear : I discussed this with our editor and she sees it the same way like me. Unless it is a direct quote from the code, the UI (or a command), we should use 'plug-in'. I also checked the RH docs and they actually treat it the same way. 

See for example https://access.redhat.com/documentation/en-us/red_hat_directory_server/11/html/administration_guide/tracking_the_bind_dn_for_plug-in-initiated_updates or https://access.redhat.com/documentation/en-us/red_hat_directory_server/11/html/administration_guide/managing_directory_entries#renaming-users-groups-posix-groups-and-ous (scroll two paras up from the heading '3.1.6.2. Renaming Users, Groups, POSIX Groups, and OUs' to find the examples).

PR #1479, opened 2023-04-03
Installation Quick Start Guide proofing integration

dariavladykina:
Integrated proofing corrections for the Installation Quick Start Guide from VT.


xml/art_installation-sleds.xml
```diff
@@ -923,7 +923,7 @@ disk:
      <listitem>
       <para>
        Click to <guimenu>enable</guimenu> the <literal>&disa; &stiga;</literal>
-       security policy. If installation settings are incompatible with the
+       security policy. If the installation settings are incompatible with the
```
cwickert, line 926: I disagree on that one. "The installation settings" (definitive article) means _all_ installation settings, everything that got configured throughout the installation process and is shown in the "Installation Settings" screen. However we are only referring to a small subset of the settings, only the ones that are relevant for hardening and only the ones that are actually incompatible with the hardening profile. Therefor I'd use "installation settings" (infinite), but maybe this is just me with my German hat on.

Opinions from a native speaker?

cwickert (approved):
LGTM except for the "installation settings". I might be wrong though, please clarify this before merging.

taroth21 (commented):
I agree with Christoph that we should clarify this one point, the rest looks good to me.

xml/art_installation-sleds.xml
dariavladykina: @tahliar , could you help please? :-) I'd only add that to me it seems vice versa - 'all settings' would not carry the definite article, whereas some particular ones would, so the article seems justified to me.

tahliar: Hmm, I think to me, "If **the** installation settings are incompatible" does kinda imply all the settings, or rather the settings as a whole.

But "If installation settings are incompatible" doesn't work for me, either. It definitely feels like there's a word missing. If I've understood the intent correctly, I'd say something like "If **any** installation settings are incompatible".

dariavladykina, line 926:
```suggestion
       security policy. If any installation settings are incompatible with the
```

PR #1486, opened 2023-04-19
Include Raspi Pi Quick in set

taroth21:
### PR creator: Description

We decided to move the Raspberry Pi Quick Start into the set. 

This has the following advantages:

- The Bugzilla information is inherited from the set and does not need to be updated individually for the Quick Start with each new SP.

- During localization hand-off, the Quick Start can be treated like any other deliverable (can be 'locdropped' with a DEF file).  

I also compared PDF and HTML builds before and after the changes: everything looks good. 

* see [internal locdrop howto](https://confluence.suse.com/display/documentation/Performing+a+loc+drop)

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP1
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4

### DIFF
--- a/DC-SLES-raspberry-pi
+++ b/DC-SLES-raspberry-pi
@@ -4,12 +4,12 @@
 ## ----------------------------
 ##
 ## Basics
-MAIN="art_raspberry-pi.xml"
+MAIN="MAIN.SLEDS.xml"
 ROOTID=article-raspberry-pi
 
 ## Profiling
 PROFOS="sles"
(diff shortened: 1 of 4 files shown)


jfaltenbacher (approved):
Great to have that guide included now. Thank you

tomschr (approved):
Thanks Tanja, looks great! :+1:

PR #1499, opened 2023-05-09
Modules and Extensions review

dariavladykina:
A language review of Modules and Extensions before GMC.


cwickert:
This must not be merged as it is a duplicate. I'll merge #1503 instead as it has two more fixes.

PR #1503, opened 2023-05-10
SLED Modules and Extensions Quick Start review

dariavladykina:
SLED Modules and Extensions Quick Start review before GMC.

### DIFF
--- a/xml/art_modules.xml
+++ b/xml/art_modules.xml
@@ -1311,10 +1311,10 @@
     <title>Deleting modules</title>
     <para>
      Note that you should never delete the <guimenu>Basesystem
-      Module</guimenu>. It is also not recommended to delete the
-     <phrase os="sles"><guimenu>Server Applications
-      Module</guimenu>,</phrase><phrase os="sled"><guimenu>&slewe;</guimenu>
-       and <guimenu>Desktop Applications Module</guimenu></phrase>.
+      Module</guimenu>. It is also not recommended to delete the <phrase
+      os="sles"><guimenu>Server Applications Module</guimenu></phrase><phrase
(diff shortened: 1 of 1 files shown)


xml/art_modules.xml
```diff
@@ -1313,7 +1313,7 @@
      Note that you should never delete the <guimenu>Basesystem
       Module</guimenu>. It is also not recommended to delete the
      <phrase os="sles"><guimenu>Server Applications
-      Module</guimenu></phrase><phrase os="sled">, <guimenu>&slewe;</guimenu>
+      Module</guimenu>,</phrase><phrase os="sled"><guimenu>&slewe;</guimenu>
```
cwickert, line 1316: There must be no comma. Both the old and the fixed versions are incorrect. I'll fix this before merging.

xml/art_modules.xml
```diff
@@ -1450,7 +1450,7 @@ https://www.suse.com/documentation/sles-15/singlehtml/art_modules/art_modules.ht
...
```
cwickert, line 1453: Same as above

cwickert (approved):
Two colons within the profiling need to go, the rest looks good.

xml/art_modules.xml
dariavladykina: image

cwickert: Yes, the old version was incorrect. Your fix however would break SLES. It would become:
> It is also not recommended to delete the _Server Applications Module_,.

Note the comma and the comma before the period.

cwickert: Anyway, I fixed it, this is nothing to worry about.

cwickert:
I still consider some of the commas useless, but will merge the PR as is.

PR #1583, opened 2023-09-05
Dynamic debug for the kernel

sounix000:
### PR creator: Description

Short introduction to dynamic debug.

* bsc#https://bugzilla.suse.com/show_bug.cgi?id=1176726
* jira#https://jira.suse.com/browse/DOCTEAM-228

- SLE 15/openSUSE Leap 15.x
  - [x] SLE 15 SP4/openSUSE Leap 15.5
  - [ ] SLE 15 SP1
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4

### DIFF
--- a/xml/tuning_dynamic_debug.xml
+++ b/xml/tuning_dynamic_debug.xml
@@ -1,0 +1,192 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE chapter
+[
+  <!ENTITY % entities SYSTEM "generic-entities.ent">
+    %entities;
+]>
+<chapter xmlns="http://docbook.org/ns/docbook" xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:xlink="http://www.w3.org/1999/xlink" version="5.0" xml:id="cha-tuning-dynamic-debug">
+  <title>Dynamic debug&mdash;kernel debugging messages</title>
+  <info>
+    <abstract>
+      <para>
+        Dynamic debug is a powerful debugging feature in the Linux kernel that
+        allows you to enable and disable debugging messages at runtime without
+        the need to recompile the kernel or reboot the system.
+      </para>
+    </abstract>
+    <dm:docmanager xmlns:dm="urn:x-suse:ns:docmanager">
+      <dm:bugtracker></dm:bugtracker>
+      <dm:translation>yes</dm:translation>
+    </dm:docmanager>
+  </info>
+  <para>
+    You can use dynamic debugging in several situations, such as:
+  </para>
+  <itemizedlist>
+    <listitem>
+      <para>
+        Troubleshooting kernel issues
+      </para>
+    </listitem>
+    <listitem>
+      <para>
+        Developing drivers for new hardware
+      </para>
+    </listitem>
+    <listitem>
+      <para>
+        Tracing and auditing security events
+      </para>
+    </listitem>
+  </itemizedlist>
+  <sect1 xml:id="sec-tuning-dynamic-debug-benefits">
+    <title>Benefits of dynamic debugging</title>
+
+    <para>
+      Certain benefits of dynamic debugging are listed below:
+    </para>
+
+    <variablelist>
+      <varlistentry>
+        <term>Real-time debugging</term>
+        <listitem>
+          <para>
+            Dynamic debugging enables debugging messages without requiring a
+            system reboot. This real-time capability is crucial for diagnosing
+            issues in production environments.
+          </para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>Selective debugging</term>
+        <listitem>
+          <para>
+            You can enable debugging messages for specific parts of the kernel
+            or even individual modules, allowing you to focus on relevant
+            information.
+          </para>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>Performance tuning</term>
+        <listitem>
+          <para>
+            Use dynamic debugging to monitor and optimize kernel performance by
+            selectively enabling or disabling debugging messages based on the
+            current analysis requirements.
+          </para>
+        </listitem>
+      </varlistentry>
+    </variablelist>
+  </sect1>
+  <sect1 xml:id="sec-tuning-dynamic-debug-check">
+    <title>Checking the status of dynamic debug</title>
+
+    <para>
+      For supported kernel versions that are installed by default, dynamic
+      debug is already built-in. To check the status of dynamic debug, run the
+      following command as the root user:
+    </para>
+
+<screen>&prompt.root;<command>zcat /proc/config.gz | grep CONFIG_DYNAMIC_DEBUG</command></screen>
+
+    <para>
+      If dynamic debug is compiled into the kernel, you should see an output
+      similar to the following:
+    </para>
+
+<screen>
+CONFIG_DYNAMIC_DEBUG=y
+CONFIG_DYNAMIC_DEBUG_CORE=y
+</screen>
+  </sect1>
+  <sect1 xml:id="sec-tuning-dynamic-debug-use">
+    <title>Using dynamic debug</title>
+
+    <para>
+      To enable specific debug messages or logs within the running kernel, you
+      can use the <command>echo</command> command and write to the
+      <filename>/sys/kernel/debug/dynamic_debug/control</filename> file.
+    </para>
+
+    <para>
+      The following examples illustrate certain simple uses of dynamic debug:
+    </para>
+
+    <note>
+      <para>
+        Dynamic debug relies on specific debugging macros, such as
+        <systemitem>pr_debug</systemitem>, embedded in the kernel code. These
+        macros are used by kernel developers to insert debugging messages into
+        the code.
+      </para>
+      <para>
+        The examples in this section assume that the
+        <systemitem>pr_debug</systemitem> macro works correctly because dynamic
+        debug is allowed for the running kernel.
+      </para>
+    </note>
+
+    <variablelist>
+      <varlistentry>
+        <term>Enabling debug messages for a specific kernel source code file</term>
+        <listitem>
+          <para>
+            To enable the debug messages for a specific kernel source code
+            file, use the following example:
+          </para>
+<screen>&prompt.root;<command>echo "file <replaceable>FILE_NAME.c</replaceable> +p" > /sys/kernel/debug/dynamic_debug/control</command></screen>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>Enabling debug messages for a specific kernel module</term>
+        <listitem>
+          <para>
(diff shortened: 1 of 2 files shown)


luis-henrix:
> @luis-henrix Please review this PR.

The PR looks good to me, but it would be awesome if I could see the final
(html) version of it.  Is that available somewhere?

xml/tuning_dynamic_debug.xml
```diff
@@ -0,0 +1,209 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE chapter
+[
+  <!ENTITY % entities SYSTEM "generic-entities.ent">
+    %entities;
+]>
+<chapter xmlns="http://docbook.org/ns/docbook" xmlns:xi="http://www.w3.org/2001/XInclude" xmlns:xlink="http://www.w3.org/1999/xlink" version="5.0" xml:id="cha-tuning-dynamic-debug">
+  <title>Dynamic debug&mdash;dynamically enable/disable kernel debug-print code</title>
```
luis-henrix, line 8: I'd suggest a different title for this chapter, as it's goal is to present an overview of the dynamic debug feature, not only how to enable/disable it.  How about something like "Dynamic debug - Kernel debugging messages"?  Or "Dynamic debug - Kernel messages"?

xml/tuning_dynamic_debug.xml
```diff
@@ -0,0 +1,209 @@
...
+        </listitem>
+      </varlistentry>
+    </variablelist>
+  </sect1>
+  <sect1 xml:id="sec-tuning-dynamic-debug-check">
+    <title>Checking the status of dynamic debug</title>
+
+    <para>
+      For supported kernel versions that are installed by default, dynamic
+      debug is already enabled. To check the status of dynamic debug, run the
```
luis-henrix, line 87: "dynamic debug is already enabled": this isn't really correct.  Dynamic debug is **supported** or **available** in the kernel, but it is disabled by default.  The user has to explicitly enable it to see the debugging messages.

xml/tuning_dynamic_debug.xml
```diff
@@ -0,0 +1,209 @@
...
+      following command as the root user:
+    </para>
+
+<screen>&prompt.root;<command>zcat /proc/config.gz | grep CONFIG_DYNAMIC_DEBUG</command></screen>
+
+    <para>
+      If dynamic debug is enabled, you should see an output similar to the
```
luis-henrix, line 94: Same as above: "If dynamic debug is enabled" isn't really correct.  This check only shows that it is supported by the kernel, but not necessarily enabled.

xml/tuning_dynamic_debug.xml
```diff
@@ -0,0 +1,209 @@
...
+        <listitem>
+          <para>
+            To enable debug messages for a specific kernel module, use the
+            following example:
+          </para>
+<screen>&prompt.root;<command>echo "module <replaceable>MODULE_NAME</replaceable> +p" > /sys/kernel/debug/dynamic_debug/control</command></screen>
+        </listitem>
+      </varlistentry>
+      <varlistentry>
+        <term>Enabling debug messages with a specific log level</term>
```
luis-henrix, line 152: This one is interesting! I didn't even knew it existed :-) Can you point me at where you've seen this documented?

xml/tuning_dynamic_debug.xml
```diff
@@ -0,0 +1,209 @@
...
+  <sect1 xml:id="sec-tuning-dynamic-debug-view">
+    <title>Viewing the dynamic debug messages</title>
+
+    <para>
+      You can view the dynamic debug messages that were generated based on the
+      configurations you enabled, by running <command>dmesg</command> and
+      filtering the output with <command>grep</command>. For example:
+    </para>
+
+<screen>&prompt.root;<command>dmesg | grep -i "<replaceable>FILE_NAME.c</replaceable>"</command></screen>
```
luis-henrix, line 199: At least in the PDF file this like is broken into 2 lines; I see this:
```
# dmesg | grep -i
"FILE_NAME.c"
```
Can that be fixed?

luis-henrix (changes requested):
Thanks a lot for sharing the pdf and html files; it does help a lot to review on those formats.

Obviously my comments aren't on the doc style (that's probably better done by someone else).  Some of them are minor and can be ignored but others, in my opinion, should be taking into account.

But overall I think the text if great; thanks for working on this, @sounix000

luis-henrix:
Souvik Sarkar ***@***.***> writes:

> @sounix000 commented on this pull request.
>
>
>
>> +        <listitem>
> +          <para>
> +            Use dynamic debugging to monitor and optimize kernel performance by
> +            selectively enabling or disabling debugging messages based on the
> +            current analysis requirements.
> +          </para>
> +        </listitem>
> +      </varlistentry>
> +    </variablelist>
> +  </sect1>
> +  <sect1 xml:id="sec-tuning-dynamic-debug-check">
> +    <title>Checking the status of dynamic debug</title>
> +
> +    <para>
> +      For supported kernel versions that are installed by default, dynamic
> +      debug is already enabled. To check the status of dynamic debug, run the
>
>> "dynamic debug is already enabled": this isn't really correct. Dynamic debug
>> is **supported** or **available** in the kernel, but it is disabled by
>> default. The user has to explicitly enable it to see the debugging messages.
>
> Not doubting/arguing with you, but I am a bit confused between your comment and
> the information available here:
> https://docs.kernel.org/admin-guide/dynamic-debug-howto.html#kernel-configuration
> I checked on a SLE 15 SP5 VM, and also on OpenSUSE Leap, and in both cases found
> that the flags are set as `y` by default, so I assumed that they are enabled by
> default, and not merely supported.

Sorry, you're right.  I didn't expressed myself correctly.  I simply
wanted a different word to be used here instead of 'enabled' (and
'supported' isn't the right one!).

So, in order to start using this dynamic debug feature two things are
required:

1. Compile the feature into the kernel.  I.e. configure the kernel with
   CONFIG_DYNAMIC_DEBUG.  This is done at kernel build time, it's a static
   thing: if the kernel isn't build with this config option, it's not
   possible to use Dynamic debug.

2. Enable the dynamic debug feature at runtime.  I.e. once we have a
   kernel that was compiled with CONFIG_DYNAMIC_DEBUG, we can
   enable/disable the feature at runtime, as explained in your PR.

The problem with the text is that we are using the 'enable' word for the
two things listed above.  And that's what I found a bit confusing and that
I wanted to point out in my comment.  I'd suggest to use 'enable/disable'
for the runtime operations, and use something else when referring to the
built-in option (i.e. the CONFIG_DYNAMIC_DEBUG kernel config option).

So, here's some suggestions.  Instead of:

"For supported kernel versions that are installed by default, dynamic debug is already enabled."

use:

"For supported kernel versions that are installed by default, dynamic debug is already built-in."

Instead of:

"If dynamic debug is enabled, you should see an output similar to the following"

use:

"If dynamic debug is compiled into the kernel, you should see an output similar to the following"

I hope this clarifies my comment.  Sorry for the confusing.

luis-henrix:
Souvik Sarkar ***@***.***> writes:
>> This one is interesting! I didn't even knew it existed :-) Can you point me at where you've seen this documented?
>
> This is a mixture of several sources of information that I consulted during my research (don't know whether I interpreted the sources correctly):
> - https://www.kernel.org/doc/html/latest/core-api/printk-basics.html#message-logging-with-printk
> - https://www.kernel.org/doc/html/v4.14/admin-guide/dynamic-debug-howto.html#command-language-reference
> - https://github.com/torvalds/linux/blob/master/include/linux/kern_levels.h#L29

Thanks, Souvik.  OK, I've done some quick testing and this doesn't seem to
be accurate.  The 'format' command seems to something similar to what grep
does: the given string is searched for in the dynamic debug format string.
So, the example in your PR:

  echo "format "KERN_DEBUG" +p" > /sys/kernel/debug/dynamic_debug/control

would only show debug messages that would contain the string "KERN_DEBUG".

The handling of the log levels you're mentioning here seem to be a
totally different mechanism, not related with the dynamic debug.  For
example, you can set the kernel log levels even if you don't have
CONFIG_DYNAMIC_DEBUG compiled it.  And these log levels are usually
handled with sysctl or using the procfs file '/proc/sys/kernel/printk'.
(See https://www.kernel.org/doc/html/latest/admin-guide/sysctl/kernel.html#printk)

Anyway, I'd suggest leaving this 'format' command out of the document and
simply leave the reference to the full documentation as you already do.
Otherwise you'd also need to mention other commands that are missing.  For
example, the commands 'function' and 'line' are much more useful than
'format' command, and there aren't being described.

luis-henrix:
> So if I understand correctly, should I omit the entire `Enabling debug messages with a specific log level` section? And leave the rest of the sections as it is?

Correct, my suggestion it to drop the 'Enabling debug messages with a specific log level' section, because:
1. the command it describes (`format`) is not the most useful/common one, and
2. the description isn't correct.
Leaving the section that describes the `format` command would, in my opinion, require other more useful commands to be described as well.  And since this is just an overview of the dynamic debug feature, I'm fine with just the sections showing the 2 most common commands (`file` and `module`) and how to disable the feature.

sounix000:
@luis-henrix I have made all the changes you suggested. Please take a final look before approval. Thanks for your feedback and clarification in making the document technically correct.

xml/tuning_dynamic_debug.xml
```diff
@@ -0,0 +1,190 @@
...
+      <para>
+        The examples in this section assume that the
+        <systemitem>pr_debug</systemitem> macro works correctly because dynamic
+        debug is allowed for the running kernel.
+      </para>
+    </note>
+
+    <variablelist>
+      <varlistentry>
+        <term>Enabling debug messages for a specific file</term>
```
luis-henrix, line 132: This is a minor comment, feel free to ignore if you think it's not worth the trouble changing the wording here, but I'd use:

> Enabling debug messages for a specific kernel source code file

(and the same for the sentence bellow, of course)

luis-henrix (approved):
Thanks a lot for the updates, @sounix000.  From a technical point of view, the section now looks OK to me.

dariavladykina (commented):
LGTM!

dariavladykina (approved):

sounix000:
Backported/Cherry-picked commits:

- SLE15 SP2: https://github.com/SUSE/doc-sle/commit/d505de7bbc7f6a89c2bb1646aa55741930542a08
- SLE15 SP3: https://github.com/SUSE/doc-sle/commit/b22bfca8225130cb843ffcbac0d10b31a5f38a92
- SLE15 SP4: https://github.com/SUSE/doc-sle/commit/10f9e0e43305f378ebb45ab688b29a6bf1dc8a2b
- SLE15 SP5: https://github.com/SUSE/doc-sle/commit/5f3129a9c664691ba8d446b3604f520ddd9e3306

PR #1594, opened 2023-09-20
fix line overflow

jfaltenbacher:
PR creator: Description

Fixed line overflow for PDF version of _Modules and Extensions Quick Start_. 
I noticed it in zh-cn (see https://documentation.suse.com/zh-cn/sles/15-SP5/pdf/article-modules_zh_cn.pdf) Chapter 4, Example 3. But it also occurs in the English Guide.
Please review @cwickert 

PR creator: Are there any relevant issues/feature requests?

- https://jira.suse.com/browse/DOCTEAM-1018

PR creator: Which product versions do the changes apply to?

- SLE 15/openSUSE Leap 15.x
  - [x] SLE 15 SP5/openSUSE Leap 15.5
  - [ ] SLE 15 SP1
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4

### DIFF
--- a/xml/art_modules.xml
+++ b/xml/art_modules.xml
@@ -1637,8 +1637,8 @@ Following packages were found in following modules:
 
 Package  Module or Repository
            SUSEConnect Activation Command
---- -------------------------------------------------------------------------------------
----------- ------------------------------------------------------------------------------
+--- ----------------------------------------------------------------------------
+---------- ---------------------------------------------------------------------
 vim Basesystem Module (sle-module-basesystem/15.4/x86_64)
       SUSEConnect --product sle-module-basesystem/15.4/x86_64
 vim Installed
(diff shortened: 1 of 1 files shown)


cwickert (commented):
Shortening the lines is not the solution.

1. The screens should display exactly the output that you get on a terminal. We should not change it, but display it _as is_.
2. The line length depends on the text length. If the text is longer, the line will get longer to match. The reason it does not match is that daps/fo apply hyphenation to the screens (when they IHMO should not). Originally, `Basesystem_Module_15_SP4_x86_64:SLE-Module-Basesystem15-SP4-Updates` is one word. Only because fo hyphenates at `-Updates`, text and line no longer match. IMHO this is a bug.
3. If you now shorten the line for the PDF, it becomes too short in HTML
Before:
Screen Shot 2023-09-28 at 11 24 34
After:
Screen Shot 2023-09-28 at 11 25 27

The correct and simple solution is to make the whole text smaller. It's bigger than the other ones because that particular screen lacks the processing instruction `<?dbsuse-fo font-size="0.70em"?>`

Fixed in 6192008cd41f7671b03bf0894a9b168ef536b9d9

cwickert:
I fixed the English and the localized versions down to 15 SP3. Other branches are not affected and/or no longer supported.

The change builds cleanly, but the maintenance branches were already broken. Not my fault. :-/

PR #1595, opened 2023-09-20
Updated drop-in files section (bsc#1212777)

tbazant:

* bsc#1212777
* jsc#...

- SLE 15/openSUSE Leap 15.x
  - [x] SLE 15 SP4/openSUSE Leap 15.5
  - [x] SLE 15 SP1
- SLE 12
  - [x] SLE 12 SP5
  - [x] SLE 12 SP4

### DIFF
--- a/xml/systemd.xml
+++ b/xml/systemd.xml
@@ -1483,7 +1483,7 @@ UMask=007</screen>
     <sect2 xml:id="sec-boot-systemd-custom-drop-in">
       <title>Creating drop-in files</title>
       <para>
-        For minor changes of a configuration file, use so-called drop-in files
+        For changes of configuration files, use so-called drop-in files
         or <emphasis>drop-ins</emphasis>. Drop-ins let you extend both unit and
         daemon configuration files without having to edit or override the files
         themselves. Drop-ins have higher precedence over main configuration
--- a/xml/systemd.xml
+++ b/xml/systemd.xml
@@ -15,9 +15,9 @@
   <para>
     &systemd; initializes the system. It has the process ID 1. &systemd; is
     started directly by the kernel and resists signal&nbsp;9, which normally
-    terminates processes. All other programs are either started directly by
+    terminates processes. All other programs are started directly by
     &systemd; or by one of its child processes. &systemd; is a replacement for
-    the System V init daemon and fully compatible with System V init (by
+    the System V init daemon and is fully compatible with System V init (by
     supporting init scripts).
   </para>
   <para>
@@ -1303,7 +1303,7 @@ WantedBy=multi-user.target<co xml:id="co-service-wrapper-target"/></screen>
                 <literal>multi-user.target</literal> also starts the init
                 script when booting into <literal>graphical.target</literal>.
                 If it should only be started when booting into the display
-                manager, user <literal>graphical.target</literal>.
+                manager, use <literal>graphical.target</literal>.
               </para>
             </callout>
           </calloutlist>
@@ -1438,7 +1438,7 @@ WantedBy=multi-user.target<co xml:id="co-service-wrapper-target"/></screen>
       </para>
       <para>
         As an exercise, change how long the system waits for &mariadb; to
-        start. As root, run <command>systemctl edit --full
+        start. As &rootuser;, run <command>systemctl edit --full
         mariadb.service</command>. The opened file is similar to the following:
       </para>
 <screen>
@@ -1580,7 +1580,7 @@ RateLimitIntervalSec=0</screen>
           </listitem>
           <listitem>
             <para>
-              <literal>50</literal> is reserved for unit files drop-ins created
+              <literal>50</literal> is reserved for unit drop-in files created
               with <command>systemctl set-property</command>.
             </para>
           </listitem>
@@ -1617,7 +1617,7 @@ RateLimitIntervalSec=0</screen>
             <para>
               <command>systemd-analyze cat-config
               <replaceable>DAEMON_NAME_OR_PATH</replaceable></command> copies
-              contents of a configuration file and drop-ins for a &systemd;
+              the contents of a configuration file and drop-ins for a &systemd;
               daemon, for example:
             </para>
 <screen>&prompt.user;systemd-analyze cat-config systemd/journald.conf</screen>
@@ -2052,7 +2052,7 @@ WantedBy=multi-user.target
       <para>
         Stopping <systemitem class="daemon">dbus</systemitem> or restarting it
         in the running system is similar to an attempt to stop or restart PID
-        1. It breaks &systemd; client/server communication and make most
+        1. It breaks the &systemd; client/server communication and makes most
         &systemd; functions unusable.
       </para>
       <para>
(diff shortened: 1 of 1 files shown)


xml/systemd.xml
```diff
@@ -1485,51 +1485,75 @@ UMask=007</screen>
...
       <title>Creating drop-in files</title>
       <para>
-        For minor changes of a configuration file, use so-called drop-in files.
-        Drop-in files let you extend the configuration of unit files without
-        having to edit or override the unit files themselves.
+        For minor changes of a configuration file, use so-called drop-in files
+        or <emphasis>drop-ins</emphasis>. Drop-ins let you extend the
+        configuration of unit files without having to edit or override the unit
+        files themselves. Drop-ins have higher precedence over main
+        configuration files. They are stored in the following directories:
```
Werkov, line 1492: There are two classes of config files: _unit_ files and daemon config files.
So drop-ins let you extend both classes of configs.

Werkov (changes requested):
Looks good (@fbuihuu may refine the numeric ranges but I think these are the latest values).
I'd only emphasize more the difference between unit files (with their drop-ins) and daemon config files (with their drop-ins).

fbuihuu:
> I'd only emphasize more the difference between unit files (with their drop-ins) and daemon config files (with their drop-ins).

I agree.

For viewing/editing unit files there're `systemctl cat`/`systemctl edit` respectively but for config files there's only `systemd-analyze cat-config` currently.

fbuihuu:
> Looks good (@fbuihuu may refine the numeric ranges but I think these are the latest values).

```
[0 - 19]  : systemd upstream
[20 - 25] : systemd SUSE
[26 - 29] : systemd packages
[30 - 49] : unused
50 : `systemctl set-property` for unit files only
```
I'm not sure where these ranges are coming from. They look OK although I'm not sure to understand why we have such small ranges reserved for downstream and packages. Also why `[30 - 49]` is wasted.

Maybe we shouldn't go into such level of details if the doc is aimed at users/customers, simply stating that they should use the range `[60 - 99]` only might be enough, no ?

xml/systemd.xml
```diff
@@ -1485,51 +1483,78 @@ UMask=007</screen>
...
-        having to edit or override the unit files themselves.
+        For minor changes of a configuration file, use so-called drop-in files
+        or <emphasis>drop-ins</emphasis>. Drop-ins let you extend both unit and
+        daemon configuration files without having to edit or override the files
+        themselves. Drop-ins have higher precedence over main configuration
+        files. They are stored in the following directories:
       </para>
+      <variablelist>
+        <varlistentry>
+          <term><filename>/etc/systemd/*.conf.d/</filename></term>
```
Werkov, line 1494: Paths for unit files:
/etc/systemd/system/*.service.d/

xml/systemd.xml
```diff
@@ -1485,51 +1483,78 @@ UMask=007</screen>
...
+          <listitem>
+            <para>
+              Drop-ins added and customized by system administrators.
+            </para>
+          </listitem>
+        </varlistentry>
+        <varlistentry>
+          <term><filename>/usr/lib/systemd/*.conf.d/</filename></term>
```
Werkov, line 1502: Path for unit files:
/usr/lib/systemd/system/*.service.d/

See `systemd.unit(5)` for full _unit search path_.

xml/systemd.xml
```diff
@@ -1485,51 +1483,78 @@ UMask=007</screen>
     <sect2 xml:id="sec-boot-systemd-custom-drop-in">
       <title>Creating drop-in files</title>
       <para>
-        For minor changes of a configuration file, use so-called drop-in files.
-        Drop-in files let you extend the configuration of unit files without
-        having to edit or override the unit files themselves.
+        For minor changes of a configuration file, use so-called drop-in files
```
Werkov, line 1486: Maybe drop _minor_?
As a drop-ins may be the preferred way of config modification in the future.
Does it make sense, @fbuihuu ?

Werkov (changes requested):
I've added a few points for unit/non-unit drop-ins distinction.

The ranges aren't changed in this PR, that is what is in SLE documentation since some time (and they're consistent with [openSUSE doc](https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Shipping_unit_file_drop-ins)).

I notice the added commit includes also some stylistic changes -- I didn't look at them, please add them to a separate commit (next time ;-) for easier review.

xml/systemd.xml
tbazant, line 1486:
```suggestion
        For changes of configuration files, use so-called drop-in files
```

dariavladykina (commented):
Hi, please note some suggestions here. Thanks!

tbazant (commented):
applied suggestions

tbazant:
> I've added a few points for unit/non-unit drop-ins distinction.
> 
> The ranges aren't changed in this PR, that is what is in SLE documentation since some time (and they're consistent with [openSUSE doc](https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Shipping_unit_file_drop-ins)).
> 
> I notice the added commit includes also some stylistic changes -- I didn't look at them, please add them to a separate commit (next time ;-) for easier review.

my bad, i thought technical matters are complete, then i asked for style review which is out workflow. any additions to the current state BTW? or should i merge finally?

fbuihuu:
> The ranges aren't changed in this PR, that is what is in SLE documentation since some time (and they're consistent with [openSUSE doc](https://en.opensuse.org/openSUSE:Systemd_packaging_guidelines#Shipping_unit_file_drop-ins)).

Maybe it's still possible the change the ranges affected to downstream. I currently don't see why we affected such small ranges fo SUSE purposes. Something like the following seems better:

```
[0 - 19] : upstream systemd
[20 - 29]: SUSE systemd
[30 - 39]: SUSE packages
[40 - 49]: 3rd party packages
```

BTW the openSUSE doc states:

> Hence by using a two-digit number above range [0-29], users will be assured that none of the drop-in files shipped by SUSE will override users own drop-in files in the future.

I think this should be corrected to claim that only drop-ins with `priority >= 50`  are safe from a user POV.

fbuihuu:
> 
> ```
> [0 - 19] : upstream systemd
> [20 - 29]: SUSE systemd
> [30 - 39]: SUSE packages
> [40 - 49]: 3rd party packages
> ```

I updated the openSUSE doc to use these ranges. SUSE systemd/packages ranges having a short range doesn't make really sense.

I think the SLE doc should do the same. I don't think these "system" ranges are used by customers.

tbazant:
> > ```
> > [0 - 19] : upstream systemd
> > [20 - 29]: SUSE systemd
> > [30 - 39]: SUSE packages
> > [40 - 49]: 3rd party packages
> > ```
> 
> I updated the openSUSE doc to use these ranges. SUSE systemd/packages ranges having a short range doesn't make really sense.
> 
> I think the SLE doc should do the same. I don't think these "system" ranges are used by customers.

i've updated the SLE sources accordingly, will be published soon

PR #1673, opened 2024-03-14
Remove leftovers after switching to doc-kit

cwickert:
### PR creator: Description

Remove the dangling symlinks to `entity-decl.ent` in the `l10n` folder as well as the unused `entities.abbrev` file(s). None of this is used after the move to doc-kit.

Only SLE 12 SP5, other versions are already fixed.


taroth21 (approved):
LGTM, thanks!

PR #1685, opened 2024-04-04
Upgrade Guide: integrate proofing corrections

dariavladykina:
### PR creator: Description

Integrate proofing corrections for the Upgrade Guide from proofing drop 1.

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP5/openSUSE Leap 15.5
  - [ ] SLE 15 SP1
- SLE 12
  - [ ] SLE 12 SP5
  - [ ] SLE 12 SP4

### DIFF
--- a/xml/common_intro_available_doc.xml
+++ b/xml/common_intro_available_doc.xml
@@ -45,7 +45,7 @@
    <!--based on 'Support Resources' listed at https://www.suse.com/support/kb/-->
    <listitem>
     <para>
-     If you have run into an issue, also check out the Technical Information
+     If you run into an issue, check out the Technical Information
      Documents (TIDs) that are available online at <link xlink:href="https://www.suse.com/support/kb/"/>.
      Search the &suse; Knowledgebase for known solutions driven by customer need.
      </para>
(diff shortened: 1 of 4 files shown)


xml/common_intro_available_doc.xml
```diff
@@ -45,7 +45,7 @@
    <!--based on 'Support Resources' listed at https://www.suse.com/support/kb/-->
    <listitem>
     <para>
-     If you have run into an issue, also check out the Technical Information
```
taroth21, line 48: This needs to be fixed in doc-kit actually (because this file will be overwritten with the next doc-kit run). I'll fix it in doc-kit, but we can keep it here, too.

xml/common_intro_convention.xml
```diff
@@ -106,7 +106,7 @@
    <para>
     Commands can be split into two or multiple lines by a backslash character
     (<literal>\</literal>) at the end of a line. The backslash informs the shell that
-    the command invocation will continue after the line's end:
```
taroth21, line 109: Same here (doc-kit)

taroth21 (approved):
Will add the two changes to the common* files also to doc-kit (see my comments).

PR #1713, opened 2024-05-15
Remove Minimal VM

dmpop:
### PR creator: Description

Remove all Minimal VM-related content.


tomschr (commented):
Thanks @dmpop all looks fine. :+1: 

Maybe one minor thing to ask for? In the `DEF-SLES-manuals` file  line 20 refers to "SLES-minimal-vm". Could you remove that?

These DEF files are/were(?) a Karl thing. I don't know if they are still used. But I think it would be a good idea to keep it consistent. When the doc is gone, the line should be removed as well.

Thank you!

PR #1739, opened 2024-06-24
Security and Hardening Guide: integrate proofing corrections

dariavladykina:
### PR creator: Description

Integrate proofing corrections for the Security and Hardening Guide from proofing drop 3.

- SLE 15/openSUSE Leap 15.x
  - [ ] SLE 15 SP6/openSUSE Leap 15.6
  - [ ] SLE 15 SP5/openSUSE Leap 15.5

- SLE 12
  - [ ] SLE 12 SP5

### DIFF
--- a/xml/apparmor_profiles.xml
+++ b/xml/apparmor_profiles.xml
@@ -66,7 +66,7 @@
    <listitem>
     <para>
      Capability entries are profile entries for any of the POSIX.1e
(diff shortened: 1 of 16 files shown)


taroth21 (changes requested):
One comment and suggestion here, the rest looks good to me. I first was not sure about the profiling changes in `slemicro_pam_login.xml` but as they have been confirmed with our maintainer for the SLE micro docs, they should be fine.

taroth21 (approved):
Thanks!

PR #1754, opened 2024-07-29
Add metadata for SLE 12 SP5

taroth21:
### PR creator: Description

Add metadata from 'Product Docs Metadata per deliverable' spreadsheet. 
Metadata for the deliverables that also exist in Code 15 have been (manually) backported from https://github.com/SUSE/doc-sle/commit/b8e3d0e9b357932a66942711732b856e30c7f901#diff-088e4f6f8df50179fb4221c42e630106703e4e8da5ed561784ee942590f7c6c7 (metadata for oldest supported Code 15 version).

Link to our [internal page](https://confluence.suse.com/display/documentation/Adding+metadata+to+all+SUSE+documentation).

### Open Questions

- For the guides that exist only in this branch: check if the task metadata is appropriate.

### DIFF
--- a/xml/book_gnomeuser.xml
+++ b/xml/book_gnomeuser.xml
@@ -33,6 +33,24 @@
    </dm:bugtracker>
    <dm:translation>yes</dm:translation>
   </dm:docmanager>
+  <meta name="title" its:translate="yes">&gnomeuser;</meta>
+  <meta name="series" its:translate="no">Products &amp; Solutions</meta>
+  <meta name="description" its:translate="yes">How to use the GNOME desktop on &productname;</meta>
+  <meta name="social-descr" its:translate="yes">Use the GNOME desktop</meta>
+  <meta name="task" its:translate="yes">
+    <phrase>Configuration</phrase>
+    <phrase>Maintenance</phrase>
+  </meta>
+  <revhistory xml:id="rh-book-gnomeuser">
+    <revision>
+      <date>2019-12-09</date>
+      <revdescription>
+        <para>
+          Updated for the initial release of &productname; &productnumber;.
+        </para>
+      </revdescription>
+    </revision>
+  </revhistory>
  </info>
  <xi:include href="gnomeuser_intro.xml"/>
 <!-- ===================================================================== -->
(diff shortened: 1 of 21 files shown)


xml/book_gnomeuser.xml
```diff
@@ -33,6 +33,24 @@
    </dm:bugtracker>
    <dm:translation>yes</dm:translation>
   </dm:docmanager>
+  <meta name="title" its:translate="yes">&gnomeuser;</meta>
+  <meta name="series" its:translate="no">Products &amp; Solutions</meta>
+  <meta name="description" its:translate="yes">How to use the GNOME desktop on &productname;</meta>
+  <meta name="social-descr" its:translate="yes">Use the GNOME desktop</meta>
+  <meta name="task" its:translate="yes">
```
taroth21, line 40: @janajaeger : For the GNOME Smart Doc, I noticed that @jfaltenbacher used different tasks (which I find a bit more appropriate): 

<meta name="task">
        <phrase>Configuration</phrase>
        <phrase>Installation</phrase>
</meta>

Should we agree on using the latter? If yes, I need to update this in all the other maintenance branches, too.

janajaeger: If that means that you have to touch it in four or five other branches, I'd say, don't. We have far too much metadata on our hands as it is. Rather revert Julia's to be consistent?

jfaltenbacher: from my point of view: The Gnome User guide is way more extensive and not only features first steps but also maintenance instructions whereas the Smart Docs article only focuses on first steps (Installation). But Tanja and me can discuss tomorrow in our call

taroth21: That difference makes sense, Julia - then I suggest to leave the tasks as they are for the GNOME Smart Doc and the GNOME User Guide.

janajaeger: @jfaltenbacher make sure your task meta tag also carries the its:translate="no" attribute ... not relevant to this PR, but as we a are discussing the metadata here, I thought, I'd raise it here ;)

janajaeger (approved):
LGTM

PR #1822, opened 2025-03-27
Product renaming (SuMa -> SMLM)

taroth21:
### PR creator: Description

With the upcoming version 5.1, **SUSE Manager** will be renamed to **SUSE Multi-Linux Manager** (already released SuMa versions keep their name and their documentation URLs).

The SLE 15 SP7 documentation needs to reflect the product name change. 

Also, the 'generic' URL `https://documentation.suse.com/suma` (which always redirects to the docs for the latest released SuMa version) will be replaced with `https://documentation.suse.com/multi-linux-manager` moving forward.

https://jira.suse.com/browse/DOCTEAM-1754

- SLE 15/openSUSE Leap 15.x
  
15 SP7 only - NO backports to older versions!

### DIFF
--- a/xml/ay_software.xml
+++ b/xml/ay_software.xml
@@ -81,7 +81,7 @@
       <term>SUSE-manager-server</term>
       <listitem>
        <para>
-        SUSE Manager Server
+        &smlm; Server
        </para>
       </listitem>
      </varlistentry>
@@ -89,7 +89,7 @@
       <term>SUSE-manager-retail-branch-server</term>
       <listitem>
        <para>
-        SUSE Manager for Retail
+        &smlm-retail;
        </para>
       </listitem>
      </varlistentry>
@@ -97,7 +97,7 @@
       <term>SUSE-manager-proxy</term>
       <listitem>
        <para>
-        SUSE Manager Proxy
+        &smlm-proxy;
        </para>
       </listitem>
      </varlistentry>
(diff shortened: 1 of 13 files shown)


xml/ay_software.xml
```diff
@@ -81,23 +81,23 @@
       <term>SUSE-manager-server</term>
       <listitem>
        <para>
-        SUSE Manager Server
```
taroth21, line 84: How the AY folks want to handle this for older SuMa products vs. the new SMLM product is still under discussion, see https://bugzilla.suse.com/show_bug.cgi?id=1239854 

Therefore I only updated the clear text names for the products (and replaced them with the new entities) and did not modify the available values for the productname tag yet...

xml/klp.xml
```diff
@@ -295,10 +295,10 @@ Activate with: SUSEConnect -p sle-module-live-patching/&productnumber-regurl;/x8
      When the <package>kernel-default</package> package is installed, the update
      manager prompts you to reboot the system. To prevent this message from
      appearing, you can filter out kernel updates from the patching operation.
-     This can be done by adding package locks with Zypper. &susemgr; also makes
+     This can be done by adding package locks with Zypper. &smlm; also makes
      it possible to filter channel contents (see
-     <link xlink:href="https://documentation.suse.com/suma/4.3/en/suse-manager/administration/live-patching.html">Live
```
taroth21, line 300: Adjusted pointer to latest release version of the docs. Might need to be updated after SMLM 5.1 is released.

lvicoun (approved):
LGTM. Thanks!

xml/klp.xml
taroth21: Not that I would know of...

